PR #5, opened 2022-09-22
simplify useHandle for synchronous repo.find

nornagon:
i think it doesn't make sense for `setHandle` to be exposed here and this
should just be `useMemo`, but taking it one step at a time.


pvh:
superseded but thanks :)

PR #6, opened 2022-10-31
Fix build errors in network-websocket and react-demo

alexjg:


pvh:
Thanks! The surprising one is the typo...

PR #7, opened 2022-11-22
remove peers on disconnect

geoffreylitt:
(result of me and @paulsonnentag pairing)

Reloading a client tab in Chrome creates a new peer. (Unclear if this is expected behavior?)

Previously we would see the server continue to send messages to the now-disconnected peer; thus reloading the same tab a lot would result in a lot of "phantom peers" building up.

This PR handles socket disconnection by removing the peer from the network adapter and collection/doc synchronizers.

It's been tested with a sync server running on a DigitalOcean box, when we connect a few clients and reload them a lot, we've confirmed that no messages get sent out to disconnected peers.

Questions / todos where we need more context from @pvh:

- Is it correct/reasonable to remove the peer from all synchronizers once it disconnects? Wasn't sure how long-lived peers are expected to be.
- Is it expected behavior that reloading a tab in Chrome creates a new peer with a new shared worker ID?
- We should probably add similar behavior to other network adapters, I haven't looked at anything except the Node WS adapter
- Are "leave" messages expected from a client before it disconnects? There's some stub code for handling them but they don't seem to happen in practice.

### DIFF
--- a/packages/automerge-repo-network-websocket/src/NodeWSServerAdapter.ts
+++ b/packages/automerge-repo-network-websocket/src/NodeWSServerAdapter.ts
@@ -30,11 +30,16 @@ export class NodeWSServerAdapter
   connect(peerId: PeerId) {
     this.peerId = peerId
     this.server.on("connection", (socket) => {
-      // When a socket closes, or disconnects, remove it from the array.
-      // TODO
-      //      socket.on("close", () => {
-      //        this.sockets = this.sockets.filter((s) => s !== socket)
-      //})
+      // When a socket closes, or disconnects, remove it from our list
+      socket.on("close", () => {
+        for (const [otherPeerId, otherSocket] of Object.entries(this.sockets)) {
+          if(socket === otherSocket) {
+            this.emit("peer-disconnected", { peerId: otherPeerId as PeerId })
+            delete this.sockets[otherPeerId as PeerId]
+          }
+        }
+
+      })
 
       socket.on("message", (message) =>
         this.receiveMessage(message as Uint8Array, socket)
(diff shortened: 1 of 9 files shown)


packages/automerge-repo-network-websocket/src/NodeWSServerAdapter.ts
```diff
@@ -111,6 +120,9 @@ export class NodeWSServerAdapter
         )
         break
       case "leave":
```
pvh, line 122: Originally, this library was very specific about peers creating individual channels via "join"ing them and then chatting on just that channel about a single docId. This was horribly inefficient if you had lots of docIds and so over time that got de-emphasized. We still do peer discovery via a particular channel (named "sync_channel") but beyond that channels are now really just a key on a message used to multiplex out incoming traffic.

All this is to say that "leave" was never really used but at least conceptually more important when you might join or leave channels based on your interest in a particular document.

packages/automerge-repo-network-websocket/src/NodeWSServerAdapter.ts
```diff
@@ -30,11 +30,16 @@ export class NodeWSServerAdapter
...
-      // When a socket closes, or disconnects, remove it from the array.
-      // TODO
-      //      socket.on("close", () => {
-      //        this.sockets = this.sockets.filter((s) => s !== socket)
-      //})
+      // When a socket closes, or disconnects, remove it from our list
+      socket.on("close", () => {
+        for (const [otherPeerId, otherSocket] of Object.entries(this.sockets)) {
+          if(socket === otherSocket) {
+            this.emit("peer-disconnected", { peerId: otherPeerId as PeerId })
```
pvh, line 37: my hope had been at one point that we might have a more principled way of managing the set of peers for a given document but i can't really see how to avoid copying the peers list into every docsynchronizer. this is reasonable.

pvh: we want at least the equivalent of this code for tabs in the MessageChannelNetworkAdapter as well, but AFAICT there's no way to actually detect a MessageChannel disconnection :(

pvh (commented):
Overall I think we can merge this, but it introduces a new message type that other network adapters should implement.

PR #9, opened 2022-11-28
Add option to log ops in message

paulsonnentag:


packages/automerge-repo/src/synchronizer/DocSynchronizer.ts
```diff
@@ -2,10 +2,12 @@ import EventEmitter from "eventemitter3"
 import { Synchronizer, SyncMessages } from "./Synchronizer"
 import * as Automerge from "@automerge/automerge"
 import { DocHandle, DocumentId } from "../DocHandle"
-import { ChannelId, PeerId } from "../network/NetworkSubsystem"
+import { ChannelId, DecodedMessage, PeerId } from "../network/NetworkSubsystem"
 import debug from "debug"
+import { decodeChange, DecodedSyncMessage } from "@automerge/automerge"
 const log = debug("DocSynchronizer")
 const conciseLog = debug("DocSynchronizerConcise") // Only logs one line per receive/send
+const opsLog = debug("DocSynchronizerOps") // Log list of ops of each message
```
pvh, line 10: this (and above) can use a colon `DocSynchronizer:Ops` to allow more targeted logging

packages/automerge-repo/src/synchronizer/DocSynchronizer.ts
```diff
@@ -69,13 +71,9 @@ export class DocSynchronizer
...
-      conciseLog(
-        `📣 sendSyncMessage\t${JSON.stringify({
-          documentId: this.handle.documentId,
-          peerId,
-          messageLength: message.byteLength,
-        })}`
-      )
+
+      conciseLog(`${peerId} ⬅️️  ${this.handle.documentId} (${message.byteLength} bytes)`)
+      opsLog(getChangesOfMessage(decoded))
```
pvh, line 76: I believe this will execute (and can be quite expensive) even if the logging is not enabled. you should guard it. (the debug library has instructions for how to check that)

packages/automerge-repo/src/synchronizer/DocSynchronizer.ts
```diff
@@ -173,3 +166,10 @@ export class DocSynchronizer
     })
   }
 }
+
+
+function getChangesOfMessage (decoded: DecodedSyncMessage): any  {
```
pvh, line 171: maybe rename to something more specific: `decodeChangesToJson` but i might just reiterate these two lines if it's only ever going to be there twice

pvh (changes requested):

pvh (approved):

PR #10, opened 2022-12-02
Add fix and test for syncValue not returning following an incremental load on an existing document

acurrieclark:
If a document is looked for in the repo, and it exists in the `StorageAdapter`, but not in any of the peers, it is loaded correctly, but no sync messages are generated.

This pull request unblocks the `syncValue()` method, which currently waits for a `syncing` event. `incrementalLoad` only emits a `ready` event.

NB. [This test](https://github.com/pvh/automerge-repo/blob/0746eba5c9b9159f598634a9427662b2d93817fe/packages/automerge-repo/test/Repo.test.ts#L117) was already failing when I started to look at this issue


pvh:
Awesome, thanks! Appreciate the test with the patch.

acurrieclark:
Most welcome. Thanks Peter

PR #12, opened 2022-12-14
A peer no longer receives sync messages if it leaves and then rejoins

acurrieclark:
This PR contains a failing test to show that a peer registered with a `docSynchronizer` no longer receives sync messages if it leaves and then rejoins.


acurrieclark:
Added a fix which keeps the peer state in memory, but adds the peer back to the peers list on re-join

pvh (commented):
Aha! Nice catch.

PR #13, opened 2022-12-14
Add `ChangeOptions` parameter to DocHandle change method

acurrieclark:


pvh:
Looks good but that should be an optional parameter, yea? In the interest of improving test coverage would you like to add a very simple test so I don't regress it in the future?

acurrieclark:
Does the default value not make it optional?

Happy to add a test. Will sort.

pvh:
Oh, yes, of course it does. That's fine.

PR #18, opened 2023-02-03
Add options parameter to `localforage` adapter

acurrieclark:
I was looking for a way to specify the localforage configuration for different repos (within the same application) and thought this might be the most logical solution.

### DIFF
--- a/packages/automerge-repo-storage-localforage/.mocharc.json
+++ b/packages/automerge-repo-storage-localforage/.mocharc.json
@@ -1,0 +1,5 @@
+{
+  "extension": ["ts"],
+  "spec": "test/*.test.ts",
+  "loader": "ts-node/esm"
+}
--- a/packages/automerge-repo-storage-localforage/package.json
+++ b/packages/automerge-repo-storage-localforage/package.json
@@ -10,7 +10,8 @@
   "main": "dist/index.js",
   "scripts": {
     "build": "tsc",
-    "watch": "npm-watch"
+    "watch": "npm-watch",
+    "test": "mocha --no-warnings --experimental-specifier-resolution=node"
   },
   "dependencies": {
     "automerge-repo": "^0.0.44",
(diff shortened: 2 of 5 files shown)


pvh:
can you not just configure the library without passing in? seems like that would be simpler. i do appreciate the addition of tests, though. full disclosure is that i think the localforage implementation is fine but we should probably replace it with a proper indexedb implementation that supports range queries to fetch all the pieces of a document at once. i'm more inclined just to accept this on those grounds but if you can just call localforage.config() without any of this that might be easier still.

acurrieclark:
My understanding is that you can certainly call `localforage.config()` to configure it globally, but I am looking for 2 different implementations, each in a different `repo` instance. 

I think do to that, you would need to [create 2 separate instances](https://localforage.github.io/localForage/#multiple-instances-createinstance) of `localforage` too.

acurrieclark:
Just to clarify why I might need it:
I am looking to have 2 different repos, with different syncing setups, which may have documents with the same ids.

I don’t want there to be conflict between the 2 repos, as the documents would have the same ids but be in deliberately different states.

essentially, one would be offline only, the other with online sync.

pvh:
ohhhhhhh. my original design goal was to be able to manage multiple separate collections with different syncing policies in a single repo (hence the separation of doccollection and repo) but i haven't had a usecase for it. maybe using two repos is better... have you considered how to handle moving docs from one to another / having a doc in both? would love to talk more about this.

acurrieclark:
I have been thinking about it a great deal, and the other main hurdle right now is that it's not possible to load a document in to `repo` without calling `create` then merging the existing doc in (unless I am mistaken?)

Equally, it's not currently possible to create a new handle with a specific id. So right now, it's not possible to have 2 docs in a repo with the same ID anyway. I am perhaps getting ahead of myself!

acurrieclark:
Following a discussion with @pvh about how to work on this going forward, this may no longer be a sensible way to work with multiple docs. Closing for now.

PR #20, opened 2023-02-16
automerge-repo: Fix failing test when patches are emitted

acurrieclark:
There was a small breaking change in `automerge v2.0.1` which changed how patches are emitted. They are now passed to the callback as an array of Patches, rather than individually. This had caused the test to fail.

This PR amends both the source and the test to match the new API in `automerge`.

Obviously this is a breaking change for repo too, but given that the change has already occurred upstream, I would imagine it's worth doing.

Two notes:
1. I have left the event name as the singular `patch`
2. `package.json` includes `@automerge/automerge` with a version of `*`. It might be worth restricting this to `>=2.0.1`

### DIFF
--- a/packages/automerge-repo/README.md
+++ b/packages/automerge-repo/README.md
@@ -66,7 +66,7 @@ A `DocHandle` also emits these events:
 - `change({handle: DocHandle})`  
   Called any time changes are created or received on the document. Request the `value()` from the
   handle.
-- `patch({handle: DocHandle, before: Doc, after: Doc, patch: Patch})`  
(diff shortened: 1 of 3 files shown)


pvh:
Thanks Alex, it looks good. I'll cut a release today. There's some other bug I hit yesterday which I want to look into first but I've merged it.

pvh:
PS, how would you like to be credited in the contributors?

acurrieclark:
That’s very kind of you. As Alex Currie-Clark, please

PR #23, opened 2023-02-27
Refactor

HerbCaudill:
This is still a huge grab-bag of many unrelated things and I'd be happy to break it into a set of smaller, sensible PRs. Putting in into a draft PR to facilitate discussion.


packages/automerge-repo-demo-counter/package.json
```diff
@@ -5,7 +5,6 @@
   "type": "module",
   "scripts": {
     "dev": "vite --open",
-    "build": "tsc && vite build",
```
pvh, line 8: Doesn't Vite build take care of this?

packages/automerge-repo/TODO.md
```diff
@@ -1,51 +1,54 @@
-TODO
```
pvh, line 1: ... why?

HerbCaudill: prettier

packages/automerge-repo/src/index.ts
```diff
@@ -1,24 +1,12 @@
...
+export * from "./network/NetworkAdapter"
+export * from "./network/NetworkSubsystem"
+export * from "./Repo"
+export * from "./storage/StorageAdapter"
+export * from "./storage/StorageSubsystem"
+export * from "./synchronizer/CollectionSynchronizer"
+export * from "./types"
+
+// Q: is there a reason this is exported with a different name?
+export { CollectionSynchronizer as DependencyCollectionSynchronizer } from "./synchronizer/CollectionSynchronizer"
```
pvh, line 12: we can just fix this -- this is a refactoring leftover.

also, let's keep the explicit exports -- it allows internal modules to export things for each other's use without leaking them into the public interface for the library

packages/automerge-repo/src/network/NetworkAdapter.ts
```diff
@@ -0,0 +1,54 @@
+import EventEmitter from "eventemitter3"
+import { PeerId, ChannelId } from "../types"
+
+export abstract class NetworkAdapter extends EventEmitter<NetworkAdapterEvents> {
+  peerId?: PeerId // hmmm, maybe not
```
pvh, line 5: this will probably move as we work on auth, hence the weird comment

packages/automerge-repo-network-messagechannel/src/index.ts
```diff
@@ -1,24 +1,14 @@
...
-import debug from "debug"
-import EventEmitter from "eventemitter3"
+import { ChannelId, NetworkAdapter, PeerId } from "automerge-repo"
 import { MessagePortRef } from "./MessagePortRef"
 import { StrongMessagePortRef } from "./StrongMessagePortRef"
 import { WeakMessagePortRef } from "./WeakMessagePortRef"
 
-const log = debug("messagechannel")
+import debug from "debug"
+const log = debug("ar:messagechannel")
```
pvh, line 7: this should be a full string: "automerge-repo"

packages/automerge-repo-react-hooks/src/RepoContext.ts
```diff
@@ -0,0 +1,4 @@
+import { Repo } from "automerge-repo"
```
pvh, line 1: I'd put this in with `useRepo` since that's what it's for and this is otherwise just one line.

packages/automerge-repo-storage-memory/test/index.test.ts
```diff
@@ -0,0 +1,5 @@
+import { expect } from "chai"
```
pvh, line 1: let's move this into a test helper -- you should never use this in real life. if you don't want to actually store things: you don't have to store them! (just initialize the object without storage)

packages/automerge-repo/src/helpers/pause.ts
```diff
@@ -0,0 +1,2 @@
+export const pause = (t = 100) =>
```
pvh, line 1: we don't use this with an argument anywhere -- we should make the default 0 or just not support having an argument.

even moreso (and maybe not in this PR) we should probably just... not do this at all.

packages/automerge-repo/src/network/NetworkSubsystem.ts
```diff
@@ -146,26 +87,42 @@ export class NetworkSubsystem extends EventEmitter<NetworkEvents> {
...
+      Object.entries(this.#adaptersByPeer).forEach(([id, peer]) => {
+        this.#log(`sending broadcast to ${id}`)
         peer.sendMessage(id as PeerId, channelId, message, true)
       })
     } else {
-      const peer = this.peerIdToAdapter[peerId]
+      const peer = this.#adaptersByPeer[peerId]
       if (!peer) {
-        log(`Tried to send message to disconnected peer: ${peerId}`)
-        return
```
pvh, line 156: this is probably a mistake to remove

packages/automerge-repo/src/storage/StorageAdapter.ts
```diff
@@ -0,0 +1,5 @@
+export abstract class StorageAdapter {
```
pvh, line 1: this is wrong but we don't have to fix it now

packages/automerge-repo/src/storage/StorageSubsystem.ts
```diff
@@ -1,97 +1,80 @@
...
-    mergedArray.set(item, offset)
-    offset += item.length
-  })
-
-  return mergedArray
-}
 export class StorageSubsystem {
-  storageAdapter: StorageAdapter
+  #storageAdapter: StorageAdapter
+  #changeCount: Record<DocumentId, number> = {}
```
pvh, line 8: yeah this approach is poor and should be replaced by something hash-based for better concurrency (later)

packages/automerge-repo/src/storage/StorageSubsystem.ts
```diff
@@ -1,97 +1,80 @@
...
+      this.#saveTotal(documentId, doc)
     } else {
-      this.saveIncremental(documentId, doc)
+      this.#saveIncremental(documentId, doc)
     }
   }
+
+  // TODO: make this, you know, good.
+  #shouldCompact(documentId: DocumentId) {
+    return this.#changeCount[documentId] >= 20
```
pvh, line 78: still ugh

pvh: Using change count is probably roughly the right metric here but the number of changes between compactions is going to be determined by the per-key overhead of the underlying storage adapter. A round-trip to S3 costs a lot more than grabbing the next object from a range query in IndexedDB or just tagging on new changes to the end of a POSIX file. 

Probably a good way to improve this would be to set the change count per compaction at the StorageAdapter level... but for now, this isn't causing any pain so we can leave it be while we deal with worse problems.

packages/automerge-repo/src/synchronizer/CollectionSynchronizer.ts
```diff
@@ -1,96 +1,109 @@
...
     this.repo = repo
   }
 
-  async onSyncMessage(
-    peerId: PeerId,
-    channelId: ChannelId,
-    message: Uint8Array
-  ) {
+  /**
+   *
```
pvh, line 26: write this comment

packages/automerge-repo/src/synchronizer/CollectionSynchronizer.ts
```diff
@@ -1,96 +1,109 @@
...
 
-  async fetchDocSynchronizer(documentId: DocumentId) {
-    // TODO: we want a callback to decide to accept offered documents
-    if (!this.syncPool[documentId]) {
-      const handle = await this.repo.find(documentId)
-      this.syncPool[documentId] =
-        this.syncPool[documentId] || this.initDocSynchronizer(handle)
+  /** Returns a synchronizer for the given document, creating one if it doesn't already exist.  */
+  fetchDocSynchronizer(documentId: DocumentId) {
+    // TODO: add a callback to decide whether or not to accept offered documents
```
pvh, line 45: coming soon to an authenticator near you

packages/automerge-repo/src/synchronizer/CollectionSynchronizer.ts
```diff
@@ -1,96 +1,109 @@
...
+
+    if (!this.#syncPool[documentId]) {
+      const handle = this.repo.find(documentId)
+      this.#syncPool[documentId] = this.initDocSynchronizer(handle)
```
pvh, line 49: this is broken: it always makes a new synchronizer

packages/automerge-repo/src/synchronizer/DocSynchronizer.ts
```diff
@@ -1,178 +1,156 @@
...
+  #sendSyncMessage(peerId: PeerId, doc: A.Doc<unknown>) {
+    this.#log(`sendSyncMessage ->${peerId}`)
 
-      conciseLog(`${peerId} ⬅️️  ${this.handle.documentId} (${message.byteLength} bytes)`)
-      if (opsLog.enabled) { // guard opsLog, so decodeChange is not called unnecessarily, because it can be expensive
-        opsLog(decoded.changes.flatMap(change => {
-          return decodeChange(change).ops.map((op) => JSON.stringify(op))
-        }))
-      }
+    // TODO: Right now this can send duplicate messages even if nothing has changed
```
pvh, line 66: i believe this was fixed in automerge upstraem

packages/automerge-repo/src/synchronizer/DocSynchronizer.ts
```diff
@@ -1,178 +1,156 @@
...
+
+    const syncState = this.#getSyncState(peerId)
+    const [newSyncState, message] = A.generateSyncMessage(doc, syncState)
+    this.#setSyncState(peerId, newSyncState)
+    if (message) {
+      this.logMessage(`sendSyncMessage 🡒 ${peerId}`, message)
```
pvh, line 72: →

packages/automerge-repo/src/synchronizer/DocSynchronizer.ts
```diff
@@ -1,178 +1,156 @@
...
-      )
-    )
+    this.#log(`beginSync: ${peerId}`)
+    const doc = await this.#handle.provisionalValue()
 
-    this.sendSyncMessage(peerId, documentId, doc)
+    // HACK: if we have a sync state already, we round-trip it through the encoding system to make
+    // sure state is preserved. This prevents an infinite loop caused by failed attempts to send
+    // messages during disconnection.
+    const syncStateRaw = this.#getSyncState(peerId)
```
pvh, line 116: we should save the sync states but only for peers which say they're likely to come back: ephemeral tabs in the browser for example shouldn't have their sync states saved because they can't reconnect, if they reload they come back empty with a new peerID

pvh: note that for ephemeral tabs we should support a read-only mode for storage. this would allow ephemeral processes to load off "disk" and then present themselves for sync already caught up to whatever the local state is.

packages/automerge-repo/test/helpers/DummyStorageAdapter.ts
```diff
@@ -0,0 +1,19 @@
+import { DocumentId, StorageAdapter } from "../../src"
```
pvh, line 1: great! now we can remove the memory adapter (remember we should also remove the NPM module)

packages/automerge-repo/test/DocSynchronizer.test.ts
```diff
@@ -1,52 +1,70 @@
 import assert from "assert"
-import { DocHandle, DocumentId } from "../src/DocHandle"
+import { DocumentId, PeerId } from "../src/types"
+import { DocHandle } from "../src/DocHandle"
 import { DocSynchronizer } from "../src/synchronizer/DocSynchronizer"
-import { PeerId } from "../src/network/NetworkSubsystem"
+import { eventPromise } from "../src/helpers/eventPromise"
+
+type TestDoc = {
```
pvh, line 7: you made a type for this in its own file and everything

packages/automerge-repo/src/DocCollection.ts
```diff
@@ -1,58 +1,105 @@
...
+    this.#handleCache[documentId] = handle
     return handle
   }
 
-  // TODO: this should really insist on initial value of T
-  // (but: we need to make sure the storage system will collect it)
-  // (next: we need to have some kind of reify function)
-  // (then: cambria!)
+  /** Returns all the handles we have cached. */
+  get handles() {
```
pvh, line 35: remove

packages/automerge-repo/src/DocHandle.ts
```diff
@@ -1,181 +1,211 @@
...
-      this.emit("syncing")
+  /**
+   * A Repo can call this when it doesn't have the document and has advertised our interest in it.
+   * This blocks access to the document until we get it from a peer.
+   *
+   * TODO: might be good for this to timeout and go to a "not found" state if the document isn't
+   * available after a certain amount of time. but not sure what we would do with a doc in that
+   * state. We'd also need to retry etc.
+   */
+  waitForSync() {
```
pvh, line 84: what should this be called? it doesn't actually do any waiting...
maybe `requestSync()`? 

this is kind of messy and confusing... let's move the conditional up

packages/automerge-repo/src/DocHandle.ts
```diff
@@ -1,181 +1,211 @@
...
   }
 
-  change(callback: (doc: T) => void, options: ChangeOptions<T> = {}) {
-    this.value().then(() => {
-      const newDoc = Automerge.change<T>(this.doc, options, callback)
-      this.__notifyChangeListeners(newDoc)
-    })
+  /** Applies an Automerge change function to the document. */
+  async change(callback: A.ChangeFn<T>, options: ChangeOptions<T> = {}) {
+    await this.value()
```
pvh, line 167: we should note that this is blocking to make sure you don't call change() before you get an initial value by accident

packages/automerge-repo-demo-counter/package.json
HerbCaudill: apparently not — this is from the starter site that `yarn create vite` generates.

packages/automerge-repo-network-messagechannel/src/index.ts
HerbCaudill: ✅

packages/automerge-repo-storage-memory/test/index.test.ts
HerbCaudill: ✅

packages/automerge-repo/src/network/NetworkSubsystem.ts
HerbCaudill: ✅

packages/automerge-repo/src/index.ts
HerbCaudill: ✅

packages/automerge-repo/src/helpers/pause.ts
HerbCaudill: ✅

packages/automerge-repo/src/DocCollection.ts
HerbCaudill: this is used by tests

HerbCaudill:
Let us all now forget this trainwreck of a PR ever happened. 😅 I promise to behave in the future.

PR #39, opened 2023-03-16
Fix timeout implementation in DocHandle

HerbCaudill:
With `timeoutDelay` set to 700,000 ms 🤔 the tests were never finishing on their own, because the timeouts from `pause(this.#timeoutDelay)` were never cleared even if all tests completed. 

I've added a `withTimeout(p,t)` helper that takes a promise and a timeout. If the promise is resolved before the timeout ends, the timeout is cleared. If the timeout ends first, a `TimeoutError` is thrown. 

Since `DocHandle.value` and `DocHandle.loadAttemptedValue` only differ in the set of states they wait for, I've refactored the two so that the common logic isn't duplicated: 

```ts
  async value(awaitStates: HandleState[] = [READY]) {
      // ...
  }

  async loadAttemptedValue() {
    return this.value([READY, REQUESTING])
  }
```

### DIFF
--- a/packages/automerge-repo/src/DocHandle.ts
+++ b/packages/automerge-repo/src/DocHandle.ts
@@ -139,8 +139,12 @@ export class DocHandle<T> //
     return this.#machine?.getSnapshot().value
   }
 
-  #statePromise(state: HandleState) {
-    return waitFor(this.#machine, s => s.matches(state))
+  /** Returns a promise that resolves when the docHandle is in one of the given states */
+  #statePromise(awaitStates: HandleState | HandleState[]) {
+    if (!Array.isArray(awaitStates)) awaitStates = [awaitStates]
+    return Promise.any(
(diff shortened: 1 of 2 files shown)


pvh:
Ah yes -- thanks. The timeout thing is a problem though because when you're loading a large corpus, many handles get enqueued and have to wait a long time before their data can be fetched. This is probably an area we need to spend some design energy on: how to avoid head-of-line blocking, as well as probably a bunch of actual optimization. (Downloading a few megs shouldn't take long.)

HerbCaudill:
Ahhh ok. Well in that case setting the default high is probably a good call.

acurrieclark:
Is it worth exposing the `timeout` value as an option somewhere? Either as an optional parameter on the `value` call or globally when setting up the repo?

acurrieclark:
Incidentally, an error popped up today which revealed that XState's `waitFor` has a default timeout of 10,000ms. It can be overridden by passing in an options object.

PR #40, opened 2023-03-19
Auth provider

### LINKED ISSUE #25: Support authentication & authorization
What follows is a sketch of a proposed approach for supporting authentication & authorization within automerge-repo. 

### Why

Automerge Repo currently offers no way to authenticate a peer, and very little in the way of access control.

Our current security model is the "Rumplestiltskin rule": If you know a document's ID, you can read that document, and everyone else who knows that ID will accept your changes.

That model is good enough for a surprising number of situations — the ID serves as an unguessable secret "password" for the document — but it has limitations. Without a way to establish a peer's identity, we can't revoke access for an individual peer — say if someone leaves a team, or if a device is lost. And we can't distinguish between read and write permissions, or limit access to specific documents.

An application might implement authentication and authorization in any number of ways, so this should be pluggable — like the existing network and storage adapters.

So initializing a repo might look something like this:

```ts
import { SuperCoolAuthProvider } from 'supercool-auth-library'

const authOptions = {
  // ...options specific to this type of authentication
}
const auth = new SuperCoolAuthProvider(options)

const repo = new Repo({ network, storage, auth })
```

### API

An auth provider inherits from this abstract class:

```ts
export abstract class AuthProvider extends EventEmitter<AuthProviderEvents> {
  /**
   * Can this peer prove their identity? The provider implementation will
   * use the web socket to communicate with the peer.
   */
  abstract async authenticate(peerId: PeerId, socket?: WebSocket): Promise<true | Error> {}

  // The following methods may be overriden by the provider and would replace
  // the existing `sharePolicy` that we pass to a repo.

  /** Should we tell this peer about the existence of this document? */
  async okToAdvertise(peerId: PeerId, documentId: DocumentId): Promise<boolean> {
    return false
  }

  /** Should we provide this document (and changes to it) to this peer when asked for it by ID? */
  async okToSend(peerId: PeerId, documentId: DocumentId): Promise<boolean> {
    return false
  }

  /** Should we accept changes to this document from this peer? */
  async okToReceive(peerId: PeerId, documentId: DocumentId): Promise<boolean> {
    return false
  }
}
```

### Authentication

The auth provider's `authenticate` method is invoked when the network adapter emits a
`peer-candidate` event.

```ts
// NetworkSubsystem.ts
networkAdapter.on('peer-candidate', async ({ peerId, channelId, socket }) => {
  const { authenticationResult } = await authProvider.authenticate({ peerId, socket })
  if (!authenticationResult.isValid) {
    const { error } = authenticationResult
    this.emit('peer-authentication-failed', { peerId, channelId, error })
  } else {
    // ...

    this.emit('peer', { peerId, channelId })
  }
})
```

#### Advertising

TODO

#### Sending changes

TODO

#### Receiving changes

TODO

HerbCaudill:
Fixes #25 

This is a first stab at an AuthProvider API. This has evolved a bit from the initial proposal in #25. 

- `AuthProvider` is not an abstract class, but a base implementation that is maximally generous. 
- `okToReceive` has been removed[^1], and `okToSend` has been renamed to `okToSync`.
-  A `transform` property has been added, consisting of `inbound` and `outbound` message transformers. This allows an AuthProvider to encrypt subsequent traffic as part of its authentication scheme. 

### Usage

These examples are all present in the tests. 

#### Adding an `AuthProvider` to a repo

```ts
const permissive = new AuthProvider() // AuthProvider is maximally permissive by default

const aliceRepo = new Repo({
  network: [new MessageChannelNetworkAdapter(aliceToBob)],
  peerId: "alice" as PeerId,
  authProvider: permissive
})
```
 
#### Restricting access by peer

```ts
const authProvider = new AuthProvider({
  okToAdvertise: async (peerId: PeerId) => peerId !== "peer1",
})
``` 

#### Restricting access by document

```ts
const handle = collection.create()
const authProvider = new AuthProvider({
  okToAdvertise: async (_, documentId) => documentId !== handle.documentId,
})
```

#### Restricting access by peer and document

```ts
class NotForCharlieAuthProvider extends AuthProvider {
  excludedDocs: DocumentId[] = []

  // make sure that charlie never learns about excluded documents
  notForCharlie: SharePolicy = async (peerId, documentId) => {
    if (this.excludedDocs.includes(documentId!) && peerId === "charlie")
      return false
    return true
  }

  okToAdvertise = this.notForCharlie
  okToSync = this.notForCharlie
}

const authProvider = new NotForCharlieAuthProvider()

const notForCharlieHandle = aliceRepo.create<TestDoc>()
const notForCharlie = notForCharlieHandle.documentId
authProvider.excludedDocs.push(notForCharlie)

```

#### Maximally restrictive AuthProvider

```ts
const restrictive = new AuthProvider({
  authenticate: async () => {
    return {
      isValid: false,
      error: new Error("nope"),
    }
  },
  okToAdvertise: async () => false,
  okToSync: async () => false,
})
```

In the above example, the provider is customized by passing  `authenticate` etc. in a config object to the base class's constructor. Alternatively, a reusable custom provider can be created by extending the base class: 

```ts
class RestrictiveAuthProvider extends AuthProvider {
  authenticate = async () => {
    return {
      isValid: false,
      error: new Error("nope"),
    }
  }
  okToAdvertise = async () => false
  okToSync = async () => false
}
const restrictive = new RestrictiveAuthProvider()
``` 

#### Using network communication

We'll make a (very insecure) password auth provider that sends a password challenge, and compares the password returned to a hard-coded password list.

```ts
class PasswordAuthProvider extends AuthProvider {
  // The auth provider is initialized with a password response, which it will provide when
  // challenged.
  constructor(private passwordResponse: string) {
    super()
  }

  #challenge = "what is the password?"

  #passwords: Record<string, string> = {
    alice: "abracadabra",
    bob: "bucaramanga",
  }

  authenticate: AuthenticateFn = async (peerId, channel) => {
    return new Promise<AuthenticationResult>(resolve => {
      // send challenge
      channel.send(new TextEncoder().encode(this.#challenge))

      channel.on("message", msg => {
        const msgText = new TextDecoder().decode(msg)
        switch (msgText) {
          case this.#challenge:
            // received challenge, send password
            channel.send(new TextEncoder().encode(this.passwordResponse))
            break

          case this.#passwords[peerId]:
            // received correct password
            resolve(AUTHENTICATION_VALID)
            break

          default:
            // received incorrect password
            resolve(authenticationError("that is not the password"))
            break
        }
      })
    })
  }
}
``` 

#### Transforming network traffic

The idea here is that rather than authenticate peers, the auth provider encrypts and decrypts messages using a secret key that each peer knows. No keys are revealed, but the peers can only communicate if they know the secret key.

```ts
function encryptingAuthProvider(secretKey: string) {
  return new AuthProvider({
    transform: {
      inbound: payload => {
        const decrypted = decrypt(payload.message, secretKey)
        return { ...payload, message: decrypted }
      },
      outbound: payload => {
        const encrypted = encrypt(payload.message, secretKey)
        return { ...payload, message: encrypted }
      },
    },
  })
}
``` 

[^1]: `okToReceive` would have misled developers to think that you could enforce write permissions that way; but the peer who shares document changes with you isn't necessarily the peer who made the changes. Write permissions would need to be implemented some other way — for example, the way [crdx](https://github.com/HerbCaudill/crdx) does, with cryptographic proof of authorship embedded into the CRDT.


HerbCaudill:
I'm closing this for now; I'll rebase on latest code and resubmit.

PR #41, opened 2023-03-19
Custom UUID

HerbCaudill:
Allows passing an idGenerator function to the Repo constructor, so I can use e.g. [nanoid](https://github.com/ai/nanoid).


HerbCaudill:
Hey, guys — I just realized that this was still marked as draft. LMK if there's anything that needs to happen before we can merge this.

acurrieclark:
This makes good sense to me. The only thing which crossed my mind was that, as this is used to create `DocumentId`s and as (in the current implementation) ephemeral messages are "detected" by prepending `m/` to the `ChannelId`, there could be sync messages lost if the generated ID happens to start with `m/`.

I know that `nanoid` doesn't have `/` in its standard alphabet, but should we get `ChannelId` removed first?

PR #42, opened 2023-04-08
Added missing `.js` extension to a few more imports

acurrieclark:
There is also the option of setting `"moduleResolution": "nodenext",` in `tsconfig` which would help enforce this, but it seems to also break some of the type imports for me in VSCode.

### DIFF
--- a/packages/automerge-repo/src/DocHandle.ts
+++ b/packages/automerge-repo/src/DocHandle.ts
@@ -13,10 +13,10 @@ import {
   StateValue,
   TypegenDisabled,
 } from "xstate"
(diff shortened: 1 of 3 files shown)


pvh:
@HerbCaudill I thought we were getting rid of these. Did I mix something up?

HerbCaudill:
No, you need the js extension in all imports for an esm library to be used in node.

pvh:
I guess I must have read the patch backwards. My impression is that you can't load automerge-repo in node at all, because it's written in Typescripts, so you're going to have to transpile the thing first anyway.

PR #45, opened 2023-04-12
Add useLocalAwareness & useRemoteAwareness hook

neftaly:
This PR adds hooks for ephemeral shared state, similar to [Yjs awareness](https://docs.yjs.dev/api/about-awareness).
WIP, has not been converted to TS.


neftaly:
I'm not sure where to put `use-awareness-example-project`. Should this be a new package in the monorepo, separate from automerge-repo-react-hooks?

Also, does it need tests? There aren't any for the other hooks.

packages/automerge-repo-sync-server/package.json
```diff
@@ -5,7 +5,7 @@
   "license": "MIT",
   "type": "module",
   "scripts": {
-    "start": "node ./src/index.js"
+    "start": "set DEBUG=* && node ./src/index.js"
```
neftaly, line 8: whoops

neftaly:
Note that these hooks are only designed to be used once per channel. Can hooks be memoized?

packages/automerge-repo-sync-server/package.json
pvh: would you mind just removing this from the PR? i'm happy to merge if so. given your broadcastchannel note we should probably also do something more sensible to reduce "echo" but i don't want to block on that.

neftaly: Sure, fixed. Making an issue for broadcastchannel/

neftaly: Sure. Created #50

neftaly:
@pvh do you need anything else to get this merged?

neftaly:
I have found the current API of this quite useful, however there are a couple of shortcomings with the current behavior. Specifically, you can currently only have one instance of useLocalAwareness and useRemoteAwareness each per channel ID. 

This makes sharing data from different locations difficult. The way around this is to pass data & an event handler down to child functions as a prop. It is likely possible to fix this in useRemoteAwareness short-term (though I have no current plans to do so). Fixing this in useLocalAwareness would involve something more like global state management ala Zustand.

I intend to fix this longer-term by making awareness use an underlying Flyd stream with a state reducer. It would be frontend agnostic, and could then be wrapped in a React hooks library with the same API as the current hooks. 

A separate minor DX issue I have is that channel IDs must be specified in order for local & remote hooks to talk to each other. I would like to automate this somehow, and remove the possibility of accidentally using the wrong channel ID, or worse, re-using the same channel ID with different awareness instances that shouldn't be talking to each other.

neftaly:
Per @pvh's comments on Slack, we should avoid using documentIDs as a basis for channelIDs, and opt for user-level presence rather than per-repo. This can be fixed in the example in future; the autoburger article and docs should also reflect this.

packages/automerge-repo-react-hooks/package.json
```diff
@@ -13,7 +13,10 @@
     "watch": "npm-watch"
   },
   "dependencies": {
-    "@automerge/automerge-repo": "^0.1.2"
+    "automerge-repo": "^0.1.2",
+    "eventemitter3": "^5.0.0",
+    "react": "^18.2.0",
```
neftaly, line 18: do you want this to be a peer dep? otherwise good to go

neftaly:
I am thinking of wrapping the hook args in a single `{}` props object, to make future changes easier

neftaly:
I would like to re-write the example after useBootstrap lands.

PR #54, opened 2023-04-24
Delete documents

### LINKED ISSUE #48: Removing repo documents
I wanted to start a discussion on how we might go about removing a document/handle from a repo.

It seems to me that there are 2 stages to removing a document:

1. A document can be locally "forgotten". This is relatively straightforward, as it would simply involve removing the document from storage and deleting it from the `DocCollection` cache.
2. We can request for the document to be removed from one or more peers. This is obviously a more complex scenario, as both users and developers will want differing levels of control. There is also the matter of how delete requests might be propagated amongst peers.

I would like to propose that we implement stage 1 of the above. The mechanism of locally forgetting a document is useful as a standalone option and would also ease the implementation of stage 2, when we settled on a process.

Happy to put together a `forget` PR. If there are any thoughts on naming the API, please let me know.

acurrieclark:
This adds a `delete` method to a repository which:

- Removes a handle from the cache
- Removes the underlying document data from storage
- Emits a `delete-document` event from the repo
- Emits a `delete` event from the handle
- Sets the internal handle state to `DELETED`
- closes #48 

Usage is `repo.delete(documentId)`.

At the moment, this makes no effort to communicate the deletion to the network.

Please let me know your thoughts and I will be sure to add documentation when we are happy with the API

### DIFF
--- a/packages/automerge-repo/README.md
+++ b/packages/automerge-repo/README.md
@@ -48,16 +48,22 @@ A `Repo` exposes these methods:
 - `find<T>(docId: DocumentId)`  
   Looks up a given document either on the local machine or (if necessary) over any configured
   networks.
+- `delete(docId: DocumentId)`  
+  Deletes the local copy of a document from the local cache and local storage. _This does not currently delete the document from any other peers_.
 - `.on("document", ({handle: DocHandle}) => void)`  
+  Registers a callback to be fired each time a new document is loaded or created.
+- `.on("delete-document", ({handle: DocHandle}) => void)`  
   Registers a callback to be fired each time a new document is loaded or created.
 
 A `DocHandle` is a wrapper around an `Automerge.Doc`. Its primary function is to dispatch changes to
 the document.
 
-- `handle.change((doc: T) => void)` Calls the provided callback with an instrumented mutable object
+- `handle.change((doc: T) => void)`  
+  Calls the provided callback with an instrumented mutable object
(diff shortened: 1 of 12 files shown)


pvh:
This looks great. What should happen if a user tries to find() that document again? I would posit that it should resynchronize, or at least that there should be *some* way of getting out of this state at runtime. (Restarting the process would have the desired effect.)

Even if we don't agree on what that behaviour should be, I think it should be tested / documented. I'm comfortable with going for a different behaviour (and awaiting user input) as long as we commit to something.

acurrieclark:
My feeling on this is that an explicit `find` call should indeed try to reload the document from the network.

As we have previously noted, if another repository connects which has an open `sharePolicy`, then it's documents will automatically be synced back. I would suggest that this would not be the expected _automatic_ behaviour.

I think this is another good excuse to set the default `sharePolicy` to return false.

pvh:
We should definitely take this patch. Have you got the energy to carry on thinking this through a bit? What behaviour would you want for your app?

acurrieclark:
The API I described above would certainly be useful, even if I only used it locally in the short term.

However, with this PR I can achieve everything I need for now. It at least provides a method to manage any deletion outside of repo.

Very happy to keep thinking about this.

pvh:
Cool, happy to take this approach. I do think that if you have a backup server then adding support for a deletion message would be a good idea.

PR #55, opened 2023-04-25
Fix: WebSocket Adapter

acurrieclark:
The message type had been commented out on the client sync message, preventing any docs from syncing


acurrieclark:
This certainly needs fixing, but it seems as though there might be another issue to surface here. Marking as draft while I put a test together

PR #56, opened 2023-04-25
Add Tests Action

acurrieclark:
Thought it might be useful to have a basic github action set up to run any tests on PRs and pushes to main.

Would require actions to be enabled in settings.


pvh (approved):
SGTM!

PR #59, opened 2023-04-27
Proposal: New signature for handle `change` event

### LINKED ISSUE #43: Consecutive changes do not emit distinct documents
When a handle is changed, the `change` event emits as expected. We can then resolve the handle value to get underlying document.

However, if two changes are made in quick succession, only the final value is resolved:

```typescript
const TEST_ID = "test-document-id" as DocumentId
const handle = new DocHandle<{foo: string}>(TEST_ID, { isNew: true });

handle.on("change", async ({ handle }) => {
  const d = await handle.value();
  console.log(d);
});

handle.change((doc) => {
  doc.foo = "bar";
});

handle.change((doc) => {
  doc.foo = "baz";
});
```

will log
```
{ foo: 'baz' } <-- should be 'bar'
{ foo: 'baz' }
```

A simple solution to this would be to include the document in the event parameter object? Happy to put together a PR if this is acceptable.

acurrieclark:
#43 outlines the case where successive, quick changes to a handle do not currently result in distinct documents being returned in the `change` event closure.

This proposal moves the `change` event trigger to the document `patchCallback` (which is only triggered when a change causes the document `heads` to update).

The additional detail provided in the callback is forwarded on as a property of the `change` event closure argument.

If anyone knows of a way to mark the `patch` event as `@deprecated` so that it offers some guidance to editors, that might be useful.

As is stands, this **does not represent a breaking change**.

I have kept the property names the same as those used in the `patchCallback`, but I wondered if it might make more sense in our context to use `doc` (in place of `after`) and `oldDoc` (for `before`).

This would close #43

### DIFF
--- a/packages/automerge-repo-network-messagechannel/package.json
+++ b/packages/automerge-repo-network-messagechannel/package.json
@@ -23,5 +23,8 @@
         ".ts"
       ]
     }
+  },
+  "devDependencies": {
+    "@types/chai": "^4.3.4"
   }
 }
--- a/packages/automerge-repo-network-messagechannel/test/index.test.ts
+++ b/packages/automerge-repo-network-messagechannel/test/index.test.ts
@@ -1,6 +1,8 @@
-import { expect } from "chai"
-import { ChannelId, PeerId } from "automerge-repo"
+import { assert, expect } from "chai"
+import { ChannelId, PeerId, Repo } from "automerge-repo"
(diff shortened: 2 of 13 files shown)


acurrieclark:
There is [an issue](https://github.com/automerge/automerge/issues/593) with the `after` property in the `PatchCallback` which needs to be resolved before this could be used properly.

acurrieclark:
Closing as this has been superseded by #73

PR #63, opened 2023-04-28
Fix incremental save following an incremental load

acurrieclark:
Related to https://github.com/automerge/automerge/issues/592

Following a load of a document from storage, then next incremental save is of the whole document. In Automerge `2.0.2` this would throw an error. While the error no longer exists in `2.0.3` we are still bloating the local store unnecessarily.

This ensures that all incremental saves are indeed incremental.


pvh:
landed, thanks!

PR #64, opened 2023-04-28
Omit `PatchCallback` from `ChangeOptions`

acurrieclark:
If a `patchCallback` is passed to the `DocHandle` `change` method, then that callback overwrites the one created on `init`. This results in no more `patch` events being emitted.

At the moment this is just a typescript update, but it might be prudent to throw an error here, or at least log a warning to the console?

### DIFF
--- a/packages/automerge-repo/src/DocHandle.ts
+++ b/packages/automerge-repo/src/DocHandle.ts
@@ -210,7 +210,10 @@ export class DocHandle<T> //
   }
 
   /** `change` is called by the repo when the document is changed locally  */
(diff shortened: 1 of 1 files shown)


pvh:
Oof, yes. @orionz this seems relevant to your interests. What do you think the best behaviour here would be? Throw an error? Refuse to proceed? Hm.

pvh:
Just a note on this PR that @alexjg is working on an API update to do away with this system in the Rust code entirely as we move to arbitrary point-to-point patches (which will make many things, like live diffing much easier!)

I'm going to close this one since it just introduces a little complexity that we'd have to unpick later even though it is otherwise a good idea.

PR #68, opened 2023-05-03
Allow documents to be created with a fixed ID and initial schema

ecstatic-morse:
For context, I'm playing around with `automerge` for a side-project, and I need to store per-user data with a defined schema. The easiest way to accomplish this seems to be a combination of the trick in [the cookbook](https://automerge.org/docs/cookbook/modeling-data/#setting-up-an-initial-document-structure) with a fixed `documentId`. This seems like a common pattern, so I think some way of accomplishing it should be upstreamed. Here's what I've been using locally.


packages/automerge-repo/src/DocCollection.ts
```diff
@@ -44,28 +43,25 @@ export class DocCollection extends EventEmitter<DocCollectionEvents> {
...
-    //   return { ...myInitialValue, ...doc }
-    // }
-
-    // or
-    // - pass a "reify" function that takes a `<any>` and returns `<T>`
-
-    const documentId = uuid() as DocumentId
-    const handle = this.#getHandle<T>(documentId, true) as DocHandle<T>
+  create<T>(opts?: {
+    schema?: T & object
```
ecstatic-morse, line 56: AFAICT `automerge` expects the root of a document to be map-like, but this expectation isn't encoded into the type system for it or `automerge-repo` (e.g. with `DocHandle<T extends object>`). However, knowing that `schema` is an `object` makes the deep copy below a bit nicer.

acurrieclark:
This looks like a sound implementation to me. I am hesitant to review this alone as I know that there has been some thought put in to this before I joined.

Can I ask what the reasoning behind the fixed `DocumentId` is? Is the idea to have something which could be made unique to each user to avoid a clash? Just for my own interest.

@pvh what do you think?

pvh:
No, you should never reuse document IDs. When you merge changes from
another user, the root object (and everything below it) will conflict and
you will only have one document or the other (perhaps even a frankenstein
of both!) after a merge. This is why there is no affordance to provide a
document ID.

If you really want to reuse the document, you could create a document
during initialization and then distribute it as part of the application's
bundle, but I would personally probably not do that.

ecstatic-morse:
@pvh This PR handles the conflict problem by hard-coding the actor ID and timestamp for the initial change—as [described in the guide](https://automerge.org/docs/cookbook/modeling-data/#setting-up-an-initial-document-structure)—and using a stable iteration order over the object that defines the initial schema (instead of initializing each field by hand). 

I understand that Cambria solves this in a more robust way, but I would like to use the sync protocol in `automerge-repo` and as I explained above I need a globally unique document with a fixed schema. Is there a better way to accomplish this?

pvh:
Hey, I'm back from my vacation and I'd like to talk more about this. I suspect the best approach is just to save a starter document and load it into the repo on first boot... but maybe that wouldn't work for reasons I don't understand? Let's discuss live outside of this issue -- can you ping me on the Slack?

pvh:
I'm clearing out the backlog of PRs here and while I agree with the problem statement here, I don't think this is the way I'd like to support it in core. (Though it's certainly fine to do if it works for your app!) I'm going to close this one but if you swing back around to this stuff let me know and I'd be happy to iterate on the problem.

PR #69, opened 2023-05-10
Disallow floating promises

ecstatic-morse:
Ensure that all promises are either awaited or are explicitly detached (with `void`).

While working on an alternate version of #68, I found quite a few places where async functions were called without awaiting their result, making it easy to accidentally run things in parallel. Also, it was sometimes difficult to tell what the  `Promise<void>` returned by some of the internal async functions signified. For example, `CollectionSynchronizer.addDocument` returns a promise that resolves when the `sharePolicy`s for all its peers is known, it has nothing to do with when the document is synced. Finally some internal functions were `async` despite never awaiting anything.

This PR fixes the above issues and updates some tests so that `eslint` passes.

### DIFF
--- a/packages/automerge-repo/src/synchronizer/CollectionSynchronizer.ts
+++ b/packages/automerge-repo/src/synchronizer/CollectionSynchronizer.ts
@@ -20,7 +20,7 @@ export class CollectionSynchronizer extends Synchronizer {
   }
 
   /** Returns a synchronizer for the given document, creating one if it doesn't already exist.  */
-  async #fetchDocSynchronizer(documentId: DocumentId) {
+  #fetchDocSynchronizer(documentId: DocumentId) {
     if (!this.#docSynchronizers[documentId]) {
       const handle = this.repo.find(documentId)
       this.#docSynchronizers[documentId] = this.#initDocSynchronizer(handle)
@@ -62,7 +62,7 @@ export class CollectionSynchronizer extends Synchronizer {
     const documentId = channelId as unknown as DocumentId
     const docSynchronizer = await this.#fetchDocSynchronizer(documentId)
 
-    docSynchronizer.receiveSyncMessage(peerId, channelId, message)
+    await docSynchronizer.receiveSyncMessage(peerId, channelId, message)
 
     // Initiate sync with any new peers
     const peers = await this.#documentGenerousPeers(documentId)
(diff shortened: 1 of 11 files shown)


packages/automerge-repo/src/synchronizer/CollectionSynchronizer.ts
```diff
@@ -62,7 +62,7 @@ export class CollectionSynchronizer extends Synchronizer {
     const documentId = channelId as unknown as DocumentId
     const docSynchronizer = await this.#fetchDocSynchronizer(documentId)
 
-    docSynchronizer.receiveSyncMessage(peerId, channelId, message)
+    await docSynchronizer.receiveSyncMessage(peerId, channelId, message)
 
     // Initiate sync with any new peers
     const peers = await this.#documentGenerousPeers(documentId)
```
ecstatic-morse, line 68: Before, this ran in parallel with `receiveSyncMessage` above, which was probably not intended.

acurrieclark:
Thanks for your continued work on this, Dylan. I will try to look at this shortly.

Can I just check, are any of the changes to the tests due to the alterations you have made to the async functions, or are they purely because of including them in eslint?

ecstatic-morse:
@acurrieclark the test changes are indeed meant to eliminate `no-floating-promises` errors in the tests. I tried to preserve the intent of the test rather than just adding `void` everywhere, unfortunately it makes the diff a bit noisy.

Also it looks like the error comes from `.ts` files in `test` being included in `tsconfig.json`. I'll try to figure out what's going on. Maybe a missing `dev-dependency`?

ecstatic-morse:
So currently a few of the tests in `automerge-repo` tests depend on workspace packages (which in turn depend on `automerge-repo`), but these dependencies aren't listed in `package.json`, they just happen to always be present in the workspace-global `node_modules` once it is built. Adding them to `devDependencies` creates a cyclic dependency which causes `lerna run build` to fail on CI with a weird error message (it succeeds locally).

I guess the solution is to move the integration tests to another package to break the cycle.

acurrieclark:
I assume these are only showing up now because the test files have been added to the `tsconfig.json` includes?

ecstatic-morse:
@acurrieclark that's correct. You can omit them from both `eslint` and `tsc` as before but...you probably want to be type-checking your tests?

pvh:
Hi! I'm back from my vacation and catching up on PRs and cutting new releases. This patch looks great -- thanks so much for contributing it. Some of the gratuitous async tags were the result of historical versions of the API where these functions were async for one reason or another but I think in general less asynchrony is better. I'm going to merge this though I suspect there may be some breakage in blutack.

pvh:
Merged & pushed, thanks!

PR #71, opened 2023-05-10
Buffer sync messages until the DocHandle is ready

### LINKED ISSUE #70: Implementation of `DocSynchronizer.receiveSyncMessage` may cause messages to be received out of order
While writing #69 , I noticed that `receiveSyncMessage` waits for the doc handle to be ready before calling `update`

https://github.com/automerge/automerge-repo/blob/ee508301f84339cad316bd4fb68216397c72bd51/packages/automerge-repo/src/synchronizer/DocSynchronizer.ts#L135-L160

`receiveSyncMessage` is invoked (indirectly through `CollectionSynchronizer`) in an `on("message", ...)` listener in `Repo.ts`:

https://github.com/automerge/automerge-repo/blob/ee508301f84339cad316bd4fb68216397c72bd51/packages/automerge-repo/src/Repo.ts#L98-L112

However, if we get two messages before doc handle is ready, I don't think there's any guarantee that the two listeners will resume from the `await` in the same order they reached it, which could cause sync messages to be processed out of order. I haven't reproduced this yet, so maybe I'm wrong about the spec and someone can correct me.

ecstatic-morse:
Fixes #70

I still haven't observed the bug described in #70—it's still theoretical—but this removes a `TODO` at least :smile:.

### DIFF
--- a/packages/automerge-repo/src/synchronizer/DocSynchronizer.ts
+++ b/packages/automerge-repo/src/synchronizer/DocSynchronizer.ts
@@ -178,5 +178,7 @@ export class DocSynchronizer extends Synchronizer {
       for (const { peerId, message } of this.#pendingSyncMessages) {
         this.#processSyncMessage(peerId, message)
       }
+
+      this.#pendingSyncMessages = []
   }
 }
--- a/packages/automerge-repo/src/synchronizer/DocSynchronizer.ts
+++ b/packages/automerge-repo/src/synchronizer/DocSynchronizer.ts
@@ -20,6 +20,8 @@ export class DocSynchronizer extends Synchronizer {
   /** Sync state for each peer we've communicated with (including inactive peers) */
   #syncStates: Record<PeerId, A.SyncState> = {}
 
+  #pendingSyncMessages: Array<{peerId: PeerId, message: Uint8Array}> = []
+
   constructor(private handle: DocHandle<any>) {
     super()
     const docId = handle.documentId.slice(0, 5)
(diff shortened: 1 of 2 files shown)


ecstatic-morse:
The tests passed without [6e80415](https://github.com/automerge/automerge-repo/pull/71/commits/6e80415027809fccea650ba8ff7da6f5e445a333), which means this code isn't run in any of the tests. It might be best to come up with a test before merging this. Up to you.

packages/automerge-repo/src/synchronizer/DocSynchronizer.ts
```diff
@@ -158,4 +173,12 @@ export class DocSynchronizer extends Synchronizer {
       return newDoc
```
ecstatic-morse, line 173: Unrelated, but why doesn't the preceding call to `this.#sendSyncMessage` take `newDoc` instead of `doc`? Don't we want to sync the latest changes?

acurrieclark: Interesting. Do you mean the other way around? Looks like it should take `newDoc` to me

acurrieclark: I can see why this would still work, as `newDoc` only differs from `doc` with the changes from the peer, which of course the peer already has. Still doesn't look quite right to me though.

pvh:
I was on the fence with this one but decided to accept a little extra complexity here.

PR #72, opened 2023-05-13
Public API for sync access to a handle's doc

acurrieclark:
This offers simple access to a `DocHandle`'s underlying document. There is currently no check made as to whether the document is ready, so this could return an empty doc.


pvh (approved):

pvh:
One small comment @acurrieclark is that if you use await, you make your whole function async, so the example code would be better off maybe using .then() syntax or something.

PR #74, opened 2023-05-13
Makes `change` synchronous and prevents changes during `requesting` state

pvh:
As described -- we might want not to change the change signature but I think it's relatively safe?


acurrieclark (approved):
This looks exactly as it should be to me. I see no reason for now why `change` should be async.

PR #77, opened 2023-05-30
Ensure `DocHandle` state is updated before change event is emitted

acurrieclark:
In [this discussion on Slack](https://automerge.slack.com/archives/C61RJCM9S/p1685460896830499), @jamesgpearce noticed that checking the sync `handle.doc` within a change event callback would return the previous doc version, as the underlying document is not updated until the state event has returned the new value.

This PR alters the point at which a change event is emitted, by checking the document for changes after each state transition first.

### DIFF
--- a/packages/automerge-repo/src/DocHandle.ts
+++ b/packages/automerge-repo/src/DocHandle.ts
@@ -121,13 +121,6 @@ export class DocHandle<T> //
               const { callback } = payload
               const newDoc = callback(oldDoc)
 
-              const docChanged = !headsAreSame(newDoc, oldDoc)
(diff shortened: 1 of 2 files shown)


jamesgpearce:
Well, for what it's worth, it looks good to me!

pvh:
We talked a bit on Slack about why I want to look more closely at this before merging... Basically, there are some race conditions and sensitivities around when various events are emitted and this is changing that flow. The change is probably for the better, but I need to look carefully at the invariants and make sure it's still working as expected.

pvh:
This is merged! (I don't know why Github doesn't realize yet.)

PR #78, opened 2023-05-30
Use Automerge's exposed `saveIncremental`

acurrieclark:
A straightforward swap. Figured it would be better to use the default exposed function to guard against any future breaking changes in the Automerge API.


pvh:
Ah, thanks -- when I wrote this it wasn't yet exposed.

PR #79, opened 2023-05-30
Ensure docs are ready before changing in `syncs a bunch of changes` test

acurrieclark:
This test would occasionally fail locally for me if the `docHandle was not ready. Adds an `await handle.valueI()` when needed.


pvh:
Is there a reason not to just `await doc.value()` without the test? It will resolve immediately if the doc is there and the function is already async.

acurrieclark:
Interestingly, I found that by always awaiting the value, the test took significantly longer to run. In a first draft I extended the timeout on the test but tried this and the test completed much quicker.

PR #83, opened 2023-06-07
Add release action

acurrieclark:
This PR:

- Upgrades Lerna to v6
- Adds a github release action

The action requires a `NPM_TOKEN` secret to be exposed as an action secret, and it works as follows:

1. Check if the root `package.json` version has been changed. If not, the action ends there.
2. Install dependencies
3. Build all packages
4. Run tests
5. Publish to npm
    - If the version includes alpha/beta, the release is tagged as `next`
6. Tag the release on the repo

The action is configured to run on **any push to main**.

I have tested this in a fork of the repo, with a dummy `NPM_TOKEN` and all seems to function as expected.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,80 @@
+name: Release
+on:
+  push:
+    branches:
+      - main
+
+permissions:
+  contents: write
+jobs:
+  check_if_version_upgraded:
+    name: Check if package version has been upgraded
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.version-updated.outputs.current-package-version }}
+      has_updated: ${{ steps.version-updated.outputs.has-updated }}
+    steps:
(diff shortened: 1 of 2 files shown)


pvh:
Sounds good... do we want to automatically cut preview releases? Maybe not, maybe just run smoke tests. I haven't automated releases on any packages before out of cowardice. You reckon this is best practice?

acurrieclark:
As it stands, a preview release won't be automatically cut on each push unless the version number in the root `package.json` is also bumped. The action will run on all merges, but won't continue if the version number hasn't changed. The tests still run in a separate action regardless.

This is certainly what I am using on my own packages and I set one up for Automerge last year.

It would technically allow anyone with write permission to `main` to create a release at any time, but we would have to manage that amongst ourselves. I would likely only ever use it to push `alpha` releases, which should be automatically tagged as `next` on npm.

PR #85, opened 2023-06-10
Remove build script from demo-counter-svelte

HerbCaudill:
This package never needs to be built. Removing the build script here means that Lerna won't include it when it builds all packages, which speeds up that task.


acurrieclark (approved):
Good idea, @HerbCaudill

PR #97, opened 2023-06-15
Remove SYNC_CHANNEL

HerbCaudill:
We weren't doing anything with this channel.


HerbCaudill:
ah never mind then. we can take care of this as part of the network subsystem changes

PR #98, opened 2023-06-15
Merge `DocCollection` into `Repo`

acurrieclark:
This is the start of a larger refactor which aims to reduce the complexity of the architecture a little.

This PR merges `DocCollection` and `Repo` into a single class, removing the need for some of the event exchanges which triggered the adding and removal of documents. A few notes:

1. I have left the actual events in, as they have always been emitted on the `Repo` object and may already be in use in userland.
2. There were only 2 tests in `DocCollection.test.ts`, and they already existed in `Repo.test.ts`, with the exception for a check that `handle.isReady()` is true immediately after creating a handle. That has been copied across.
3. I have made the various subsystems (network, storage and ephemeral) private then exposed them publicly with a getter, so there is no danger of them being replaced.

I have stopped short of removing the listeners on the `networkSubsystem` and passing them to the `CollectionSynchronizer` and `EphemeralData` classes, but I think that would be the next logical step.


acurrieclark:
This will take some effort to land given the current raft of changes, and I think energy will be better spent elsewhere.

I'll revisit this if and when a refactor is warranted.

pvh:
I'm going to reopen it if it's alright, so that I don't forget that we agreed to do this. My momentum is high on this bit of the code right now, so it might be relatively easy to do.

acurrieclark:
Understood

alexjg:
I rebased and merged this manually

Issue #92, opened 2023-06-13
Improve gossip protocol for ephemeral messages


HerbCaudill:
In `NetworkSubsystem`, if we see a broadcast message (currently only ephemeral messages), we rebroadcast it to all our other peers. 

https://github.com/automerge/automerge-repo/blob/4fce1a0a11c6482c078b0d4750c4123d9d9a72e3/packages/automerge-repo/src/network/NetworkSubsystem.ts#L48-L66

As noted in the comments, this is not a great gossip protocol. It relies on the network forming a tree.  If there are cycles, this approach will loop messages around forever.

PR #101, opened 2023-06-23
Network acceptance tests

HerbCaudill:
Consolidates tests for network adapters in a new package `automerge-repo-network-acceptance-tests`, which the adapters import and run. 

The packages exports a single function, `runAdapterTests`. This function runs a series of tests against a set of three peers, each represented by one or more instantiated network adapters.

The adapter passes a `setup` function, which must return an object with the following properties:

- `adapters`: A tuple representing three peers' network configuration. Each element can be either a single adapter or an array of adapters. Each will be used to instantiate a Repo for that peer.

- `teardown`: An optional function that will be called after the tests have run. This can be used to close connections, etc. 

For example, here's the test suite for `BroadcastChannel`: 

```ts
import { BroadcastChannelNetworkAdapter } from "../src"
import { type SetupFn, runAdapterTests } from "automerge-repo-network-acceptance-tests"

describe("BroadcastChannel", () => {
  const setup: SetupFn = async () => {
    const a = new BroadcastChannelNetworkAdapter()
    const b = new BroadcastChannelNetworkAdapter()
    const c = new BroadcastChannelNetworkAdapter()

    return { adapters: [a, b, c] }
  }

  runAdapterTests(setup)
})
``` 

This is the output when running that suite: 

```
↯ yarn workspace automerge-repo-network-broadcastchannel test             

  BroadcastChannel
    Adapter acceptance tests 
      ✔ can sync 2 repos (178ms)
      ✔ can sync 3 repos (579ms)
      - can broadcast a message


  2 passing (770ms)
  1 pending

Done in 11.30s.
```

(Note that the broadcast test is currently skipped because it goes into an endless messaging loop on non-tree-shaped networks; when #92 is fixed, that test should work).


packages/automerge-repo/test/Repo.test.ts
```diff
@@ -322,7 +322,7 @@ describe("Repo", () => {
     it("can broadcast a message", async () => {
       const { aliceRepo, bobRepo, teardown } = await setup()
 
-      const channelId = "m/broadcast" as ChannelId
```
HerbCaudill, line 325: UNRELATED - but this is added by EphemeralNetwork

packages/automerge-repo/src/Repo.ts
```diff
@@ -85,7 +85,7 @@ export class Repo extends DocCollection {
     // When we get a new peer, register it with the synchronizer
     networkSubsystem.on("peer", async ({ peerId }) => {
       this.#log("peer connected", { peerId })
-      await synchronizer.addPeer(peerId)
```
HerbCaudill, line 88: UNRELATED - unnecessary `await`

PR #108, opened 2023-06-27
Compaction intervals [draft]

pvh:
At the I&S lab summit we prototyped a little drawing toy which produces data at ~60hZ. This makes the current "every 20 changes" `saveTotal()` call pretty gratuitous (3Hz). This PR changes the default compaction interval to 1000 changes or 1m, whichever is longer. 

I started writing a test for it but the private methods stymied the approach I took and I got distracted from the work. That said, I think this (or something like this) is a better design than the current "every 20 changes" model and we should land something along these lines.

If anyone else has an idea how to test, I'm all ears.

### DIFF
--- a/packages/automerge-repo/src/storage/StorageSubsystem.ts
+++ b/packages/automerge-repo/src/storage/StorageSubsystem.ts
@@ -6,6 +6,7 @@ import { StorageAdapter } from "./StorageAdapter.js"
 export class StorageSubsystem {
   #storageAdapter: StorageAdapter
   #changeCount: Record<DocumentId, number> = {}
+  #compactedAt: number = 0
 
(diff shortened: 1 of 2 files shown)


acurrieclark:
The timer compaction will occur any time a page is loaded, which will be fine for isolated Repos, but i think will cause problems where multiple repos (eg tabs) are writing to the same storage adapter.

I know that probably shouldn't be encouraged, but we likely need a solution to that problem before we can introduce time based compaction.

pvh:
This is superseded by work in the storage patch @alexjg is wrapping up. (And will be easy to implement when that lands)

pvh:
(Leaving this PR open as a pseudo-tracking issue.)

PR #113, opened 2023-07-12
feat(changeDoc): add option parameter changeDoc

cateland:
Add the optional `ChangeOptions<T>` `options` parameter to `useDocument:changeDoc` function so it can pass it to the `DocHandle<T>:change` function.

This would be useful for example to pass along a message with the changes.


pvh (approved):
Thanks!

PR #114, opened 2023-07-13
chore(workspaces): fix workspace invocation and fix yarn version range

cateland:
Added information about `yarn` version to be used.
Fixed `yarn workspace` use by prefixing workspace name with `@automerge`.


pvh:
Thanks!

PR #115, opened 2023-07-18
Indexeddb & hashed incremental saves

pvh:
Okay, this patch grew out a bit! What we've got going on here is:
- a new automerge-storage array-key-based interface (for efficient sub-tree querying)
- new implementations of this interface for node's filesystem and direct indexedDB
- the node filesystem uses the first two characters of a sha-256 hash of the filename to split up the main directory tree
- a bunch of storage subsystem improvements to read from it
- a variety of small adjustments to places like the syncserver and tests and so on

Overall, I'm pretty happy with how it all shook out, and it seems to work in the tests and in the demo I tried.

Still, this one is a bit risky! It touches a lot of important code, and moves the data files around. I'd appreciate an extra set of eyes on it.

One other note: we're moving both the data files in the browser and on the sync server with this one. I haven't changed the implementations of either but as part of getting ready for 1.0, I wanted to make this one breaking change. 

At the moment, the best / simplest upgrade path is just to upgrade your sync-server first and have your clients connect and load their data all once. They'll feed the sync server everything it needs. Afterwards, upgrade the client and it will download the data the same way.

That said, if there's demand for a migration tool I can write one. Just let me know.


packages/automerge-repo-demo-todo/package.json
```diff
@@ -13,7 +13,7 @@
     "@automerge/automerge-repo-network-broadcastchannel": "^0.1.4",
     "@automerge/automerge-repo-network-websocket": "^0.1.4",
     "@automerge/automerge-repo-react-hooks": "^0.1.4",
-    "@automerge/automerge-repo-storage-localforage": "^0.1.4",
```
neftaly, line 16: any chance the localforage adapter can be archived somewhere? for learning purposes

pvh: sort of -- i don't know if we can implement the range operators but we could look into it

packages/automerge-repo-storage-indexeddb/src/index.ts
```diff
@@ -0,0 +1,134 @@
...
+export class IndexedDBStorageAdapter extends StorageAdapter {
+  private dbPromise: Promise<IDBDatabase>
+
+  constructor() {
+    super()
+    this.dbPromise = new Promise((resolve, reject) => {
+      const request = indexedDB.open("automerge", 1)
+
+      request.onerror = function () {
+        reject("Database error: " + request.error)
```
neftaly, line 12: Can you just reject(request.error)? 
It seems to be just a standard Error object https://developer.mozilla.org/en-US/docs/Web/API/IDBRequest/error

packages/automerge-repo-storage-indexeddb/src/index.ts
```diff
@@ -0,0 +1,134 @@
...
+    return new Promise((resolve, reject) => {
+      const transaction = db.transaction(["documents"], "readwrite")
+      const objectStore = transaction.objectStore("documents")
+      const request = objectStore.put(
+        { key: keyArray, binary: binary },
+        keyArray
+      )
+
+      request.onerror = function () {
+        reject("Unable to save data to database!")
```
neftaly, line 59: please include the error object (so we can tell why it errored)

packages/automerge-repo-storage-indexeddb/src/index.ts
```diff
@@ -0,0 +1,134 @@
...
+  async load(keyArray: string[]): Promise<Uint8Array | undefined> {
+    const db = await this.dbPromise
+
+    return new Promise((resolve, reject) => {
+      const transaction = db.transaction(["documents"])
+      const objectStore = transaction.objectStore("documents")
+      const request = objectStore.get(keyArray)
+
+      request.onerror = function () {
+        reject("Unable to retrieve data from database!")
```
neftaly, line 36: please include the error object (so we can tell why it errored)

packages/automerge-repo-storage-indexeddb/src/index.ts
```diff
@@ -0,0 +1,134 @@
...
+    const db = await this.dbPromise
+    const lowerBound = keyPrefix
+    const upperBound = [...keyPrefix, "\uffff"]
+    const range = IDBKeyRange.bound(lowerBound, upperBound)
+
+    return new Promise((resolve, reject) => {
+      const transaction = db.transaction(["documents"])
+      const objectStore = transaction.objectStore("documents")
+      const request = objectStore.openCursor(range)
+      const arrays: Uint8Array[] = []
```
neftaly, line 95: does this need to be `=new Uint8Array([])`? (not sure myself)

neftaly, line 95: just to confirm, this will be like this:
`new Array([ new Uint8Array(), ... ])`

packages/automerge-repo-storage-indexeddb/src/index.ts
```diff
@@ -0,0 +1,134 @@
...
+      const transaction = db.transaction(["documents"], "readwrite")
+      const objectStore = transaction.objectStore("documents")
+      const request = objectStore.delete(range)
+
+      request.onsuccess = function (event: Event) {
+        resolve()
+      }
+
+      request.onerror = function (event: Event) {
+        reject("Unable to remove data from database!")
```
neftaly, line 130: please include the error object (so we can tell why it errored)

packages/automerge-repo-storage-indexeddb/package.json
```diff
@@ -13,8 +13,7 @@
     "watch": "npm-watch"
   },
   "dependencies": {
-    "@automerge/automerge-repo": "^0.1.4",
-    "localforage": "^1.10.0"
+    "@automerge/automerge-repo": "^0.1.4"
   },
   "watch": {
     "build": {
```
neftaly, line 19: might be worth adding ".tsx" to the extensions list

packages/automerge-repo-storage-nodefs/src/index.ts
```diff
@@ -4,34 +4,41 @@ import { StorageAdapter } from "@automerge/automerge-repo"
...
       if (err) throw err
     })
   }
 
-  remove(docId: string): void {
-    fs.rm(this.fileName(docId), err => {
-      if (err) console.log("removed a file that does not exist: " + docId)
+  remove(key: string[]): void {
+    fs.rm(this.fileName(key), err => {
+      if (err) console.log("removed a file that does not exist: " + key)
```
neftaly, line 41: if you keep this, it should probably be `console.warn`

neftaly:
I still prefer if the `reject` calls return the actual `request.error`, and not a "cannot do x" error, to help with stack traces

packages/automerge-repo-storage-nodefs/src/index.ts
```diff
@@ -4,34 +4,41 @@ import { StorageAdapter } from "@automerge/automerge-repo"
 export class NodeFSStorageAdapter implements StorageAdapter {
   directory: string
 
-  constructor(directory = ".amrg") {
+  constructor(directory = "automerge-repo") {
```
neftaly, line 7: prefer if this was called prefix or something, as term is less overloaded, but I don't know what the convention is for storage adapters

packages/automerge-repo-storage-nodefs/src/index.ts
```diff
@@ -4,34 +4,62 @@ import { StorageAdapter } from "@automerge/automerge-repo"
...
         else resolve(data)
       })
     })
   }
 
-  save(docId: string, binary: Uint8Array): void {
-    fs.writeFile(this.fileName(docId), binary, err => {
+  save(key: string[], binary: Uint8Array): void {
+    fs.writeFile(this.fileName(key), binary, err => {
       // TODO: race condition if a load happens before the save is complete.
```
neftaly, line 54: hack to fix race cond:

this.isSavingPromise = Promise.resolve()

when saving, wrap isSavingPromise in a new promise that resolves once saving is complete.

When loading, await this.isSavingPromise first.

packages/automerge-repo-storage-indexeddb/src/index.ts
pvh: fair enough. did 'em all.

packages/automerge-repo-storage-nodefs/src/index.ts
pvh: i did not!

pvh:
> I still prefer if the `reject` calls return the actual `request.error`, and not a "cannot do x" error, to help with stack traces

ah yeah, that's a good point. it's just annoying when you get an error with no context / explanation of how it came to be. maybe it's fine.

packages/automerge-repo-storage-nodefs/src/index.ts
pvh: i rewrote this extensively to make the problem go away

neftaly:
> > I still prefer if the `reject` calls return the actual `request.error`, and not a "cannot do x" error, to help with stack traces
> 
> ah yeah, that's a good point. it's just annoying when you get an error with no context / explanation of how it came to be. maybe it's fine.

You can just put comments above `reject`, they will be accessible via stack trace

pvh:
Okay, I'm pretty happy with this and am inclined to ship almost as-is. There's a small amount of cleanup to do getting the demos onto the new code but otherwise I think we can ship it.

packages/automerge-repo/src/storage/StorageSubsystem.ts
```diff
@@ -2,10 +2,10 @@ import * as A from "@automerge/automerge"
 import { DocumentId } from "../types.js"
 import { StorageAdapter } from "./StorageAdapter.js"
 import { mergeArrays } from "../helpers/mergeArrays.js"
-import sha256 from "fast-sha256"
+import * as sha256 from "fast-sha256"
```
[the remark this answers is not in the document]
pvh: Yeah, no way to get synchronous values from crypto.subtle and I need them here to avoid race conditions without rearchitecting.

pvh:
@alexjg this is in your court, right? you had thoughts about how to handle the names of the incremental objects better.

PR #128, opened 2023-08-05
Automerge URLs

pvh:
Here we have it -- the first draft of Automerge URLs. Most of the patch is working towards making the guts of Automerge-Repo use binary byte arrays internally. It's not "worked through" in the sense that we still have a bit of a hodgepodge of string keys for objects and byte arrays being encoded and decoded all over. But it's progress!

Key changes: `find<T>(automergeUrl: AutomergeUrl)` and `handle.url`.  Are there other places we want to expose or accept URLs at the moment? .changeAt could take a URL-with-heads for example.

### DIFF
--- a/packages/automerge-repo/src/DocUrl.ts
+++ b/packages/automerge-repo/src/DocUrl.ts
@@ -1,43 +1,34 @@
 import { AutomergeUrl, DocumentId, StringDocumentId } from "./types"
 import { v4 as uuid } from "uuid"
-import Base58 from "bs58check"
+import bs58check from "bs58check"
 
 export const parseAutomergeUrl = (link: AutomergeUrl) => {
   const { stringDocumentId } = parts(link)
-  const documentId = Base58.decodeUnsafe(stringDocumentId) as
+  const documentId = bs58check.decodeUnsafe(stringDocumentId) as
     | DocumentId
     | undefined
   if (!documentId) throw new Error("Invalid document URL: " + link)
   return { documentId }
 }
 
-interface UrlFromStringIdOptions {
-  stringDocumentId: StringDocumentId
-  documentId?: never
-}
-interface UrlFromBinaryIdOptions {
-  stringDocumentId?: never
-  documentId: DocumentId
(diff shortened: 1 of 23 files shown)


packages/automerge-repo/src/DocUrl.ts
```diff
@@ -0,0 +1,60 @@
...
+  if (!documentId) throw new Error("Invalid document URL: " + link)
+  return { documentId }
+}
+
+interface UrlFromStringIdOptions {
+  stringDocumentId: StringDocumentId
+  documentId?: never
+}
+interface UrlFromBinaryIdOptions {
+  stringDocumentId?: never
```
pvh, line 19: check the types here instead of having different arguments

packages/automerge-repo/src/DocUrl.ts
```diff
@@ -0,0 +1,60 @@
...
+  const documentId = Base58.decodeUnsafe(stringDocumentId)
+  return documentId ? true : false
+}
+
+export const parts = (str: string) => {
+  const [m, stringDocumentId] = str.match(/^automerge:(\w+)$/) || []
+  return { stringDocumentId }
+}
+
+export const generate = (): DocumentId =>
```
pvh, line 49: this is poorly named and i should probably replace it with a different function that returns a URL

packages/automerge-repo/src/DocUrl.ts
```diff
@@ -0,0 +1,60 @@
...
+  Uint8Array.from(uuid(null, new Uint8Array(16))) as DocumentId
```
pvh, line 50: bahaha. fix this

packages/automerge-repo/src/DocUrl.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export const encode = (id: DocumentId): StringDocumentId => {
```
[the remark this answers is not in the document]
pvh: we probably shouldn't be exporting these

PR #135, opened 2023-08-08
websockets: Add protocol negotiations

### LINKED ISSUE #127: Network backwards compatibility post v1.0
One concern I have about repo is how to manage peers which are running out of sync versions of repo.

For example, an electron app is launched running v1.0. It syncs to a server running the same version.

How can we be sure that both adapters and repo itself can communicate correctly if only one of these is updated?

Talking with @pvh about this, it seems that _at a minimum_ network messages should include some kind of version tag, likely for both repo itself and the adapter employed.

alexjg:
This works by including a `supportedProtocolVersions` field in the join message from the client and then the server either returns a `selectedProtocolVersion` in it's `peer` message or returns a new `error` message.

Fixes #127

### DIFF
--- a/packages/automerge-repo-network-websocket/src/BrowserWebSocketClientAdapter.ts
+++ b/packages/automerge-repo-network-websocket/src/BrowserWebSocketClientAdapter.ts
@@ -8,6 +8,8 @@ import * as CBOR from "cbor-x"
 import WebSocket from "isomorphic-ws"
 
 import debug from "debug"
+import {ProtocolV1} from "./protocolVersion"
+import {InboundWebSocketMessage, OutboundWebSocketMessage} from "./message"
 const log = debug("WebsocketClient")
 
 abstract class WebSocketNetworkAdapter extends NetworkAdapter {
@@ -58,7 +60,9 @@ export class BrowserWebSocketClientAdapter extends WebSocketNetworkAdapter {
       throw new Error("WTF, get a socket")
     }
     if (this.socket.readyState === WebSocket.OPEN) {
-      this.socket.send(CBOR.encode({ type: "join", senderId: this.peerId }))
+      this.socket.send(
+        CBOR.encode(joinMessage(this.peerId))
+      )
     } else {
       this.socket.addEventListener(
         "open",
@@ -66,7 +70,9 @@ export class BrowserWebSocketClientAdapter extends WebSocketNetworkAdapter {
           if (!this.socket) {
             throw new Error("WTF, get a socket")
           }
-          this.socket.send(CBOR.encode({ type: "join", senderId: this.peerId }))
+          this.socket.send(
+            CBOR.encode(joinMessage(this.peerId))
+          )
         },
         { once: true }
       )
@@ -127,7 +133,7 @@ export class BrowserWebSocketClientAdapter extends WebSocketNetworkAdapter {
   }
 
   receiveMessage(message: Uint8Array) {
-    const decoded: InboundMessagePayload = CBOR.decode(new Uint8Array(message))
+    const decoded: OutboundWebSocketMessage = CBOR.decode(new Uint8Array(message))
 
     const {
       type,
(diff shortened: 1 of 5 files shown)


alexjg:
One open question here is what to do with the error message in the `BrowserWebSocketClientAdapter`. The intended semantics of the error type is that the connection should now be closed (the server forcibly closes it) but I'm not sure what that should mean for the whole adapter. I can see two options:

1. Add some kind of `onError` event to the adapter to communicate to the `NetworkSubsystem` that this adapter is dead and should be removed
2. Store the "errored" state on the adapter and just ignore all further attempts to communicate via this adapter.

pvh:
We probably don't want to nuke the adapter -- you might be running an upgrade and the server is temporarily on the wrong version. That shouldn't poison every client everywhere. But we do want that problem to be visible. Hm.

packages/automerge-repo-network-websocket/src/BrowserWebSocketClientAdapter.ts
```diff
@@ -8,6 +8,7 @@ import * as CBOR from "cbor-x"
 import WebSocket from "isomorphic-ws"
 
 import debug from "debug"
+import {ProtocolV1} from "./protocolVersion"
```
acurrieclark, line 11: Need to add the .js suffix here

packages/automerge-repo-network-websocket/src/NodeWSServerAdapter.ts
```diff
@@ -10,6 +10,8 @@ import {
   NetworkAdapter,
   PeerId,
 } from "@automerge/automerge-repo"
+import {ProtocolV1, ProtocolVersion} from "./protocolVersion"
+import {InboundWebSocketMessage} from "./message"
```
acurrieclark, line 14: Both of these need `.js` too

acurrieclark:
Just doing some testing with a websocket implementation here, and we do need to put some handling in place for when an error is returned. At the moment the client adapter tries to emit the error as a `message` back to the subsystem, which causes a number of errors.

It currently also reattempts joining repeatedly. Looking into it.

packages/automerge-repo-network-websocket/src/NodeWSServerAdapter.ts
```diff
@@ -116,9 +119,27 @@ export class NodeWSServerAdapter extends NetworkAdapter {
 
         // In this client-server connection, there's only ever one peer: us!
         // (and we pretend to be joined to every channel)
-        socket.send(
-          CBOR.encode({ type: "peer", senderId: this.peerId, channelId })
-        )
+        if (supportedProtocolVersions === undefined) {
+          socket.send(CBOR.encode({ type: "error", errorMessage: "missing supportedProtocolVersions"}))
+          this.sockets[senderId].close()
```
acurrieclark, line 124: The client adapter is set up to try to reconnect automatically whenever the socket is closed. If we close the socket from the server end, this results in an infinite loop of attempting to rejoin, rejection because of the version and the socket being closed again. 

Might it be better if the socket is closed on the client end once it realises it cannot connect?

packages/automerge-repo-network-websocket/src/BrowserWebSocketClientAdapter.ts
acurrieclark, line 1: The "error" message needs to be explicitly handled somewhere on the client end, otherwise the default switch option is to pass the message back to the network subsystem as a `message`.

acurrieclark:
@alexjg  and I also discussed the possibility of adding a `console.error` call here to at least give the user some feedback on the error, even if it's not thrown

alexjg:
I think I'm going to need to change this so that it doesn't _require_ a `supportedProtocolVersions` field in the initial hello message but infers it to be just the current version, otherwise we're going to be backwards incompatible with current clients and that's a problem because we're also changing the storage format and so we need to be able to first upgrade the sync server, then sync with it using old clients so that it has all the old data, then upgrade the clients.

alexjg:
I've updated this to default to protocol v1 and also to log an error. We'll need to do a bunch more work around retrying and eventually killing the adapter, but I think for release this PR is probably good enough.

PR #139, opened 2023-08-08
Remove `ChannelId` from the initial NetworkSubsystem join call

acurrieclark:
In its current form, there are no channels to join on any of the network adapters. We simply trigger the network setup using  a `join` command.

This PR keeps all the logic the same, but removes references to the concept of a `sync channel` ChannelId, as it is never referred to again.

Next step would be looking how to scope messages by `DocumentId` instead of `ChannelId` and to address the `ChannelId` used for ephemeral messaging.


pvh (approved):
Yeah, okay. Maybe we should rename join/leave though. This is fine to land as-is imo.

PR #142, opened 2023-08-09
Messaging system upgrade

acurrieclark:
This augments the messaging system so that it uses fixed message types for `sync` and `broadcast`.

It replaces `channelId`s on the sync message type with `documentId`s, but keeps them on broadcast messages.

More importantly, it fixes ephemeral messaging so that it won’t loop when the network is a mesh rather than a tree.

### DIFF
--- a/packages/automerge-repo-network-messagechannel/src/index.ts
+++ b/packages/automerge-repo-network-messagechannel/src/index.ts
@@ -1,4 +1,4 @@
-import { ChannelId, NetworkAdapter, PeerId } from "@automerge/automerge-repo"
+import { Message, NetworkAdapter, PeerId } from "@automerge/automerge-repo"
 import { MessagePortRef } from "./MessagePortRef.js"
 import { StrongMessagePortRef } from "./StrongMessagePortRef.js"
 import { WeakMessagePortRef } from "./WeakMessagePortRef.js"
@@ -27,66 +27,59 @@ export class MessageChannelNetworkAdapter extends NetworkAdapter {
     log("messageport connecting")
     this.peerId = peerId
     this.messagePortRef.start()
-    this.messagePortRef.addListener("message", e => {
-      log("message port received", e.data)
-      const { origin, destination, type, channelId, message, broadcast } =
-        e.data
-      if (destination && !(destination === this.peerId || broadcast)) {
-        throw new Error(
-          "MessagePortNetwork should never receive messages for a different peer."
-        )
+    this.messagePortRef.addListener(
+      "message",
+      (e: { data: AdapterMessage }) => {
+        log("message port received", e.data)
+
+        const message = e.data
+        if ("targetId" in message && message.targetId !== this.peerId) {
+          throw new Error(
+            "MessagePortNetwork should never receive messages for a different peer."
+          )
+        }
+
+        const { senderId, type } = message
+
+        switch (type) {
+          case "arrive":
+            this.messagePortRef.postMessage({
+              senderId: this.peerId,
+              targetId: senderId,
+              type: "welcome",
+            })
+            this.announceConnection(senderId)
+            break
+          case "welcome":
+            this.announceConnection(senderId)
+            break
+          default:
+            this.emit("message", {
+              ...message,
+              data: new Uint8Array(message.data),
+            })
+            break
+        }
       }
-      switch (type) {
-        case "arrive":
-          this.messagePortRef.postMessage({
-            origin: this.peerId,
-            destination: origin,
-            type: "welcome",
-          })
-          this.announceConnection(origin)
-          break
-        case "welcome":
-          this.announceConnection(origin)
-          break
-        case "message":
-          this.emit("message", {
-            senderId: origin,
-            targetId: destination,
-            channelId,
-            message: new Uint8Array(message),
-            broadcast,
-          })
-          break
-        default:
-          throw new Error("unhandled message from network")
-      }
-    })
+    )
 
     this.messagePortRef.addListener("close", () => {
       this.emit("close")
     })
   }
 
-  sendMessage(
-    peerId: PeerId,
-    channelId: ChannelId,
-    uint8message: Uint8Array,
-    broadcast: boolean
-  ) {
-    const message = uint8message.buffer.slice(
-      uint8message.byteOffset,
-      uint8message.byteOffset + uint8message.byteLength
+  send(message: Message) {
+    const data = message.data.buffer.slice(
+      message.data.byteOffset,
+      message.data.byteOffset + message.data.byteLength
     )
+
     this.messagePortRef.postMessage(
       {
-        origin: this.peerId,
-        destination: peerId,
-        channelId: channelId,
-        type: "message",
-        message,
-        broadcast,
+        ...message,
+        data,
       },
-      [message]
+      [data]
     )
   }
 
@@ -96,7 +89,7 @@ export class MessageChannelNetworkAdapter extends NetworkAdapter {
 
   join() {
     this.messagePortRef.postMessage({
-      origin: this.peerId,
+      senderId: this.peerId,
       type: "arrive",
     })
   }
@@ -118,3 +111,18 @@ interface MessageChannelNetworkAdapterConfig {
    */
   useWeakRef?: boolean
 }
+
+//types
+
+type ArriveMessage = {
+  senderId: PeerId
+  type: "arrive"
+}
+
+type WelcomeMessage = {
+  senderId: PeerId
+  targetId: PeerId
+  type: "welcome"
+}
+
+type AdapterMessage = ArriveMessage | WelcomeMessage | Message
--- a/packages/automerge-repo-network-websocket/src/BrowserWebSocketClientAdapter.ts
+++ b/packages/automerge-repo-network-websocket/src/BrowserWebSocketClientAdapter.ts
@@ -1,13 +1,9 @@
-import {
-  ChannelId,
-  InboundMessagePayload,
-  NetworkAdapter,
-  PeerId,
-} from "@automerge/automerge-repo"
+import { Message, NetworkAdapter, PeerId } from "@automerge/automerge-repo"
 import * as CBOR from "cbor-x"
 import WebSocket from "isomorphic-ws"
 
 import debug from "debug"
+import { FromClientMessage, FromServerMessage } from "./messages"
 const log = debug("WebsocketClient")
 
 abstract class WebSocketNetworkAdapter extends NetworkAdapter {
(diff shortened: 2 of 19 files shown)


acurrieclark:
Rebased on main to include @alexjg's work on Versioning

acurrieclark:
As part of this PR I added a sessionId to ephemeral messages. This was purely so a peer with a fixed `peerId` could be reset and it's messages still be passed on. The system works by tracking the last sent message count by sender `sessionId` and not sending on anything less than the last seen count.

This obviously adds some bloat to each broadcast message. The alternative would be to mandate that a peer with a fixed peerId shouldn't send their own ephemeral messages? The only example of this we have right now is the sync server, which shouldn't need to, I don't think?

packages/automerge-repo/test/Repo.test.ts
```diff
@@ -380,6 +376,40 @@ describe("Repo", () => {
       teardown()
     })
 
+    it.only("can broadcast a message without entering into an infinite loop", async () => {
```
pvh, line 379: you left the "only" on here -- notably though the broadcast channel worked fine before with two peers. it was adding a third one that broke things.

packages/automerge-repo/test/EphemeralData.test.ts
```diff
@@ -35,10 +35,14 @@ describe("EphemeralData", () => {
...
     })
-    ephemeral.receive(
-      otherPeerId,
-      ("m/" + destinationChannelId) as ChannelId, // TODO: this is nonsense
-      CBOR.encode(messageData)
-    )
+    ephemeral.receive({
+      targetId: targetPeerId,
+      senderId: otherPeerId,
+      channelId: destinationChannelId,
```
pvh, line 41: should we do away with channelIds? there's a 1:1 mapping of ephemeral messages to documents right now so no reason to use a different name here.

acurrieclark: I don't think there is any mapping of ephemeral messages to documents right now, is there? I thought you could give it whatever channelId you wanted, as it doesn't pass through the synchronizer?

packages/automerge-repo/test/Repo.test.ts
acurrieclark: Noted. I will add another peer to make sure.

packages/automerge-repo/src/network/NetworkAdapter.ts
```diff
@@ -1,33 +1,83 @@
...
+  abstract send(message: Message): void
 
   abstract join(): void
 
   abstract leave(): void
 }
 
+// utilities
+
+export function isValidMessage(message: Message): boolean {
```
pvh, line 19: we should probably move these bits into a separate file; we can potentially also use it as a type validator

acurrieclark: Good idea. Done.

packages/automerge-repo-network-messagechannel/src/index.ts
```diff
@@ -118,3 +111,18 @@ interface MessageChannelNetworkAdapterConfig {
    */
   useWeakRef?: boolean
 }
+
+//types
+
+type ArriveMessage = {
```
pvh, line 117: duplicated types here

packages/automerge-repo-network-websocket/src/BrowserWebSocketClientAdapter.ts
```diff
@@ -55,24 +56,36 @@ export class BrowserWebSocketClientAdapter extends WebSocketNetworkAdapter {
     )
   }
 
+  private transmit(message: FromClientMessage) {
+    if (!this.socket || this.socket.readyState !== WebSocket.OPEN) {
+      throw new Error("Websocket Socket not ready!")
+    }
+
+    const encoded = CBOR.encode(message)
```
pvh, line 64: alex has pointed out the CBOR encoding is a pain as a wireformat, particularly because CBOR-X is adding extra junk to the format

packages/automerge-repo-network-websocket/src/BrowserWebSocketClientAdapter.ts
```diff
@@ -55,24 +56,36 @@ export class BrowserWebSocketClientAdapter extends WebSocketNetworkAdapter {
...
       this.socket.addEventListener(
         "open",
         () => {
           if (!this.socket) {
             throw new Error("WTF, get a socket")
           }
-          this.socket.send(
-            CBOR.encode(joinMessage(this.peerId))
-          )
+          this.transmit(joinMessage(this.peerId!))
```
pvh, line 88: i know i wrote this, but it would be nice if there were a better idiom for this "do it now or when you can" kind of thing. obviously there's await but that flavors this function with async

packages/automerge-repo-network-websocket/src/BrowserWebSocketClientAdapter.ts
```diff
@@ -83,43 +96,19 @@ export class BrowserWebSocketClientAdapter extends WebSocketNetworkAdapter {
...
-    // underlying buffer even if we just have a uint8array view on it
-    const arrayBuf = encoded.buffer.slice(
-      encoded.byteOffset,
-      encoded.byteOffset + encoded.byteLength
-    )
-    if (!this.socket || this.socket.readyState !== WebSocket.OPEN) {
-      throw new Error("Websocket Socket not ready!")
-    }
-    this.socket.send(arrayBuf)
+    this.transmit(message)
```
pvh, line 111: is there a reason to not just make "send()" send and have the other functions use it?

pvh: (presumably we don't want to send zero length messages or send without a peerId inside transmit either)

packages/automerge-repo-network-websocket/src/BrowserWebSocketClientAdapter.ts
```diff
@@ -133,16 +122,9 @@ export class BrowserWebSocketClientAdapter extends WebSocketNetworkAdapter {
   }
 
   receiveMessage(message: Uint8Array) {
-    const decoded: OutboundWebSocketMessage = CBOR.decode(new Uint8Array(message))
+    const decoded: FromServerMessage = CBOR.decode(new Uint8Array(message))
```
pvh, line 125: not necessary now but we should probably put this through encode/decode functions which can also do validation

pvh: (and protocol migration, etc)

packages/automerge-repo-network-messagechannel/src/index.ts
acurrieclark: This is part of the reason that i felt these messages could be brought into Repo and out of the adapters. Technically `join` and `peer` on the websocket connections are doing exactly the same thing.

packages/automerge-repo-network-websocket/src/BrowserWebSocketClientAdapter.ts
acurrieclark: Do we have a preference here? I just used the existing encoding for now.

packages/automerge-repo-network-websocket/src/BrowserWebSocketClientAdapter.ts
acurrieclark: Will look at that today. Transmit also accepts internal message types, not just ones which match the Adapter interface signature.

This method also changed name from `sendMessage` to `send` while I was migrating the code. There is no reason for this other than it made it easier for me, so can be changed back to `sendMessage` straightforwardly should you prefer.

packages/automerge-repo-network-websocket/src/BrowserWebSocketClientAdapter.ts
acurrieclark: Makes sense. Once we settle on a protocol, I will separate out.

packages/automerge-repo-network-websocket/src/BrowserWebSocketClientAdapter.ts
acurrieclark: Obviously the versioning we have introduced here is only useful for changes outside of this encoding decision. We won't be able to change this without breaking.

packages/automerge-repo-network-websocket/src/BrowserWebSocketClientAdapter.ts
acurrieclark: This is now done, and tests passing.

would you like it named back to `sendMessage`?

alexjg (commented):
I have a few thoughts:

1. This appears to be conflating ephemeral messages with broadcast messages. E.g. `isEphemeralMessage`  returns `true` if the message has `broadcast: true`. Are these not separate concepts? We currently have no way of broadcasting sync messages but that could change. Maybe right now we default to broadcasting ephemeral messages, but that also feels like it could change. I feel like instead we should have an `ephemeral` and `broadcast` field for these two things.
2. Rather than having a session ID per message, can we instead have a session ID as part of the `join` message and the adapter can then just tag all messages it emits with the session ID? This saves having a session ID on every ephemeral message.

packages/automerge-repo/src/network/messages.ts
```diff
@@ -0,0 +1,56 @@
+// utilities
+
+import { SessionId } from "../EphemeralData"
+import { ChannelId, DocumentId, PeerId } from "../types"
+
+export function isValidMessage(message: Message): boolean {
```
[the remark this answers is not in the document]
acurrieclark: Excellent idea, yes

packages/automerge-repo-network-websocket/src/NodeWSServerAdapter.ts
```diff
@@ -49,68 +44,43 @@ export class NodeWSServerAdapter extends NetworkAdapter {
...
+    const { type, senderId } = cbor
 
     const myPeerId = this.peerId
     if (!myPeerId) {
       throw new Error("Missing my peer ID.")
     }
-    log(
-      `[${senderId}->${myPeerId}@${channelId}] ${type} | ${message.byteLength} bytes`
-    )
+    // log(
```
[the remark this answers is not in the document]
acurrieclark: Absolutely. This was me being too lazy to work out what channelId should be replaced with. Will add the documentId if it is present

acurrieclark:
> I have a few thoughts:
> 
> 1. This appears to be conflating ephemeral messages with broadcast messages. E.g. `isEphemeralMessage`  returns `true` if the message has `broadcast: true`. Are these not separate concepts? We currently have no way of broadcasting sync messages but that could change. Maybe right now we default to broadcasting ephemeral messages, but that also feels like it could change. I feel like instead we should have an `ephemeral` and `broadcast` field for these two things.
> 2. Rather than having a session ID per message, can we instead have a session ID as part of the `join` message and the adapter can then just tag all messages it emits with the session ID? This saves having a session ID on every ephemeral message.

1. I hadn't considered the difference when i wrote this, as the previous message type just had a `broadcast: boolean` property on it. Sounds like it would be wise to unify everything under `ephemeral` for now to avoid confusion later?
2. Can we guarantee that any peer who receives a message won't send it on more than once in this case? If 2 peers are directly connected, they would both `join` (or `peer`) at some point, and could initiate a session between themselves. However, if a peer is not connected directly to the origin peer, and the message could arrive at it by more than one route, then I don't know how it would be able to determine whether it was part of a session it has already seen?

packages/automerge-repo-network-messagechannel/src/index.ts
acurrieclark: @pvh not sure where might the best place for these types? There is actually a lot of duplicated code across the broadcastchannel and messagechannel adapters, which could easily be refactored if there were a sensible place to put the common code?

PR #143, opened 2023-08-09
Add an example sync server

alexjg:
We have moved the automerge-repo-sync-server to a separate repository. This is appropriate as that codebase will now be able to become more "industrial" with the addition of monitoring and docker containers and other such cloudy things. However, we would still like to be able to easily run examples locally without this stuff, and we would also like to be able to locally experiment with changes to the sync server. Therefore, add an example sync server in examples/sync-server which can be used with the other examples but which is explicitly not intended to be used anywhere else.


examples/sync-server/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "@automerge/example-sync--server",
```
acurrieclark, line 2: is the double `--` deliberate here? Cosmetic, I know

examples/sync-server/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "private": true,
+  "version": "0.0.1",
+  "main": "index.js",
+  "license": "MIT",
+  "type": "module",
+  "scripts": {
+    "start": "node ./index.js"
```
acurrieclark, line 9: Is it worth updating the script in the root `package.json` to point to this new package?

examples/sync-server/index.js
```diff
@@ -0,0 +1,81 @@
...
+      /** @ts-ignore @type {(import("automerge-repo").PeerId)}  */
+      peerId: `storage-server-${hostname}`,
+      // Since this is a server, we don't share generously — meaning we only sync documents they already
+      // know about and can ask for by ID.
+      sharePolicy: async () => false,
+    }
+    const serverRepo = new Repo(config)
+
+    app.get("/", (req, res) => {
+      res.send(`👍 @automerge/automerge-repo-sync-server is running`)
```
acurrieclark, line 49: Not strictly necessary, but might be worth renaming this to match the example package name too

examples/sync-server/package.json
alexjg: Good spot

examples/sync-server/package.json
alexjg: Yup good shout

PR #148, opened 2023-08-11
Unavailable message

acurrieclark:
#### TODO
- [x] Do we need to treat the local situation as we would if the doc has been deleted ie. remove the handle from the local cache?
- [x] Write test to ensure this works for a mesh network
- [x] Write test to ensure that a previously unavailable document can be requested again
- [x] Write test to check that if a previously requested and unavailable document later becomes available, it is properly synced with peers
- [x] Subsequent `find` calls to a still unavailable document should fire the unavailable event
- [x] `handle.doc()` should resolve to undefined when unavailable
---
This PR does 4 main things:

1. Fixes a bug (which caused issues with the rest of the PR) whereby receiving a sync message would create a DocSynchronizer for the document, which would add the document to the repo, which would trigger a `document` event which in turn caused an additional DocSynchronizer to be created for the same document. This is done via a flag on the document in the collection synchronizer. There is likely a better way to achieve this outcome.
2. Introduces a new `request` message type. This is identical in format to the `sync` message type, except the `type` property. It is sent from a peer only in the specific case that it does not have a document and is requesting to sync it from another peer if it does.
3. Adds a `doc-unavailable` message type which is sent to peers who have requested a document to explicitly inform them that the document is not available to that peer, nor any part of the network it is connected to.
4. The DocSynchronizer tracks the status of a document on its known peers. All peers are initially marked as `unknown`. If it receives a request for a doc, that peer is marked as `wants`. If it gets a sync message with heads, it marks the peer as `has`. If it receives an explicit `doc-unavailable` message, the peer is marked as `unavailable`. The synchronizer then uses these states to check if none of its peers have access to the document, then passes the `doc-unavailable` message back to any peers who have requested the doc.

**NB.** When a document is unavailable, `handle.doc()` will return `undefined`. This PR does not make any changes as to how the handle times out; this still throws an error for now.


pvh:
Sorry, Alex -- I tried to quickly merge/rebase this and it absolutely didn't work. I've not got the context to easily update the patch but if you can do that, I can try to land this (or @alexjg could) so we can ship a beta tomorrow.

acurrieclark:
Just merged in the changes from the Ephemeral branch. Fingers crossed that makes it easier to merge when Ephemeral is ready to land

PR #150, opened 2023-08-16
working on getting ephemeral messages onto the handle

### LINKED ISSUE #50: BroadcastChannel cannot be used with > 2 connections
When more than 2 clients are connected with broadcastchannel, messages bounce around forever and the clients freeze.

### LINKED ISSUE #92: Improve gossip protocol for ephemeral messages

### LINKED ISSUE #141: Ephemeral message scope
Ephemeral messages are currently emitted with a channelId which is not used to scope messages in any way.

My understanding of the current setup is that this means that 2 independent peers connected via (for example) a central server, would receive any ephemeral messages that the other sent, even if they were accessing a completely different set of documents.

If this is the case, would it be useful to be able to scope messages to a specific document, such that it is only sent to peers with whom the document is already being synced?

Are there any other scenarios?

pvh:
Closes #141 
Closes #92 
Closes #50


acurrieclark:
I think this is largely there. The only outstanding issue, which we could handle later, is that peers more than one hop away from the initial sender will reflect the message back to the last peer on the chain. That peer will drop it rather than emitting, but it's an extra message nonetheless.

The current idea for a solution is to identify the last peer in the NetworkSubsystem if the adapter only has a single peer connected. It won't work for BroadcastChannel, but it will help everywhere else for now.

PR #151, opened 2023-08-18
cbor: Don't use typed array tags

alexjg:
cbor-x by default tags Uint8Arrays with `64`, which is the tag from [RFC8746][1] for an array of uint8. This is a nice idea but actually really irritating when implementing an interoperable implementation on a platform which doesn't support this extension to CBOR. Disable this feature to make interoperability easier.

[1]: https://www.rfc-editor.org/rfc/rfc8746.html


pvh:
bit rough around the edges but it'll do; merging

Issue #147, opened 2023-08-11
New document sync and save


acurrieclark:
I have been looking in to how we might determine if a document is not available on the network. As a side effect, I was investigating how sync messaging is triggered.

At the moment, if we **create** a document on the repo, we set up a new empty document but don't persist anything to storage. We do, however, immediately start to sync it with other peers (if the sharePolicy allows).

When we **find** a document which doesn't exist, the same process happens, only the document stays in a requesting state.

Any connected peer will have no way of determining if the first (empty?) sync message is regarding an empty doc which has been created elsewhere and will shortly begin syncing changes or if it is a non existing document that the other peer is looking for.

I would suggest that the most logical solution would be to have a new document only sync over the network if it contains changes?

It might not matter, and we could be happy that a connected peer could respond to the initial sync in either case to say that it the document isn't available from it. If so, we just ignore the message when it arrives at the creating peer.

pvh:
Yeah, this has always been a little underbaked conceptually, though really I'm surprised how far its gotten. Perhaps we should consider making the offer() and accept() a bit more explicit in the system? That would imply reifying peers a bit which honestly is probably the right idea.

acurrieclark:
@alexjg pointed out earlier that the counterpoint to this is that when a document is created locally, there now exists a document URL. If the document isn't saved somewhere (be it over the network or locally) then it is dead. 

In this case the argument would be that all created docs should be saved locally whether they contain data or not

acurrieclark:
I had a little time this afternoon to look at this some more, and have started to see what happens if we call the first sync message triggered by `find` a `request` instead of a `sync`. Essentially it has the same body, just a different type.

pvh:
i support this concept. i agree that saving empty documents is a healthier design than just dropping them.

pvh:
Do we want to block 1.0 on this?

alexjg:
Ideally we would design our network messages such that we can backwards compatibly implement this. I.e. add a `request` message and handlers for it which don't do anything different to the current logic and we can implement the "I don't have this" logic later. That way I think we don't need to block 1.0?

acurrieclark:
Closed by #154 and #142

PR #154, opened 2023-08-25
Save documents to storage as soon as we create them

alexjg:
Problem: currently it is possible to create a DocHandle, obtain the URl to that dochandle, and then never save anything to the dochandle (e.g. by closing the browser window) and therefore never save anything to storage. This means that you have a URL which points at a document which can _never_ be resolved. This is not a theoretical problem. In several example applications we have this pattern of creating a `DocHandle`, then immediately adding the URL to the window location. Because many of these applications don't call `DocHandle.change` until the user has pressed a button (e.g. to incrementa counter) it is possible to refresh the browser window before any call to `DocHandle.change` and end up in a situation where the `Repo` is trying to find a document which never existed.

Solution: Save the document as soon as the handle is created. This might waste a few bytes of storage but it means we never hand out URLs to documents which don't exist.


acurrieclark:
Looks good to me.

Closes #147

PR #158, opened 2023-08-31
Merge join into connect on network adapters

acurrieclark:
#156 resulted in a refactor which highlighted that when we set up each network adapter, we essentially call `adapter.connect()` followed immediately by `adapter.join()`. The join is superfluous.

This also renames `leave` to disconnect and removes the unused `join` on the network subsystem.

We could also perhaps remove `leave` from the subsystem too? It isn't used anywhere yet.


acurrieclark:
Leave also removed

PR #163, opened 2023-09-11
Allow broadcastchannel channel name to be configured

acurrieclark:
I am using the BroadcastChannel adapter to locally sync 2 different repos managing different docs: one which only syncs locally and the other which also has a connection to the outside world. To keep the docs separate, we need separate channels.

This PR allows a `channelName` option to be passed in to the adapter construction as part of an optional `options` object.

### DIFF
--- a/packages/automerge-repo-network-broadcastchannel/test/index.test.ts
+++ b/packages/automerge-repo-network-broadcastchannel/test/index.test.ts
@@ -3,6 +3,7 @@ import {
   type SetupFn,
   runAdapterTests,
 } from "../../automerge-repo/src/helpers/tests/network-adapter-tests.js"
+import { PeerId } from "@automerge/automerge-repo"
 
 describe("BroadcastChannel", () => {
   const setup: SetupFn = async () => {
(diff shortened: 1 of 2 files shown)


packages/automerge-repo-network-broadcastchannel/test/index.test.ts
```diff
@@ -14,4 +15,29 @@ describe("BroadcastChannel", () => {
   }
 
   runAdapterTests(setup)
+
+  it("should allow a channel name to be specified in the options", async () => {
```
pvh, line 19: this is testing two things here -- first that it can set a channel name and second that it doesn't connect. can you either update the test description or split the tests into two? other than that, this is great, ship it

pvh (changes requested):
good to go with one tiny change

packages/automerge-repo-network-broadcastchannel/test/index.test.ts
acurrieclark: Have updated the description here as I am unsure how to test specifically that the name has been set from outside. If we add further options, this could be split out further relatively simply, I would think

PR #179, opened 2023-10-03
Refactor message types

HerbCaudill:
I was having trouble following the logic of these types, with `FooMessage` vs `FooMessageEnvelope` vs `FooMessageContents`, plus the various union types. I've refactored this in a way that I find easier to understand & work with: 

- All message types extend a base type that gives you `senderId` and `targetId`
- There are no `XxxMessageContents` or `XxxMessageEnvelope` types
- `RepoMessage` (formerly `Message`) is the union of message types that a network adapter hands off to a Repo. 
- `Message` (formerly `NetworkAdapterMessage` is the union of *all* the message types a network adapter might deal with, including the ones it uses internally. 
- There is a `MessageContents` type that is just `Message` without `senderId`: This is what is provided to the NetworkSubsystem.
- You can also pass a specific message type to `MessageContents`, in which case it may omit other things that are added by NetworkSubsystem. This is only useful ephemeral messages: `MessageContents<EphemeralMessage>` also omits `sessionId` and `count`.


packages/automerge-repo/src/network/NetworkSubsystem.ts
```diff
@@ -110,25 +107,33 @@ export class NetworkSubsystem extends EventEmitter<NetworkSubsystemEvents> {
...
+        return {
+          ...message,
+          senderId: this.peerId,
+        }
+      }
     }
+
+    const outbound = prepareMessage(message)
+    this.#log("sending message", outbound)
+    peer.send(outbound as Message)
```
HerbCaudill: it felt cleaner to me to wrap the business of adding info to the message in a little locally scoped function, so we don't have to duplicate the logging and sending.

packages/automerge-repo/src/network/messages.ts
```diff
@@ -1,220 +1,147 @@
...
-// otherwise typedoc can't produce nice docs. We use this `static_assert` thing
-// to make sure the types continue to line up though.
-type _check_sync_message = static_assert<SyncMessage extends SyncMessageEnvelope & SyncMessageContents ? true : false>
 /**
  * A sync message for a particular document
  */
-export type SyncMessage = {
-  /** The peer ID of the sender of this message */
-  senderId: PeerId
+export type SyncMessage = MessageBase & {
```
alexjg, line 14: The problem with this construction is that it generates unhelpful docs. In this instance the docs look like this:

baddocs

There are two things which bother me about this:

1. The `MessageBase` type is not exported so there's no way for someone reading the docs to see what it is. 
2. Even if it was exported the `MessageBase & {...}` feels like the wrong way to present this information. The fact that there's a `MessageBase` type common to all message types feels like it obscures the things a reader would be interested in here, which is the structure of this particular type.

Unfortunately typedoc doesn't have access to the resolved types so it can't render the actual structural type. That's why I was doing the acrobatics with the `static_assert` stuff beforehand.

HerbCaudill: OK in that case I think we'd just want to inline those two properties in all the message types. I'll do that tomorrow.

HerbCaudill: I've been ruminating on this and I wonder if this is actually the right call. 

Full disclosure, I'm biased because as a developer I've personally never found autogenerated docs to be a big help to me as a developer. They always seem to be simultaneously too much and not enough: They expose you to all the complexity all at once without providing context. You're always going to need some hand-written docs to provide that context, and for completeness there's always the source code. 

But even granting that some developers will find it useful, it seems like given a trade-off between the clarity of the typedoc output and the clarity & maintainability of the actual code, we should always prioritize the latter.

alexjg: I guess my bias is in the opposite direction but that might be partly because I've spent a lot of time in Rust-land where the documentation tooling is pretty fantastic and so the standard thing to do is to write all your documentation as doc comments on the source code. Anecdotally though several people have thanked me for the API docs on automerge and automerge-repo so I think people are using it (I've also made an effort to include context in the autogenerated docs where possible, e.g. https://automerge.org/automerge/api-docs/js/ contains a lot of prose about how the various parts of the library fit together).

Really though I don't think the issue is with the general idea of autogenerated docs, it's typedoc specifically is a little limited. I think with a structural type system like typescript you ideally want the option to tell the doc generator to document the actual structural type, not the way the type is constructed. I wonder if there's a way to do that manually though. What I was attempting to do previously with the `static_assert` thing was something like this. I think what you would do is define the types in the manner you suggest for use internally, but then create special exported types would have all the properties inlined and you use a `static_assert` to make sure they don't drift. Because typescript is structural it doesn't actually matter which type you use, as long as they match up.

HerbCaudill: > Anecdotally though several people have thanked me for the API docs on automerge and automerge-repo so I think people are using it

OK, that's good information. 

I've just inlined `senderId` and `targetId`.

I do think the point stands that going forward we should be careful about sacrificing code legibility for the sake of the auto-generated docs. In this case the base class isn't really buying us that much. But I don't think it would be wise to completely swear off a fundamental TypeScript feature like type inheritance just because the docs can't handle it gracefully. Clarity of source code is important and the `static_assert` stuff gave me a migraine. 😅

alexjg: Yeah I agree it's a trade-off, in this case I think it's worth it but always something we have to think about.

PR #180, opened 2023-10-03
Not all non-ephemeral messages are sync messages

HerbCaudill:
now that we have the `request` and `unavailable` message types, we can't refer to all non-ephemeral messages as sync messages.


pvh:
good eye

PR #181, opened 2023-10-03
Repo.delete: small refactoring for legibility

HerbCaudill:
I don't know why but lines that begin with `;` upset me. 😅


pvh:
prettier will probably autocorrect these in. it's time to embrace not caring what the formatting is as long as you don't have to think about it.

Issue #149, opened 2023-08-14
Delete protocol


acurrieclark:
When we delete a document from a local repo, all it does set the local handle to a deleted state, emit an event and remove from local storage.

While there is a test to ensure that a deleted document can be refetched, I don't think this works as expected. For example, while we remove the local handle from the cache, we don't delete the local synchronizer, nor do we inform the network that we have have deleted the document.

I would suggest that we should remove the local synchronizer and probably inform the network of the deletion. Relevant peers should then remove the peer from their list of peers to synchronize with. This is if particular relevance to a sync server, which would otherwise look to continue syncing changes it receives from elsewhere.

At some point we should add a hook which a peer can use to determine if there is anything else which should be done with a delete message. For example, if you deleted in one tab, you might want to delete in a service worker and all other tabs but just stop syncing from your sync server.

acurrieclark:
Also worth writing a test to check that if a document (which has not previously been synched with another peer) has been deleted, that it cannot be requested by another peer.

PR #183, opened 2023-10-05
Use Vitest instead of Mocha

HerbCaudill:
This is an experiment replacing mocha throughout with [Vitest](https://vitest.dev/), which was already what the react-hooks package was using. It's muuuuch faster, especially in watch mode (which is on by default). Vitest 1.0 just entered beta, but this is already a vastly superior experience & I'd suggest making the switch. 

I haven't looked into the CI implications, but I can.


pvh:
I'm no fan of Mocha and fast tests are good tests!

packages/automerge-repo/test/Repo.test.ts
```diff
@@ -257,19 +260,23 @@ describe("Repo", () => {
...
-      )
+      // const bobHandle = repo.find<TestDoc>(handle.url)
+
+      // await assert.rejects(
+      //   rejectOnTimeout(bobHandle.doc(), 10),
+      //   "document should have been deleted"
+      // )
 
-      assert(!bobHandle.isReady())
+      // assert(!bobHandle.isReady())
```
acurrieclark, line 279: Are these deliberately commented out?

HerbCaudill: Good catch - I meant to ask about this. 

In both of these cases, the `bobHandle.doc()` call sets up several XState `waitFor` promises, which are rejected after the test succeeds. Mocha lets that slide, but Vitest doesn't. 

We could sort that out, but I'm not really sure that a timeout is the desired behavior here: If I ask for a deleted document, I should be told that it's been deleted -- I shouldn't turn around and ask the network for it or whatever.  

The variable name `bobHandle` is misleading, because this is all happening on a single repo. But we have to deal with this question whether we deleted the document or someone else did. Presumably this means 

I've read #48 and #54 and it seems like we left a lot of questions about deletion unresolved. I'll open an issue as a placeholder for that conversation. In the meantime, I'd suggest just omitting this part of these two tests.

HerbCaudill: > ~~ I'll open an issue ~~

Never mind, just saw #149

acurrieclark: I see. Handle.doc() should probably return undefined too, as it does when unavailable?

HerbCaudill: Yep

pvh:
Oh man, this is such a huge improvement.  I merged it after testing locally and seeing how much faster it is. Thanks so much for this!

PR #184, opened 2023-10-09
DocHandle.doc(): don't yield tick

HerbCaudill:
I've never really been clear why we needed this `await pause()`, but at some point I removed it and broke blutack and so we put it back. 

It shouldn't ever really hurt anything to have an async function yield one tick, but you also shouldn't really need it. Currently tests pass with and without it. 

If this still breaks blutack, let's get to the bottom of why; and if we can't find a better solution, try to document it and get it under test.

### DIFF
--- a/packages/automerge-repo/src/DocHandle.ts
+++ b/packages/automerge-repo/src/DocHandle.ts
@@ -290,7 +290,6 @@ export class DocHandle<T> //
   async doc(
     awaitStates: HandleState[] = [READY, UNAVAILABLE]
   ): Promise<A.Doc<T> | undefined> {
-    await pause() // yield one tick because reasons
     try {
       // wait for the document to enter one of the desired states
       await this.#statePromise(awaitStates)


pvh:
I think if blutack breaks from this it's blutack's fault and we should fix it there :)

packages/automerge-repo/src/DocHandle.ts
```diff
@@ -290,7 +290,6 @@ export class DocHandle<T> //
   async doc(
     awaitStates: HandleState[] = [READY, UNAVAILABLE]
   ): Promise<A.Doc<T> | undefined> {
-    await pause() // yield one tick because reasons
```
pvh, line 293: Is pause used elsewhere?

pvh: Oh, I see it's in rejectOnTimeout. All good.

acurrieclark:
For what it's worth, I have removed the pause in my local version and nothing has broken here yet!

pvh:
Well, it seems fine and I've merged it. If blutack is sad, we'll figure out why.

PR #185, opened 2023-10-10
Fix a handful of lint errors and remove some `any`s

acurrieclark:
This is mostly cosmetic, but means we should all have fewer errors cluttering our workspaces. No functional changes here.

### DIFF
--- a/packages/automerge-repo/src/DocHandle.ts
+++ b/packages/automerge-repo/src/DocHandle.ts
@@ -431,7 +431,7 @@ export class DocHandle<T> //
    * a user could have multiple tabs open and would appear as multiple PeerIds.
    * every message source must have a unique PeerId.
    */
-  broadcast(message: any) {
+  broadcast(message: unknown) {
     this.emit("ephemeral-message-outbound", {
       handle: this,
       data: encode(message),
@@ -474,14 +474,14 @@ export interface DocHandleChangePayload<T> {
   patchInfo: A.PatchInfo<T>
 }
 
-export interface DocHandleEphemeralMessagePayload {
-  handle: DocHandle<any>
+export interface DocHandleEphemeralMessagePayload<T> {
+  handle: DocHandle<T>
   senderId: PeerId
   message: unknown
 }
 
-export interface DocHandleOutboundEphemeralMessagePayload {
-  handle: DocHandle<any>
+export interface DocHandleOutboundEphemeralMessagePayload<T> {
+  handle: DocHandle<T>
   data: Uint8Array
 }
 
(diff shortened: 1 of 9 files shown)


packages/automerge-repo/src/DocHandle.ts
```diff
@@ -490,9 +490,9 @@ export interface DocHandleEvents<T> {
   change: (payload: DocHandleChangePayload<T>) => void
   delete: (payload: DocHandleDeletePayload<T>) => void
   unavailable: (payload: DocHandleDeletePayload<T>) => void
-  "ephemeral-message": (payload: DocHandleEphemeralMessagePayload) => void
+  "ephemeral-message": (payload: DocHandleEphemeralMessagePayload<T>) => void
```
pvh, line 493: Ah, I think this is wrong. The payload is not <T> (the doc type) but a different message type.

pvh: Oh, no, I see, nevermind.

packages/automerge-repo/src/storage/StorageSubsystem.ts
```diff
@@ -122,18 +122,18 @@ export class StorageSubsystem {
     if (!this.#shouldSave(documentId, doc)) {
       return
     }
-    let sourceChunks = this.#chunkInfos.get(documentId) ?? []
+    const sourceChunks = this.#chunkInfos.get(documentId) ?? []
     if (this.#shouldCompact(sourceChunks)) {
-      this.#saveTotal(documentId, doc, sourceChunks)
+      void this.#saveTotal(documentId, doc, sourceChunks)
```
pvh, line 127: what's with the voids?

acurrieclark: These are all hanging promises, and my understanding is that this is the best way to indicate that you are not doing anything with the result

pvh: I think if we're going to do this we should enable a linter rule that enforces it or else we should not do it at all. I have a mild preference for the latter (the voids are just another thing to read and understand IMHO) but this is a very weak opinion and I'd be fine with having them everywhere. I'd like to avoid it becoming a thing we have to *remember* to do and so only some files have them and others not.

pvh: I'm going to merge this since the rest of it is objectively better and this is fine but I would appreciate it if you'd pick either one of the two options and run with it :)

acurrieclark: Interestingly, the reason I spotted these was because my editor already flags these. Will look into one the why and make it explicit either way

acurrieclark: Seems like it's already in the eslint rules. Does it not show an error on your end?

https://github.com/automerge/automerge-repo/blob/3995f1fbe8f5fed3aa4236bfd8bad61c416675dc/packages/automerge-repo/.eslintrc#L23C37-L23C37

PR #187, opened 2023-10-10
remove the second join

pvh:
As I'm tracking down an unrelated bug on the network side I noticed we were "join()"ing twice. I think we just need the top one. Alex, you're probably most up-to-speed on this but the tests pass and it seems to me like this should work?


pvh:
#190 removes the *other* join, but we need to keep one so let's drop this one.

PR #189, opened 2023-10-11
Remove diff

### LINKED ISSUE #167: Lose extra diff
We are currently generating change patches by calling `Automerge.diff` after each change. For an average change, this `diff` takes ~35ms.

While this isn't a spectacularly long period of time, the rest of the change can sometimes only take a single digit ms amount of time. If we could somehow get the patches from the `patchCallback`, this would likely speed up the whole transaction.

We could also get the change source at the same time.

acurrieclark:
This PR:

- Uses the already calculated Patches from the patch callback when including patches in the `change` event callback
- Adds the correct patch source
- Adds an additional `sourcePeerId` property when a change has originated from a sync message

#### TODO
- [x] Add tests for patch source

Closes #167

### DIFF
--- a/packages/automerge-repo/src/DocHandle.ts
+++ b/packages/automerge-repo/src/DocHandle.ts
@@ -15,11 +15,11 @@ import {
 } from "xstate"
 import { waitFor } from "xstate/lib/waitFor.js"
 import { headsAreSame } from "./helpers/headsAreSame.js"
-import { pause } from "./helpers/pause.js"
 import { TimeoutError, withTimeout } from "./helpers/withTimeout.js"
 import type { DocumentId, PeerId, AutomergeUrl } from "./types.js"
 import { stringifyAutomergeUrl } from "./DocUrl.js"
 import { encode } from "./helpers/cbor.js"
+import { patchesFromChange } from "./helpers/patchesFromChange.js"
 
 /** DocHandle is a wrapper around a single Automerge document that lets us
  * listen for changes and notify the network and storage of new changes.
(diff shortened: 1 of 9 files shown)


pvh:
Oh man, this would totally work but I really think this is better fixed one level up the stack in Automerge-JS. This callback closure/capture stuff is pretty painful.

Related though, I think it's would be very unfortunate to change the signature for the public `update()` function like this because it makes it much harder for users to modify an automerge document using the underlying Automerge primitive functions. (Note how you had to modify all the calls: change, merge, etc.)

There's gotta be a way to do this better. I'm going to brainstorm with @orionz and see what we can come up with.

pvh:
Hmmm, if we look here https://github.com/automerge/automerge/blob/9400e195f1d01b0c10d535c1c1ccd321e684c362/rust/automerge/src/autocommit.rs#L171 we can see that the .diff should be fast if the diff cursor and heads match.

acurrieclark:
From talking to @alexjg about this, it seems that there isn't a way for `diff` to act that way at the javascript level, as by that point the patches have been discarded.

As we discussed on Monday, in an ideal world Automerge itself would return patches from function calls like `change`. If new functions were to be introduced, and their use was still faster than an additional call to diff, a separate update signature would still be required, as above. I am certainly not a fan of the closure/capture dance either, but it could be replaced very straightforwardly if new automerge functions were introduced.

I would also like to include the peerId of where the change originated too.

I will persevere with this for now, but i'll add a new update method instead of breaking the existing one.

acurrieclark:
Made a few changes, and ended up keeping the one `update` method but which now accepts 2 different callbacks; one which returns the new doc only and the other which returns the patchCallback too. This should now be non breaking.

pvh:
Okay! Diff is now roughly free. I'm going to mark this PR as closed but we should find a way to get you your sources based on this.

pvh:
Also -- please confirm everything is fast and still working as expected for you in 1.0.8. I moved a few other things out of being render blockers which should help a lot but I always feel nervous about mucking around with asynchrony.

PR #190, opened 2023-10-12
Some fixes

alexjg:
This PR fixes a number of issues which are leading to apparent sync failures when the sync server is running slowly:

* Most importantly, fix a bug where an unavailable document never transition to requesting if a peer is announced on the network adapter after the document has been marked as unavilable
* Clean up event handlers on the websocket client adapter. This was causing occasional exceptions when messages arrived on the old event handlers and attempted to send messages on a new (reconnected) socket which was not ready
* Avoid decoding sync messages before sending them - this can be quite expensive

### DIFF
--- a/packages/automerge-repo-network-websocket/src/BrowserWebSocketClientAdapter.ts
+++ b/packages/automerge-repo-network-websocket/src/BrowserWebSocketClientAdapter.ts
@@ -31,6 +31,7 @@ export class BrowserWebSocketClientAdapter extends WebSocketNetworkAdapter {
   }
 
   connect(peerId: PeerId) {
+    this.socket?.removeAllListeners()
     if (!this.timerId) {
       this.timerId = setInterval(() => this.connect(peerId), 5000)
     }
@@ -43,7 +44,6 @@ export class BrowserWebSocketClientAdapter extends WebSocketNetworkAdapter {
       log(`@ ${this.url}: open`)
       clearInterval(this.timerId)
       this.timerId = undefined
-      this.join()
     })
 
     // When a socket closes, or disconnects, remove it from the array.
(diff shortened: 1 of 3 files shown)


packages/automerge-repo-network-websocket/src/BrowserWebSocketClientAdapter.ts
```diff
@@ -43,7 +44,6 @@ export class BrowserWebSocketClientAdapter extends WebSocketNetworkAdapter {
       log(`@ ${this.url}: open`)
       clearInterval(this.timerId)
       this.timerId = undefined
-      this.join()
```
acurrieclark, line 46: Interestingly, @pvh opened #187 which removed the other join. Is that the better option? I don't suppose there is an advantage either way.

packages/automerge-repo/test/Repo.test.ts
```diff
@@ -663,6 +663,56 @@ describe("Repo", () => {
...
+      // connect it to the other end of the MessageChannel
+      const b = new Repo({
+        storage,
+        peerId: "b" as PeerId,
+        network: [new MessageChannelNetworkAdapter(ba)],
+      })
+
+      // The empty repo should be notified of the new peer, send it a request
+      // and eventually resolve the handle to "READY"
+      await handle.whenReady()
```
acurrieclark, line 712: Not necessary now, but do you think we should also add a `ready` event for this kind of case?

packages/automerge-repo-network-websocket/src/BrowserWebSocketClientAdapter.ts
alexjg: `join` is already idempotent so it doesn't really matter. I think this is a little clearer because there's one less calback involved.

PR #195, opened 2023-10-14
Debounce sync and save -- only run them every 100ms

pvh:
This has two effects: first, when there's a lot happening in the system it reduces the "noise" -- your frontend will still be responsive but saving to the local storage and sending sync messages to your peers will only run at 10hz.

Second, this also moves saving and generating sync messages out of the critical render frame loop on the browser. It's still possible to be delayed by trying to type during a slow sync message generation or saveIncremental but in most cases this will radically improve responsiveness.


pvh:
Requested reviews from @acurrieclark and @HerbCaudill mostly as a heads-up. It seems to work great in my apps but any time I make things async that were previously sync I get a little nervous so I just wanted to keep you guys from being surprised.

alexjg:
Is denounce the right primitive here? To my reading this code would have sync and save wait until there's been no activity for 100ms right? Whereas what we want is for it to happen at most once per 100ms. Or a I misreading the code?

pvh:
Yeah, you're right -- I just realized this talking to Orion. I think we want a ... quantize? Rate limit?

pvh:
Throttle. Updating the patch.

pvh:
Hmm, the normal behaviour of a throttle function is that it drops extra calls. If we have a *second* throttled call, we want to not drop it but schedule it for the end of the timeout. It's kind of a hybrid debounce/throttle. Let me see...

pvh:
Okay, updated to use a debounced throttle design with diagram and everything. Tests pass but scrutiny is very welcome.

alexjg:
This looks sound to me. Can we change every mention of "debounce" to throttle though?

pvh:
Will do.

PR #196, opened 2023-10-14
Import the useDocuments hook from blutack (untested)

pvh:
Just something I've had lying around for a while and meant to import here.


pvh:
Okay, so in React you must always use the same number of hooks in every invocation of a component. This makes it difficult to perform certain tasks, like loading a variable number of subdocuments! The `useDocuments` hook, originally developed by @paulsonnentag, gives you the ability to maintain a map of subdocuments within a component. It doesn't support changing those documents (though it could in principle) because the API was even gnarlier.

Honestly, this whole thing makes me kind of sad but it's a useful tool in the current world.

packages/automerge-repo-react-hooks/src/useHandles.ts
```diff
@@ -0,0 +1,67 @@
...
+  [url: AutomergeUrl]: T
+}
+
+/** useDocuments(docUrls)
+ * maintains a map of document states from { [docUrl]: Doc<T> }
+ * useful for lists and other places you might have a set of documents
+ * does not include change capability at the moment
+ * if you want that, consider adapting this to useHandles and send a PR
+ */
+export function useDocuments<T>(docUrls?: AutomergeUrl[]): DocUrlMap<T> {
```
sroze, line 15: Note: let's rename the file name to `useDocuments` (currently it is `useHandles`).

HerbCaudill:
reopening as new PR

PR #197, opened 2023-10-14
Set moduleResolution to Bundler in examples

HerbCaudill:
I meant to include this commit in #193. This resolves the issue with types for `useDocument` etc.


pvh:
Probably won't have time to look at this over the weekend but I'll leave a link to the docs here for myself (or whoever wants to look at this before me): https://www.typescriptlang.org/docs/handbook/module-resolution.html

I guess these projects are built with esbuild so bundler makes sense now I think more about it. Okay. I'll merge it :)

pvh:
(I see it's a draft -- feel free to merge when you've finished.)

PR #198, opened 2023-10-14
Remove build command from examples/react-use-awareness

HerbCaudill:
Running `build` from the monorepo root builds all packages that have a build command. There's never any reason to build the examples, and it slows things down.


pvh:
Ah thanks -- we should rename this project to be consistent with the others too. Also it could probably use a little explanatory text in the code / rendered UI. This is really just a note to myself (or maybe @neftaly.)

PR #200, opened 2023-10-14
Correctly clean up websocket event listeners

alexjg:


packages/automerge-repo-network-websocket/src/BrowserWebSocketClientAdapter.ts
```diff
@@ -31,33 +31,44 @@ export class BrowserWebSocketClientAdapter extends WebSocketNetworkAdapter {
...
-    this.socket.addEventListener("message", (event: WebSocket.MessageEvent) =>
+    const onMessage = (event: WebSocket.MessageEvent) => {
       this.receiveMessage(event.data as Uint8Array)
-    )
+    }
+
+    // If we're reconnecting  make sure we remove the old event listeners
+    // before creating a new connection.
+    if (this.socket) {
+      this.socket.removeEventListener("open", onOpen)
```
pvh, line 56: Wouldn't these all be new listeners? You need to capture the original listeners in the onClose. Have you tested to make sure this is actually necessary? I would hope socket shutdown would deal with this internally.

alexjg: Hmm, yeah you're right. For some reason I assumed that javascript function identity was lexical.

We can't use the `onClose` because the particular issue this is guarding against is where the `"open"` event doesn't fire until after the reconnect handler. The sequence of events is this:

* First call to `connect` creates a socket and registers an `"open"` handler. The `"open"` handler will attempt to send a join message
* The connection is slow, so the reconnect timer fires before `"open"`, the reconnect timer calls `connect` again. `this.socket` is now a new socket
* The other end replies to the first socket, `"open"` 
* The original `"open"` handler attempts to send a message, but on the new socket (because it uses `this.socket`), which isn't open yet, so we get an error that the socket is not ready

We can't close a socket which isn't open so we can't use the `"close"` handler to do cleanup

But yeah, this solution doesn't work, clearly I should have written a test. I'll get a fix with a test in later today.

PR #213, opened 2023-10-23
Don't use cbor record extensions

alexjg:


acurrieclark:
Presumable the decoder detects this automatically, so there isn’t anything breaking here?

alexjg:
Yeah for javascript libraries using `cbor-x` this is transparent. For anyone else using cbor this would be a breaking change, but I think that is basically just me.

pvh:
Yeah, this is good. Thanks @alexjg. Better to get this out of the way now than later.

PR #216, opened 2023-10-31
NodeFSStorageAdapter: remove from cache as well as file system

HerbCaudill:
I added some `remove` tests to the storage subsystem, which promptly failed for this adapter because deleted items were still present in the cache. This PR removes deleted items from the cache before deleting them from the filesystem.

PR #215, opened 2023-10-31
NodeFSStorageAdapter: refactor

HerbCaudill:
No changes in behavior, just some minor refactoring for clarity.

PR #217, opened 2023-10-31
Add arbitrary key-value storage

HerbCaudill:
This PR adds support for arbitrary key-value storage to StorageSubsystem. The `load`, `save`, and `remove` methods are for generic key/value storage, and the `loadDoc`, `saveDoc`, and `removeDoc`  methods are for Automerge documents. 

For example, these will be used by the `LocalFirstAuthProvider` to persist the encrypted team graph that encodes group membership and permissions.

```ts
 const namespace = "MyCoolAdapter" 
 const key = "ABC123" 

 await storage.save(namespace, key, value) 
  
 const reloadedValue = await storage.load(namespace, key) 
  
 await storage.remove(namespace, key) 
```

The namespace parameter is to prevent collisions with other users of the storage subsystem.  Typically this will be the name of the plug-in, adapter, or other system that is using it. For example, the LocalFirstAuthProvider uses the namespace `LocalFirstAuthProvider`.

### DIFF
--- a/packages/automerge-repo/src/storage/StorageSubsystem.ts
+++ b/packages/automerge-repo/src/storage/StorageSubsystem.ts
@@ -1,46 +1,96 @@
 import * as A from "@automerge/automerge/next"
 import debug from "debug"
-import * as sha256 from "fast-sha256"
 import { headsAreSame } from "../helpers/headsAreSame.js"
 import { mergeArrays } from "../helpers/mergeArrays.js"
 import { type DocumentId } from "../types.js"
-import { StorageAdapter, StorageKey } from "./StorageAdapter.js"
+import { StorageAdapter } from "./StorageAdapter.js"
+import { ChunkInfo, StorageKey } from "./types.js"
+import { keyHash, headsHash } from "./keyHash.js"
+import { chunkTypeFromKey } from "./chunkTypeFromKey.js"
 
-// Metadata about a chunk of data loaded from storage. This is stored on the
-// StorageSubsystem so when we are compacting we know what chunks we can safely delete
-type StorageChunkInfo = {
-  key: StorageKey
-  type: ChunkType
-  size: number
-}
-
-export type ChunkType = "snapshot" | "incremental"
-
-function keyHash(binary: Uint8Array) {
-  const hash = sha256.hash(binary)
-  const hashArray = Array.from(new Uint8Array(hash)) // convert buffer to byte array
-  const hashHex = hashArray.map(b => ("00" + b.toString(16)).slice(-2)).join("") // convert bytes to hex string
-  return hashHex
-}
-
-function headsHash(heads: A.Heads): string {
-  const encoder = new TextEncoder()
-  const headsbinary = mergeArrays(heads.map((h: string) => encoder.encode(h)))
-  return keyHash(headsbinary)
-}
-
+/**
+ * The storage subsystem is responsible for saving and loading Automerge documents to and from
+ * storage adapter. It also provides a generic key/value storage interface for other uses.
+ */
 export class StorageSubsystem {
-  #storageAdapter: StorageAdapter
-  #chunkInfos: Map<DocumentId, StorageChunkInfo[]> = new Map()
+  /** Record of the latest heads we've loaded or saved for each document  */
   #storedHeads: Map<DocumentId, A.Heads> = new Map()
+
+  /** Metadata on the chunks we've already loaded for each document */
+  #chunkInfos: Map<DocumentId, ChunkInfo[]> = new Map()
+
+  /** Flag to avoid compacting when a compaction is already underway */
+  #compacting = false
+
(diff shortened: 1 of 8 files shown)


HerbCaudill:
* main
      * **PR #217** <a href="https://app.graphite.dev/github/pr/automerge/automerge-repo/217" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/automerge/automerge-repo/217?utm_source=stack-comment).

packages/automerge-repo/src/storage/StorageSubsystem.ts
```diff
@@ -1,46 +1,148 @@
...
+      void this.#saveIncremental(documentId, doc)
+    }
+    this.#storedHeads.set(documentId, A.getHeads(doc))
+  }
+
+  /**
+   * Removes the Automerge document with the given ID from storage
+   */
+  async removeDoc(documentId: DocumentId) {
+    void this.storageAdapter.removeRange([documentId, "snapshot"])
```
pvh, line 139: is it worth doing an indexedDB migration to clean this up and stop squatting on the whole documentId namespace?

packages/automerge-repo/src/storage/StorageSubsystem.ts
```diff
@@ -1,46 +1,148 @@
...
+  /** Metadata on the chunks we've already loaded for each document */
+  #chunkInfos: Map<DocumentId, ChunkInfo[]> = new Map()
 
-function headsHash(heads: A.Heads): string {
-  const encoder = new TextEncoder()
-  const headsbinary = mergeArrays(heads.map((h: string) => encoder.encode(h)))
-  return keyHash(headsbinary)
-}
+  /** Flag to avoid compacting when a compaction is already underway */
+  #compacting = false
```
pvh, line 23: can't prevent another tab from doing this... still probably a good idea but worth testing concurrent compactions

packages/automerge-repo/src/storage/StorageSubsystem.ts
```diff
@@ -1,46 +1,148 @@
...
+  /** Saves a value in storage. */
+  async save(
+    /** Namespace to prevent collisions with other users of the storage subsystem. */
+    namespace: string,
+
+    /** Key to load. Typically a UUID or other unique identifier, but could be any string. */
+    key: string,
 
-  constructor(storageAdapter: StorageAdapter) {
-    this.#storageAdapter = storageAdapter
```
pvh, line 41: why isn't this private anymore?

packages/automerge-repo/src/storage/chunkTypeFromKey.ts
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Keys for storing Automerge documents are of the form:
+ * ```ts
+ * [documentId, "snapshot", hash]  // OR
+ * [documentId, "incremental", hash]
+ * ```
+ * This function returns the chunk type ("snapshot" or "incremental") if the key is in one of these
+ * forms.
+ */
+export function chunkTypeFromKey(key: StorageKey): ChunkType | null {
```
pvh, line 13: i'm not sure there's a good reason to differentiate these anymore. what do you think?

packages/automerge-repo/src/storage/StorageSubsystem.ts
HerbCaudill: Oh I really don't think so. Maybe it would have been cleaner to do it that way from the beginning but if there's not a problem I would just leave it. As long as nobody but the storage subsystem talks directly to the storage adapters it should be fine.

packages/automerge-repo/src/storage/StorageSubsystem.ts
HerbCaudill: The `#` prefix marks class members as private but it's a JavaScript language feature rather than TypeScript so it's enforced at runtime.

packages/automerge-repo/src/storage/StorageSubsystem.ts
HerbCaudill: yeah - maybe create a separate issue to track that?

packages/automerge-repo/src/storage/chunkTypeFromKey.ts
HerbCaudill: It's true that we're not doing anything with the information except to ensure that it's one or the other, but I think the additional clarity doesn't hurt anything & could be helpful for debugging etc.

packages/automerge-repo/src/storage/StorageSubsystem.ts
HerbCaudill: ah never mind, I responded on my phone & wasn't looking at the whole file. I've put that back the way it was.

PR #220, opened 2023-11-01
AuthProvider for localfirst/auth

HerbCaudill:
> 🚧This is a work in progress. Opening the PR for the purpose of discussion. 

This is an `AuthProvider` that uses the [localfirst/auth](https://github.com/local-first-web/auth)
library to provide Automerge Repo with authentication and end-to-end encryption, without the need
for a central server.

## How to use it

A `LocalFirstAuthProvider` is configured with information about the local user and device.

```ts
import { createUser, createDevice } from "@localfirst/auth"
import { LocalFirstAuthProvider } from "@automerge/automerge-repo-auth-localfirst"

// Create the user & device, or retrieve them from storage.
// These objects include secret keys, so need to be stored securely.
const alice = createUser("alice")
const aliceLaptop = createDevice(alice.userId, "ALICE-MACBOOK-2023")
const aliceContext = { user: alice, device: aliceLaptop }

const aliceAuthProvider = new LocalFirstAuthProvider(aliceContext)

const repo = new Repo({
  network: [SomeNetworkAdapter],
  storage: SomeStorageAdapter,
  auth: aliceAuthProvider,
})
```

The context for authentication is a localfirst/auth `Team`. For example, Alice might create a team
and invite Bob to it.

```ts
// Alice creates a team
const aliceTeam = Auth.createTeam("team A", aliceContext)
aliceAuthProvider.addTeam(aliceTeam)

// Alice creates an invitation code to send to Bob
const { seed: bobInviteCode } = aliceTeam.inviteMember()
```

Alice now needs to communicate this code, along with the team's ID, to Bob, using an existing
communications channel that she trusts. For example, she could send it via WhatsApp or Signal or
email; or she could create a QR code for Bob to scan; or she could read it to him over the phone.

Bob sets up his auth provider and his repo much like Alice did:

```ts
const bob = createUser("bob")
const bobLaptop = createDevice(bob.userId, "BOB-IPHONE-2023")
const bobContext = { user: bob, device: bobLaptop }

const bobAuthProvider = new LocalFirstAuthProvider(bobContext)

const repo = new Repo({
  network: [SomeNetworkAdapter],
  storage: SomeStorageAdapter,
  auth: bobAuthProvider,
})
```

Rather than add a `Team` to the provider, Bob registers his invitation:

```ts
bobAuthProvider.addInvitation({
  shareId: aliceTeam.id,
  invitationSeed: bobInviteCode,
})
```

[...]

PR #221, opened 2023-11-03
Clean up & refactor Repo

HerbCaudill:
The motivation for merging the DocCollection and Repo classes in #98 was to simplify the logic and make the flow of execution easier to follow and reason about.

Opportunities for simplification still remained: We were raising DocCollection's `document` and `delete-document` events in Repo, and listening for them as we did before, even though both the emitter and the listener were in the same class.

- The logic from the `document` event handler has been moved to a private method, `#registerHandle({handle, isNew})`. This is called from `Repo.create` and `Repo.find`.
- The contents of the `delete-document` event handler have been inlined in `Repo.delete`.

I've done some further cleaning up in related files that I've piggybacked onto this:

- I've tried to make all async logic as readable as possible -- by replacing `.then` callbacks with `await`, using helpers like `pause` and `eventPromise`, and so on.
- DocHandle's events all included the handle itself in the event payload, which I've removed as they're unnecessary: You obligatorily have a reference to the handle if you're handling its events.
- We were swallowing errors in a couple of spots -- when waiting for the network to be ready, and when deleting documents. I'm not sure what the thinking there is, but it seems like the application might want to know about failures in either of those cases, so I've removed the `.catch` clauses.

There should be no changes in behavior other than that last one.


HerbCaudill:
* main
        * **PR #218** <a href="https://app.graphite.dev/github/pr/automerge/automerge-repo/218" target="_blank">Graphite</a> 
          * **PR #221** <a href="https://app.graphite.dev/github/pr/automerge/automerge-repo/221" target="_blank">Graphite</a>  👈
          * **PR #219** <a href="https://app.graphite.dev/github/pr/automerge/automerge-repo/219" target="_blank">Graphite</a> 
            * **PR #220** <a href="https://app.graphite.dev/github/pr/automerge/automerge-repo/220" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/automerge/automerge-repo/221?utm_source=stack-comment).

packages/automerge-repo/src/Repo.ts
```diff
@@ -44,229 +48,175 @@ export class Repo extends EventEmitter<RepoEvents> {
...
-
   /**
-   * Creates a new document and returns a handle to it. The initial value of the document is
-   * an empty object `{}`. Its documentId is generated by the system. we emit a `document` event
-   * to advertise interest in the document.
+   * When we create a new document or look up a document by ID, wire up storage and network
+   * synchronization.
    */
-  create<T>(): DocHandle<T> {
-    // TODO:
```
HerbCaudill, line 181: I'll create an issue to track this TODO

HerbCaudill:
* main
        * **PR #218** <a href="https://app.graphite.dev/github/pr/automerge/automerge-repo/218" target="_blank">Graphite</a> 
          * **PR #221** <a href="https://app.graphite.dev/github/pr/automerge/automerge-repo/221" target="_blank">Graphite</a>  👈
            * **PR #219** <a href="https://app.graphite.dev/github/pr/automerge/automerge-repo/219" target="_blank">Graphite</a> 
              * **PR #222** <a href="https://app.graphite.dev/github/pr/automerge/automerge-repo/222" target="_blank">Graphite</a> 
                * **PR #220** <a href="https://app.graphite.dev/github/pr/automerge/automerge-repo/220" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/automerge/automerge-repo/221?utm_source=stack-comment).

acurrieclark:
I agree that the handle isn't required on the change event object, but I think removing that at this stage would technically be breaking.

HerbCaudill:
> I agree that the handle isn't required on the change event object, but I think removing that at this stage would technically be breaking.

That's true. I figured it was really unlikely that anyone is depending on that, but I suppose maybe now that automerge-repo is at v1 and has been officially announced we need to be strict about this. 

@pvh let me know what you think: Currently `docHandle` includes a reference to itself in its event payloads, so you could have code in the wild like 

```ts
handle.on("change", ({ doc, handle }) => {
  console.log(handle)
})
```

which would break with these changes. 

If necessary I can pull those changes out into a separate pr that would be intended for the next major version.

pvh:
I do indeed have code which does that.

HerbCaudill:
OK I'll revert those changes. 

Out of curiosity, is there a reason I'm missing why you'd want to pass the emitter itself in the event payload? As opposed to using the reference that you already have. 

```ts
handle.on("change", ({ doc }) => {
  console.log(handle)
})
```

acurrieclark:
I think it's a hangover from the earlier version of Repo where the only property returned in the change event object was the handle. I agree that there is no reason to keep it except for the fact that it would be breaking. I would definitely vote to remove it.

If you have destructured the change event object and your handle is still called `handle` then it won't break anything, but in any other case I think it will.

HerbCaudill:
closing, will open revised PR against latest repo code

PR #223, opened 2023-11-03
Peer sync report paul

paulsonnentag:
Persist sync state and add API to read the sync state of peers.

- Store sync state
- Dynamically load sync state as peers connect from storage.
- Add API to access the synced heads of remote peers (currently only works for directly connected peers)


packages/automerge-repo-network-broadcastchannel/package.json
pvh, line 1: @alexjg I removed a bunch of these peerDependencies since they don't seem to be accurate or important here. I don't think we need to specify specific versions of Automerge these libraries work with, since we're already saying which version of Automerge Repo you need and that depends on Automerge core.

pvh: (Oh, I tagged you because I'm hoping you will either agree or disagree.)

packages/automerge-repo/test/Repo.test.ts
```diff
@@ -768,6 +787,139 @@ describe("Repo", () => {
...
+        A.getHeads(charlieHandle.docSync())
+      )
+
+      teardown()
+    })
+
+    it("can report the connected peers", async () => {
+      const { bobRepo, charlieRepo, teardown } = await setup()
+
+      assert.deepStrictEqual(bobRepo.peers, ["alice", "charlie"])
```
pvh, line 918: This test is failing here. @paulsonnentag is that expected?

packages/automerge-repo/src/Repo.ts
```diff
@@ -172,6 +183,11 @@ export class Repo extends EventEmitter<RepoEvents> {
     return this.#handleCache
   }
 
+  /** Returns a list of all connected peer ids */
+  get peers(): PeerId[] {
+    return this.#synchronizer.peers
```
pvh, line 188: Do we need this? I'm not sure a repo *has* peers, only documents do.

packages/automerge-repo/src/synchronizer/CollectionSynchronizer.ts
```diff
@@ -35,8 +35,21 @@ export class CollectionSynchronizer extends Synchronizer {
 
   /** Creates a new docSynchronizer and sets it up to propagate messages */
   #initDocSynchronizer(handle: DocHandle<unknown>): DocSynchronizer {
-    const docSynchronizer = new DocSynchronizer(handle)
+    const docSynchronizer = new DocSynchronizer({
+      handle,
+      onLoadSyncState: peerId => {
```
pvh, line 40: This is a little janky; I'll see if I can find a cleaner way to do this but I don't hate it.

packages/automerge-repo/src/DocHandle.ts
```diff
@@ -327,6 +328,19 @@ export class DocHandle<T> //
     })
   }
 
+  /** `setRemoteHeads` is called by the doc synchronizer
+   * @hidden
+   */
+  setRemoteHeads(peerId: PeerId, heads: A.Heads) {
```
pvh, line 334: I don't love this API for reasons we've already discussed but I'll put up with it for now.

packages/automerge-repo/test/Repo.test.ts
pvh: Added a timeout and now it's good.

PR #231, opened 2023-11-15
WIP: Stable peer ids

paulsonnentag:
- Create a stable storage ID that's generated once for each storage adapter. The sync state is persisted per storage ID. This makes the persisted sync state more useful because the same sync state can be reused even when the peer ID changes. Later this will make it also easier to assign sync state to logical entities like people and devices.

- On connect, each peer transmits:
  - their peerId
  - their storageId (if they have a storage adapter)
  - an isEphemeral flag (sync state of ephemeral peers will not be persisted)

### DIFF
--- a/packages/automerge-repo-network-broadcastchannel/src/index.ts
+++ b/packages/automerge-repo-network-broadcastchannel/src/index.ts
@@ -14,7 +14,12 @@
  *
  */
 
-import { Message, NetworkAdapter, PeerId } from "@automerge/automerge-repo"
+import {
+  NetworkAdapter,
+  type Message,
+  type PeerId,
+  type StorageId,
+} from "@automerge/automerge-repo"
 
 export type BroadcastChannelNetworkAdapterOptions = {
   channelName: string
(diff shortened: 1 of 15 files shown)


pvh:
One small note before I review properly; I think `isEphemeral` might be a bit of a confusing name for a new reader. Maybe `hasStorage`? Is there a way to say what we actually want the peer to do: `rememberMe`? `illBeBack`?

acurrieclark:
Couple of quick thoughts:

If a number of peers (eg. tabs in a browser) are sharing the same storage, it looks like they will all have the same storage ID. Presumably that would stop them all being able to sync with one another?

This PR changes the `NetworkAdapter` contract, which I would be breaking for anyone who has written their own network adapter.

I think that (for now) this could be worked around by fixing the Repo's `peerId` and storing that locally?

alexjg:
Re. the breaking change. You're referring to the fact that previously `NetworkAdapter.connect` was of type `(peerId): void` and now it's `(peerId, storageId, boolean): void`? Firstly, I think we can just use default parameters for `storageId` and `isEphemeral` to avoid the typescript compile errors that will arise. From the perspective of existing `NetworkAdapter` implementations then this is fine I think. The peer ID is still use to identify the network connection and the storage ID is not used by old implementations anyway.

That said, I think we should probably accept `storageId` and `isEphemeral` as keys of an object to reduce the risk of future compatibility hazards. I.e. `NetworkAdapter.connect` should probably look like `connect(peerId, {storageId, isEphemeral})`.

Regarding multiple peers using the same storage. That's fine. They will each use a different peer ID (peer IDs being ephemeral things which are created when the repo starts up) and so they'll all be able to sync with each other.

acurrieclark:
I see. As long as the rest of the system operates as before if the adapter doesn't return a storageId, then I agree it wouldn't be breaking.

Presumably a peer sharing storage with another peer would need to be explicitly marked as `isEphemeral`? (Or whatever name we settle on). Would that be breaking?

PR #232, opened 2023-11-15
remove "failed" state

pvh:
The "failed" state is final -- it means that if we hit a timeout we give up permanently. That's not ideal because it has no path to recovery. This has been causing problems in trail-runner which has enough data that sometimes documents might take more than 60s to load, at which point no recovery is possible without killing and rebooting the service worker.

There's undoubtedly more work to do here, but by removing the "failed" state, we at least admit the possibility of recovery here.


pvh:
Merged so I can go back to using trail-runner for now, but asking @acurrieclark for additional input post-merge.

PR #234, opened 2023-11-21
Allow a broadcastChannel name to be properly set in the adapter

acurrieclark:
When I implemented this originally, I somehow managed to write a test which didn't test anything I was trying to achieve. No surprise that the feature didn't actually work.

This should remedy the issue.


pvh (approved):

pvh:
Thanks Alex!

PR #237, opened 2023-11-30
Make `isEphemeral` optional in network adapter's `peer-candidate` event. 

HerbCaudill:
Otherwise this is a breaking change.

In the listener in `NetworkSubsystem`, defaults to false; I assume this is what we'd want.


pvh:
Actually, I'll just merge this one anyway.

PR #238, opened 2023-11-30
NetworkAdapter comments

HerbCaudill:
The "hmmm maybe not" there has made me question reality for long enough.

### DIFF
--- a/packages/automerge-repo/src/network/NetworkAdapter.ts
+++ b/packages/automerge-repo/src/network/NetworkAdapter.ts
@@ -11,7 +11,7 @@ import { StorageId } from "../storage/types.js"
  * until the adapter emits a `ready` event before it starts trying to use it
  */
 export abstract class NetworkAdapter extends EventEmitter<NetworkAdapterEvents> {
-  peerId?: PeerId // hmmm, maybe not
+  peerId?: PeerId
   storageId?: StorageId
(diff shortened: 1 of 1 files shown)


packages/automerge-repo/src/network/NetworkAdapter.ts
```diff
@@ -11,27 +11,25 @@ import { StorageId } from "../storage/types.js"
  * until the adapter emits a `ready` event before it starts trying to use it
  */
 export abstract class NetworkAdapter extends EventEmitter<NetworkAdapterEvents> {
-  peerId?: PeerId // hmmm, maybe not
```
pvh, line 14: The "hmm, maybe not" here is that the network adapter gets its peerId at connect() time but we have it on the constructor. Perhaps we should clean that up?

HerbCaudill: ? It's not on the constructor - the abstract class doesn't specify a constructor interface, and none of the implementations take a peerId in the constructor, only in `connect()`.

pvh: yes -- of course, my misread there. thanks.

pvh:
I manually reimplemented this due to the conflicts. At least I got rid of the "hmm".

PR #239, opened 2023-12-01
console log when we exceed 1000ms for a load

pvh:
Any objection? I usually avoid checking in `console.warns()` but at least at the moment I want to emphasize finding and fixing these slow documents.


acurrieclark:
I can absolutely see the value in this. Is it worth making it optional?

pvh:
yea, it should probably be tied to a debug setting. i'll update the patch.

pvh:
I'm going to drop this PR -- I still think it's a good idea and we can revive it later but it's not an active problem and I think building on the perfLogger model would be a better plan.

PR #240, opened 2023-12-03
add repo test for cloning loaded docs

kid-icarus:
Feel free to close this out, I'm adding this here as an example of the issue I'm running into.  I'd like to load a saved document, and create a new handle for the loaded document, but the only way I can think of is via creating a new handle and merging the loaded document in a `handle.change`. Unfortunately, that ends up with a squashed history, as this test shows.

Would it make sense for `repo.create` to have a document parameter? It'd be nice to do something to the effect of:

```ts
repo.create(A.load(someSavedDoc))
```

### DIFF
--- a/packages/automerge-repo/test/Repo.test.ts
+++ b/packages/automerge-repo/test/Repo.test.ts
@@ -146,6 +146,46 @@ describe("Repo", () => {
       assert.deepStrictEqual(handle2.docSync(), { foo: "bar", baz: "baz" })
     })
 
+    it("retains history when creating a handle to a loaded document", async () => {
+      const { repo } = setup()
+      const handle = repo.create<TestDoc>()
+      handle.change(d => {
+        d.foo = "foo"
+      })
+      handle.change(d => {
+        d.foo = "bar"
+      })
+      handle.change(d => {
+        d.foo = "baz"
+      })
+      handle.change(d => {
+        d.foo = "a"
+      })
+      handle.change(d => {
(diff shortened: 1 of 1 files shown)


kid-icarus:
The solution for me was:

1. Create a doc - `const doc = A.load(payload)`
2. Create a handle - `const handle = repo.create()`
3. Set the document - `handle.update(() => doc)`

That last bit, "setting" the document via [handle.update](https://github.com/automerge/automerge-repo/blob/main/packages/automerge-repo/src/DocHandle.ts#L320-L327), was unintuitive. Thanks a ton for your suggestion, @acurrieclark.

The fact that `update` is hidden per the JSDoc tag, combined with the comment, "update is called by the repo when we receive changes from the network," feels like I'm breaking convention. If the intention is to not expose `update` to consumers, it seems sensible to add a convenience method as if you've proposed.

I'll go ahead and close this out.

pvh:
No! I think this is a good feature request -- I think we should probably 1) make update a supported interface and 2) add a specific API call for this use-case (and for saving!) so you don't have to figure it out for yourself. 

Thinking about the API, it seems like `repo.import(buffer)` would make the most sense to me. What do you think, @kid-icarus?

kid-icarus:
Sounds great! Looks like @acurrieclark opened a PR to add it, thanks again! :)

kid-icarus:
Going to close this out in light of #251 :)

Issue #228, opened 2023-11-09
Don't bundle React with the hooks package


kid-icarus:
I was unable to use the package directly, as you're bundling React with the package itself, and unfortunately, I'm still on React 17. This leads to a situation where the `useRepo` context provider always returns `null`. For now, I've copied the hooks into my repo.

In general, as a library author you typically don't want to bundle react with the package. This is typically accomplished via whatever you're using for bundling/building the lib. For Webpack, you can use the [externals](https://webpack.js.org/configuration/externals/) prop in webpack config. I see you're using Vite, and I believe there are plugins for vite and/or rollup, such as [vite-plugin-externals](https://github.com/crcong/vite-plugin-externals)

pvh:
Hey, good call. Thanks! I'll fix this.

pvh:
Actually, if you want to send a patch that solves the problem for you, I can test it as well. Otherwise I'll probably end up just guessing at the solution and then asking you to confirm anyway :)

kid-icarus:
https://github.com/automerge/automerge-repo/pull/241

msakrejda:
This can be closed, right?

kid-icarus:
Hey @msakrejda, just got around to testing this. All works as expected, closing!

PR #241, opened 2023-12-03
build hooks package with vite and stop bundling react

kid-icarus:
This ensures that consumers use their own version of React instead of bundling potentially multiple, conflicting versions. Addresses https://github.com/automerge/automerge-repo/issues/228

Additionally, I've added a `yarn visualize` command that allows you to visualize the bundle's size. You can see from these screenshots. As you can see here, we've reduced the bundle's size by roughly 10x via excluding react and react-dom out of the bundle.

Screenshot 2023-12-02 at 8 45 21 PM

Screenshot 2023-12-02 at 8 47 38 PM


pvh:
Thank you! I've been focused on some other work for a bit but I should have some time this week to put into this and a few other PRs that have been accumulating.

pvh:
Landed; thanks!

PR #251, opened 2023-12-06
Adds `import` method to add an existing document to the repo

acurrieclark:
Following discussion in #240, I thought I would add an import method to repo for adding an existing document to the repo.

Points for discussion:

1. Should the parameter be an automerge document, or a Unit8Array? I have made it a document for now, as this made most sense to me, and as it allows us to check the validity of the document and have it typed correctly.
2. What should the error message be which is thrown if an invalid document is passed in.


kid-icarus:
Added an export method in #256

acurrieclark:
Updated the signature to take a `Uint8Array` instead and added the method to the readme

PR #256, opened 2023-12-12
feat(automerge-repo): add export method

kid-icarus:
Following up from https://github.com/automerge/automerge-repo/pull/251, this adds an export convenience method that will export a document without having to dip into automerge's `A.save()`

### DIFF
--- a/packages/automerge-repo/README.md
+++ b/packages/automerge-repo/README.md
@@ -45,6 +45,8 @@ A `Repo` exposes these methods:
   networks.
 - `delete(docId: DocumentId)`  
   Deletes the local copy of a document from the local cache and local storage. _This does not currently delete the document from any other peers_.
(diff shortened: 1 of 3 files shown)


acurrieclark:
Good idea.

I think it's probably worth mentioning in the readme that this returns a promise, which could resolve to undefined too.

kid-icarus:
@acurrieclark Updated readme and JSDoc of the function, let me know if you have any suggestions to improve the wording.

acurrieclark:
Looks good to me!

pvh:
I think `export` should probably be a method on a handle for consistency. I'll update the patch accordingly.

kid-icarus:
Ah, I agree having the export method live on the handle makes more sense. Thanks @pvh!

PR #257, opened 2023-12-12
chore: add lint action

kid-icarus:
It seems like our good friend `eslint` has not been run in a while :) I've added a GitHub action that runs eslint, and I've also cleaned up or ignored relevant errors so that the action will pass.

There are a number of warnings that we can clean up as well, but I think this is a good start. Additionally, if we want to add eslint configuration to other packages within the monorepo, we can simply add a `lint` yarn script, and lerna is smart enough to run the command on packages that have changed.


pvh:
I like it, thanks!

PR #258, opened 2023-12-12
improve handling of changing URLs in useDocument and useHandle

geoffreylitt:
Previously, useDocument did not behave well for documents that took a long time to load:

1) while a load was blocking the UI, an old stale doc would be shown (potentially for multiple seconds for large docs).
2) loads that finished out of order could result in an inconsistent final state.

We have seen this cause problems in the "DocConnector" sidebar in trail-runner.

Below are demos of the two issues with a [demo app](https://github.com/inkandswitch/tiny-essay-editor/tree/knapsack/src/knapsack/components) based on the trail-runner sidebar. When you click an entry on the left, it loads a new doc URL and shows it in the main pane. The main pane shows the doc URL as well as the doc contents. (I've assigned each url a color to make visual inspection easier.)

1) stale docs during loads: while the Embark essay loads, we see the old doc still:

https://github.com/automerge/automerge-repo/assets/934016/2c2d6e65-8c00-4792-b85f-b47ada9da0e9

2) when I click on the slow-loading Embark essay and then another already-loaded doc, we end up with the Embark essay doc contents, which don't match the doc url that's supposed to be shown in the main pane:

https://github.com/automerge/automerge-repo/assets/934016/b1f9e2cf-83ad-4f85-a456-b58df3a1df1f

## solution

- Upon a handle changing, the old doc contents are cleared out to avoid staleness inconsistencies.
- Out-of-order loads are accounted for; when a load finishes it's ignored if the handle has since changed.

Here are demos of the fixed behavior in the same two cases shown above:

1) Now while the Embark essay loads we see a loading state in the main pane because the doc url is set to undefined:

https://github.com/automerge/automerge-repo/assets/934016/f76d3c6c-59ca-4834-8dd2-085c430f4897

2) When we click the same out of order as before, we end up in the correct consistent state. (When the Embark essay finishes loading, that's ignored because the handle has moved on. You can't fully see this visually but I've confirmed it with console logs)

https://github.com/automerge/automerge-repo/assets/934016/f9b416c3-ff9a-42d9-a71f-fe470c0d9b73

## todos

- need to figure out how to do automated tests for this. 
- I saw this broke one existing test, need to see if that's a real issue or not.

### DIFF
--- a/packages/automerge-repo-react-hooks/src/useDocument.ts
+++ b/packages/automerge-repo-react-hooks/src/useDocument.ts
@@ -1,6 +1,6 @@
 import { ChangeFn, ChangeOptions, Doc } from "@automerge/automerge/next"
 import { AutomergeUrl, DocHandleChangePayload } from "@automerge/automerge-repo"
-import { useLayoutEffect, useRef, useState } from "react"
+import { useEffect, useRef, useState } from "react"
 import { useRepo } from "./useRepo.js"
 
 /** A hook which returns a document identified by a URL and a function to change the document.
@@ -18,13 +18,9 @@ export function useDocument<T>(
   const repo = useRepo()
 
   const handle = documentUrl ? repo.find<T>(documentUrl) : null
-
   const handleRef = useRef(null)
 
-  // Doing this in a useLayoutEffect seems to help make sure that
-  // a loading state gets rendered before we block the UI thread
-  // with a slow load.
-  useLayoutEffect(() => {
+  useEffect(() => {
     // When the handle has changed, reset the doc to an empty state.
     // This ensures that if loading the doc takes a long time, the UI
     // shows a loading state during that time rather than a stale doc.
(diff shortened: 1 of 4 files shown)


geoffreylitt:
Made a few tweaks:

- switched back from `useLayoutEffect` to `useEffect` because it was causing problems in tests and wasn't clear it was making any difference. (I was trying to coax React into scheduling differently around slow blocking receiveSyncMessage calls, but we should just stop blocking the UI thread for seconds instead.)
- Added some unit tests
- QA: confirmed my demo videos from above still work, and also tried TEE and the todos app in automerge-repo, all seems well

geoffreylitt:
Also I guess it's worth noting that this PR is a breaking change to the hook behavior, not sure if we're trying to follow any particular versioning policy on this project.

paulsonnentag:
Looks good to me!

PR #260, opened 2023-12-14
put remote heads gossiping behind a feature flag

alexjg:
I think we're not yet sure whether the remote heads gossiping feature is fully baked, but we have other fixes which I think we should release. Put the feature behind a feature flag so we can continue developing it on `main` without blocking release.


paulsonnentag:
I think this is a good idea. Should we log a warning if you call subscribeToRemotes and don't have remote heads gossiping on?

alexjg:
Good shout, done.

PR #268, opened 2024-01-11
Remove onInvalidAutomergeUrl from useBootstrap

neftaly:
This PR removes `onInvalidAutomergeUrl` from useBootstrap.

The alternative is to use a react error boundary. From my understanding, this wasn't possible at the time.


neftaly:
I would like to be able to do something like `error instanceof InvalidAutomergeUrlError` or similar in the error boundary, but this would require a separate PR.

pvh:
I think this is a great idea but is technically a breaking change to the API. I've been hesitant to merge it because of that even though I suspect nobody will be using it.

pvh:
Also, it's worth noting that passing in a bad AutomergeUrl should emit an error from the DocHandle so this is probably totally redundant anyway.

pvh:
Moot with usebootstrap gone.

PR #269, opened 2024-01-11
fix: worker.plugins is now a function warning

openscript:
When starting one of the examples vite throws the following warning:

> worker.plugins is now a function that returns an array of plugins. Please update your Vite config accordingly.

This PR updates the Vite config accordingly.


pvh:
Thanks!

PR #272, opened 2024-01-17
Fix types on react hooks

HerbCaudill:
- On `useDocument` and `useHandle`, changes the type of the id parameter from `AutomergeUrl` to `AnyDocumentId`, since that's what `repo.find()` takes. 
- Enables strict type checking for this package, and adds/modifies types as necessary.

### DIFF
--- a/packages/automerge-repo-react-hooks/src/useBootstrap.ts
+++ b/packages/automerge-repo-react-hooks/src/useBootstrap.ts
@@ -28,10 +28,14 @@ export const useHash = () => {
 
 // Get a key from a query-param-style URL hash
 const getQueryParamValue = (key: string, hash: string) =>
-  new URLSearchParams(hash.substr(1)).get(key)
+  new URLSearchParams(hash.slice(1)).get(key)
 
-const setQueryParamValue = (key: string, value, hash): string => {
-  const u = new URLSearchParams(hash.substr(1))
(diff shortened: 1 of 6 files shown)


packages/automerge-repo-react-hooks/src/useBootstrap.ts
```diff
@@ -28,10 +28,14 @@ export const useHash = () => {
 
 // Get a key from a query-param-style URL hash
 const getQueryParamValue = (key: string, hash: string) =>
-  new URLSearchParams(hash.substr(1)).get(key)
+  new URLSearchParams(hash.slice(1)).get(key)
```
HerbCaudill, line 31: `substr` is deprecated.

packages/automerge-repo-react-hooks/src/useDocument.ts
```diff
@@ -58,5 +60,5 @@ export function useDocument<T>(
     handle.change(changeFn, options)
   }
 
-  return [doc, changeDoc]
+  return [doc, changeDoc] as const
```
HerbCaudill, line 63: This tells TypeScript to infer that the return value is a tuple and not an array, which frees us up from having to annotate the return type manually (which is good because it's easy to get the return type annotation slightly wrong, as it was in this case).

PR #282, opened 2024-01-25
test: Add n-peers chain test

mykola-vrmchk:


acurrieclark:
This is an interesting test, which I will dive in to further when I have the time. For now, could I suggest that to add latency to the network adapter, you create a test version which wraps the existing adapter but adds the latency, rather than updating the main adapter config?

I can't think of a reason why it would be useful to have latency configurable on the adapter outside of this test.

acurrieclark:
Looks like the tests aren't running correctly as there a couple of `only` tests still in there

mykola-vrmchk:
@acurrieclark I refactored a test a bit, could you take another look? I see a value in it being merged. I can make any changes you request. And yes, it seems like https://github.com/automerge/automerge-repo/pull/290 fixed it.

acurrieclark:
Thanks for putting in the extra work on this. I like the additional test.

@alexjg any thoughts?

Issue #266, opened 2024-01-02
"Error: DocHandle is not ready" with useBootstrap


neftaly:
I encountered a sharp edge building a simple `useBootstrap` app. I must have initialized useBootstrap with an empty or broken `onNoDocument` function, which resulted in the document ID being saved to localstorage, but the document itself not being created.

Whenever I tried to write anything, I got the following console.error:
```Uncaught Error: DocHandle#... is not ready. Check `handle.isReady()` before accessing the document.```

It was not immediately obvious what was happening, and the error persisted between reloads. I have encountered this before and would like it to be easier to diagnose for someone unfamiliar with the library. 
* The error message could provide a link to a documentation page with more information?
* The chrome devtools could log a message like "useBootstrap: loading existing documentId from localstorage"?

neftaly:
For anyone encountering this error, what is happening is that automerge-repo is waiting for a peer to share the document (which won't happen as it doesn't exist). The solution is to delete the `automergeUrl` key from localstorage and the hash URL, or to change useBootstrap.key to something different.

neftaly:
A simple way to recreate this is to use `useBootstrap` without a repo storage adapter. This should be a supported use-case (I don't want persistent storage while I'm testing things).

neftaly:
I think a documentation page would be the best bet, with advice on disabling storage/hashurls for development.

msakrejda:
Closing this since `useBootstrap` has been removed.

PR #285, opened 2024-02-01
Remove useBootstrap

neftaly:
This PR removes localStorage (document URL persistence) from useBootstrap. It is intended to enable useBootstrap to be used without a storage adapter, as per #266.


neftaly:
I think it would be nice to have some logic that detected if storage adapters were defined, though I want to keep the scope small and figure out if this is worth it later. It would be really nice to persist the document URL using whatever storage adapters were in use, but it would probably be a bit too hairy. 

I would like to figure out if the URLs are causing a similar sort of issue w.r.t persistence in a development environment, however I suspect this would be better solved elsewhere in the tooling.

neftaly:
It would be helpful to show an example of localStorage URL persistence in a tutorial or cookbook or similar

pvh:
We talked about this a while back but I think we ought to just deprecate useBootstrap entirely. I agree that the localStorage is the pits but the longer I go, the more I lean towards doing away with it. What do you think?

neftaly:
Sounds good, I would rather this existed in a tutorial or example.

neftaly:
There is still an example using useBootstrap in `packages/automerge-reac-hooks/README.md` (and the identical comment on `packages/automerge-reac-hooks/index.ts`).

I don't have the bandwidth to fix the example though sorry! But it's not urgent and shouldn't affect anything.

pvh:
I had a heap of rebase conflicts so I just reimplemented this but it's done now. Thanks @neftaly.

PR #286, opened 2024-02-01
Quick fixes

acurrieclark:
The broadcastchannel tests was using the old `connect` signature, before `peerMetaData` was added. Added an empty object for now

Tidied up a TODO to remove a ts-ignore


packages/automerge-repo-network-broadcastchannel/test/index.test.ts
```diff
@@ -36,9 +36,9 @@ describe("BroadcastChannel", () => {
       }, 100)
     })
 
-    a.connect("a" as PeerId)
-    b.connect("b" as PeerId)
-    c.connect("c" as PeerId)
+    a.connect("a" as PeerId, {})
```
alexjg, line 39: Is this actually a backwards incompatible change that we made to `NetworkAdapter`? If so it seems to me like we should instead add a default argument for the peer metadata.

acurrieclark: That sounds like a more sensible path to me. I shall make that change instead

acurrieclark: Updated the adapters to make peerMetadata optional

PR #287, opened 2024-02-01
Adapter peer metadata fix

acurrieclark:
Some of the network adapters were not correctly sending `peerMetadata`, nor were they properly passing the correct `peerMetadata` back to the network subsystem.

This fixes the bug and adds a test to the adapter acceptance test suite


acurrieclark:
Tests have failed because the message type hasn't been narrowed properly, which is odd. Will look into it and fix

acurrieclark:
Well, that escalated quickly. TSC was using stricter rules that my editor, which surprised me a little. I have enable strict mode in all the tsconfigs (it was already enabled in most) so that editors catch issues first.

PR #289, opened 2024-02-05
fix: don't leave timers pending for 2 minutes on each waitForState call

yarolegovich:
* doc can't go from ready to unavailable state
* doc() method starts a wait promise for each awaitStates entry (`awaitStates=[ready,unavailable]` by default) 
* for ready state documents that keeps creating unnecessary timers each with 2 minute timeouts 
* for some test runners this causes node process to hang until all timers resolve. we had to add [forceExit](https://github.com/dxos/dxos/blob/main/packages/apps/composer-app/project.json#L63) in our code after migrating to automerge as tests were not exiting

### DIFF
--- a/packages/automerge-repo/test/DocHandle.test.ts
+++ b/packages/automerge-repo/test/DocHandle.test.ts
@@ -1,7 +1,7 @@
 import * as A from "@automerge/automerge/next"
 import assert from "assert"
 import { decode } from "cbor-x"
-import { describe, it } from "vitest"
+import { describe, it, vi } from "vitest"
 import { generateAutomergeUrl, parseAutomergeUrl } from "../src/AutomergeUrl.js"
 import { eventPromise } from "../src/helpers/eventPromise.js"
 import { pause } from "../src/helpers/pause.js"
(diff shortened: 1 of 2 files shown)


packages/automerge-repo/test/DocHandle.test.ts
```diff
@@ -72,6 +72,21 @@ describe("DocHandle", () => {
     assert.equal(doc?.foo, "bar")
   })
 
+  it("no pending timers after a document is loaded", async () => {
```
alexjg, line 75: I feel like this test is a little brittle because it's entirely possible we would add other timers - e.g. something which announces presence every so often on a DocHandle. I understand why this test is here though, we don't currently have an API for saying "everything stop now". I think we should just add a comment that once we have a `Repo.stop` API and accompanying tests then this test will no longer be necessary.

acurrieclark:
Thank you for this PR. We certainly shouldn't be leaving unnecessary timers running once the state has settled. 

The change is great, if you could take a little look at the test we can get this merged.

acurrieclark (approved):

PR #290, opened 2024-02-05
fix: peer doesn't receive a document on connecting with another peer

yarolegovich:
Fixes the following case:
* Alice has a document that Bob wants to discover.
* Bob appears as a peer candidate and Alice sends a sync message.
* Bob is recorded in the list of known peers, but sync message processing fails (in the new test this happens because Alice isn't yet in the list of known peers).
* Alice appears as peer candidate but Bob decides to skip sync, because he already saw Alice (even though her document status is recorded as "has" and he doesn't have a document).


pvh:
All merged, thanks for spotting this.

Issue #265, opened 2024-01-01
Any reason for `StorageAdapter` to be an abstract class as opposed to an interface


sroze:
Just wondering why would [`StorageAdapter` be an abstract class](https://github.com/automerge/automerge-repo/blob/5269a79f4f472a0653f39888b8941e49a4586226/packages/automerge-repo/src/storage/StorageAdapter.ts#L9) and not an interface? The class forces us to `extends` something from this repository while an interface would allow more freedom.

PR #291, opened 2024-02-06
Add network and storage adapter interfaces

acurrieclark:
#265 points out that using interfaces for our adapters, rather than forcing to extend our own abstract adapters, would make the creation of new adapters more accessible.

This adds an interface for both network and storage adapters and updates the `RepoConfig` type to accept them rather than the abstract type.

I don't think this is breaking, but would appreciate a review. Specifically, I don't know where it is best to leave the JSDoc comments; on the abstract class or the interface? Left them on both for now.


alexjg:
I like this change. The abstract classes don't have any implementations and I don't think we rely on instanceof checks anywhere, should we just mark them as deprecated?

acurrieclark:
That would be my instinct, yes.

I have done some tests, and it looks like the comments don't propagate from the interface to the concrete implementation if they are removed from the abstract class, so we might need to leave them on both.

alexjg:
In that case let's leave comments on both but add a deprecation notice to the abstract classes using `@deprecated` with a `{@link}` to the interface to use instead.

acurrieclark:
Links added with `@deprecated`. Looks how I would expect, but I am no expert at JSDoc.

packages/automerge-repo/src/network/NetworkAdapter.ts
```diff
@@ -1,29 +1,24 @@
...
- *
- */
-export interface PeerMetadata {
-  storageId?: StorageId
-  isEphemeral?: boolean
-}
+import { NetworkAdapterInterface } from "./NetworkAdapterInterface.js"
 
 /** An interface representing some way to connect to other peers
+ * @deprecated
```
alexjg, line 10: I think this should be more like `@deprecated use {@link NetworkAdapterInterface} instead`. You can run `npx typedoc` to see what the output looks like.

acurrieclark:
There are a few warnings from running `typedoc`, but none relating to this PR following your suggestions

PR #294, opened 2024-02-21
feat: add opfs storage adapter

openscript:
This adds a minimal implementation of a [OPFS](https://developer.mozilla.org/en-US/docs/Web/API/File_System_API/Origin_private_file_system) storage adapter.

### DIFF
--- a/packages/automerge-repo-storage-opfs/LICENSE
+++ b/packages/automerge-repo-storage-opfs/LICENSE
@@ -1,0 +1,19 @@
+Copyright (c) 2024, openscript GmbH
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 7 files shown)


pvh:
Thanks, @openscript! I'll give it a proper review soon and let you know if I have any requests before merging. I don't suppose you've benched this against IndexedDB? I was thinking this morning I'm not sure which one to recommend to folks.

openscript:
No I've never really benched it, not in the context of automerge but also not outside of that. Just before I was looking for a benchmark, but I didn't really find a general and current browser storage benchmark. Maybe a project for a rainy weekend :)

pvh:
Hi! I'm catching up on PRs and spent a bit of time in here. I don't think it's ready to merge as-is. I'd like to make a few suggestions for improvement (which if time allows in the future I may take on myself.)

First: I suspect that passing in the root directory handle instead of sourcing it from the window directly would make this adapter compatible with the Chrome File System APIs as well -- this is quite a cool concept because it could allow sharing of a downloaded storage pool with processes running outside the browser. 

That said, if that were going to work you'd need to align your file location / key mapping system with the nodefs implementation. This is probably a good idea anyway, as the current implementation will suffer quite badly from the cost of scanning large filesystem directories if the user has a heap of data.

I'm going to mark this as a draft for now but I suspect we should be able to show that this would be a better default implementation for the project than the nodefs implementation. In fact, it might be easier to port that over... hmmm...

pvh:
Had a close look to consider landing this as part of the next release but I realized it's putting all the files into one flat directory. We have had reports that this approach leads to serious performance problems for large collections. In fact, it's the same reason that both git itself and the NodeFS implementation fragment the namespace by prefix. 

Fixing this wouldn't be too hard -- we can probably crib from the NodeFS implementation -- but I don't think it's something I can sneak in at this hour of my work day. Are you still using it, @openscript? Any experience reports after a few months?

PR #295, opened 2024-02-24
Awareness tests

RangerMauve:


msakrejda:
Thanks for the contribution, but I'll close this since we're deprecating these with #528. Please let us know if you have any feedback on that PR. Also, I'm still adding tests there so I'll try to incorporate the cases from these (the API is different but some of the setups that should be tested are very similar).

PR #301, opened 2024-02-29
fix error handler in BrowserWebSocketClientAdapter

HerbCaudill:
The websocket error handler signature is different on node and the browser

On node you get an `ErrorEvent`:

```ts
interface ErrorEvent {
  error: any
  message: string
  type: string
  target: WebSocket
}
```

In the browser you just get a generic `Event`:

```ts
interface Event {
  //...
  readonly srcElement: EventTarget | null
  readonly target: EventTarget | null
  readonly timeStamp: DOMHighResTimeStamp
  readonly type: string
  preventDefault(): void
  stopImmediatePropagation(): void
  stopPropagation(): void
  //...
}
```

The `isomorphic-ws` package only gives us the node types, and our error handler for
`BrowserWebSocketClientAdapter` naïvely trusts that those types are not filthy lies. As a result
when a ws connection fails in the browser, references to `event.error.code` throw additional errors
because `event.error` is undefined. No harm is done but it's unnecessary noise.

This PR changes the signature of the error handler so that it accepts the node event or the browser
event, and then sniffs properties to tell the difference.

### DIFF
--- a/packages/automerge-repo-network-websocket/src/BrowserWebSocketClientAdapter.ts
+++ b/packages/automerge-repo-network-websocket/src/BrowserWebSocketClientAdapter.ts
@@ -100,14 +100,25 @@ export class BrowserWebSocketClientAdapter extends WebSocketNetworkAdapter {
     this.receiveMessage(event.data as Uint8Array)
   }
 
-  onError = (event: WebSocket.ErrorEvent) => {
-    const { code } = event.error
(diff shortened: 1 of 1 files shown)


acurrieclark:
I grappled with this myself a few weeks ago, but didn't think to just invert the conditional to remove the error. Seems like a sensible option to me for now.

PR #296, opened 2024-02-26
useDocument listen to delete events

joshuahhh:
Currently useDocument doesn't listen for deletion, so docs stick around post-deletion.

(I'm not 100% sure about how deletion works, so plz lmk if this is off-base.)

PR #303, opened 2024-03-02
tests: start moving away from renderHook

joshuahhh:
I started writing tests for #296, and felt the hook-testing setup could be improved.

On `renderHook`, the [testing-library docs](https://testing-library.com/docs/react-testing-library/api#renderhook) say "You should prefer `render`", and I agree. `renderHook` is a strange wrapper around `render` – there's a component being rendered behind the scenes, but the interface doesn't look like a component. I'd rather have tests that show what the hooks do in the familiar setting of a component.

Relatedly, I feel uneasy with the current testing setup because they don't prove the hooks are causing the component to re-render when docs change. It might be that the hooks return mutable objects that are mutated by some listener somewhere, causing tests to pass, without triggering component renders (as is essential for reactivity).

The approach I like which addresses these concerns is to use a test component, which calls a mock function `onRender` on each render. This way, we get authoritative information about when the component renders and what it renders with.

I've implemented this approach for the `useDocument(s)` tests and I think it's a nice improvement. If you agree, I can follow this path for the rest of the tests.

Thanks!


pvh (commented):
These look good! Thanks for making them :)

PR #304, opened 2024-03-02
feat: add redis adapter

kid-icarus:
Adding this for review. I've tested this locally and it seems to work just fine. Ultimately I'd like to add some integration tests here, as it has none, similar to the nodefs storage adapter.

For that to work, I was considering leveraging a [test container](https://testcontainers.com/guides/getting-started-with-testcontainers-for-nodejs/) to ensure that Redis is installed, but that would require an update to our test github action (to install docker).

I meant to open this PR earlier but thought I'd get a conversation started :)

### DIFF
--- a/packages/automerge-repo-storage-redis/LICENSE
+++ b/packages/automerge-repo-storage-redis/LICENSE
@@ -1,0 +1,19 @@
+Copyright (c) 2024-Present, Ryan Kois
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
--- a/packages/automerge-repo-storage-redis/package.json
+++ b/packages/automerge-repo-storage-redis/package.json
@@ -1,0 +1,25 @@
+{
+  "name": "@automerge/automerge-repo-storage-redis",
+  "version": "1.0.0",
+  "description": "Redis storage adapter for Automerge Repo",
+  "repository": "https://github.com/automerge/automerge-repo/tree/master/packages/automerge-repo-storage-redis",
+  "author": "Ryan Kois <ryan.kois@gmail.com>",
+  "license": "MIT",
+  "type": "module",
+  "main": "dist/index.js",
+  "scripts": {
+    "build": "tsc",
+    "watch": "tsc -w"
+  },
+  "dependencies": {
+    "@automerge/automerge-repo": "workspace:*",
+    "debug": "^4.3.4",
+    "ioredis": "^5.3.2"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "devDependencies": {
+    "typescript": "^5.3.3"
+  }
+}
--- a/packages/automerge-repo-storage-redis/src/index.ts
+++ b/packages/automerge-repo-storage-redis/src/index.ts
@@ -1,0 +1,89 @@
+/**
+ * This module provides a storage adapter for Redis.
+ *
+ * @packageDocumentation
+ */
+
+import {
+  Chunk,
+  StorageAdapterInterface,
+  type StorageKey,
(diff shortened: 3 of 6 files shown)


acurrieclark:
This looks good to me, but the only reason we know the existing adapters work is because we are using them in production!

As you mention, there aren't any specific tests for any of the Storage Adapters, and it would in indeed be good to have a standard set for all of them.

Without being in a position to test it myself, I am loathe to merge just yet. I imagine this will be something people are looking for though.

pvh:
I think as the system matures we ought to start putting the storage & network adapters in their own repos (though then we lose automated test integration... hmm) but this is definitely something I've wanted to see for a while so thanks for sharing it!

packages/automerge-repo-storage-redis/LICENSE
```diff
@@ -0,0 +1,19 @@
+Copyright (c) 2024-Present, Ryan Kois
```
bijela-gora, line 1: Looks like in this file you missed the first line. Like [here](https://github.com/facebook/react/blob/64f354cf27a17c4e5b0dfcd908e47940cf947771/LICENSE#L1C1-L1C12).

packages/automerge-repo-storage-redis/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "type": "module",
+  "main": "dist/index.js",
+  "scripts": {
+    "build": "tsc",
+    "watch": "tsc -w"
+  },
+  "dependencies": {
+    "@automerge/automerge-repo": "workspace:*",
+    "debug": "^4.3.4",
+    "ioredis": "^5.3.2"
```
bijela-gora, line 17: It feels like the `ioredis` must be peer dependency. What do you think?

packages/automerge-repo-storage-redis/src/index.ts
```diff
@@ -0,0 +1,89 @@
...
+
+export class RedisStorageAdapter implements StorageAdapterInterface {
+  private redis: Redis
+
+  /**
+   * Create a new {@link RedisStorageAdapter}.
+   * @param opts - Options to pass to the Redis client.
+   */
+  constructor(
+    opts?: RedisOptions
```
bijela-gora, line 26: Maybe Redis instance must be passed as a constructor parameter?

packages/automerge-repo-storage-redis/src/index.ts
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  async remove(keyArray: string[]): Promise<void> {
+    await this.redis.del(keyArray.join(":"))
+  }
+
+  async loadRange(keyPrefix: string[]): Promise<Chunk[]> {
+    const lowerBound = [...keyPrefix, '*'].join(":")
+    const result: Chunk[] = []
+    const stream = this.redis.scanStream({
```
bijela-gora, line 52: Nice!

bijela-gora (commented):
Hi,

I decided to check the list of PRs before trying to code a Redis-based implementation of `StorageAdapterInterface`. And here it is. Thank you for your work! Don't you mind I will review it?

bijela-gora:
@pvh hello,

Your phrase: "though then we lose automated test integration... hmm" reminded me the [Promises/A+ Compliance Test Suite](https://github.com/promises-aplus/promises-tests). What do you think of the idea of developing a set of tests that could be used to test any implementation of the `StorageAdapterInterface`?

pvh:
Sounds great if it can be kept simple. I think the tricky part here is operational complexity -- for things like nodefs & indexeddb there are easily available implementations but for Redis, Postgres, or other backends, we'll need to have the ability to spin up the service at runtime. For SaaS-only backends (say, S3) there's yet more complexity.

That said, I'm quite interested in a solution. I think @HerbCaudill was the one who designed the system we use to test the network adapters and as we grow beyond a monorepo, I think something along the lines you propose could be valuable.

packages/automerge-repo-storage-redis/package.json
kid-icarus: Yeah that makes sense, I agree

bijela-gora:
@kid-i

pvh:
@kid-icarus How do you feel about hosting this as a standalone package at /automerge/automerge-repo-storage-redis/?

I think my intuition is that adapters which rely on things not found in most browsers / OS should have their own repo and then whatever CI we set up can be independent.

bijela-gora:
@pvh

I'd like to share that I created an SQLite storage adapter for JS/TS projects: https://github.com/bijela-gora/automerge-repo-storage-better-sqlite3

I created it because I'm going to use it in my pet project.

Thank you for the awesome Automerge project!

kid-icarus:
> @kid-icarus How do you feel about hosting this as a standalone package at /automerge/automerge-repo-storage-redis/?
> 
> I think my intuition is that adapters which rely on things not found in most browsers / OS should have their own repo and then whatever CI we set up can be independent.

That makes sense from a principled perspective 👍🏻

One pro of a separate repo is more control over CI, being able to throw whatever external dependencies into a Dockerfile, etc. Another pro is that if there are breaking changes in automerge(-repo), namely the types in this specific case, the burden isn't placed on the core team to update the adapter in the monorepo.

One con would be extra care to specify automerge and automerge-repo as a combination of dev and peer dependencies, so that adapter consumers aren't using duplicate, potentially incompatible versions of them. This also means that as newer versions of automerge(-repo) are released, we'll need to bump those dev dependencies. Dependabot can help with that, but it's just the general con of something living in a monorepo vs an external repo.

akellbl4:
Not sure how far `ioredis` from `node-redis`, but here is [a part of the description](https://github.com/redis/ioredis#readme) from `ioredis` package.

> For new projects, node-redis is the recommended client library. [node-redis](https://github.com/redis/node-redis) is the open-source (MIT license) Redis JavaScript client library redesigned from the ground up and actively maintained. [node-redis](https://github.com/redis/node-redis) supports new (hash-field expiration) and future commands and the capabilities available in Redis Stack and Redis 8 (search, JSON, time-series, probabilistic data structures).

I guess if API lines up well enough storage can allow passing an instance of both `ioredis` and `node-redis`

PR #305, opened 2024-03-02
chore: update storage adapters to use interface

kid-icarus:
I noticed some deprecation warnings as I was doing some other work, so swapped out `extends StorageAdapter` to `implements StorageAdapterInterface` per the deprecation notes.


acurrieclark:
Thank you for spotting this. The Network adapters were updated when the interfaces were introduced, but I completely forgot about the Storage Adapters.

PR #306, opened 2024-03-07
adding documentation for the leave message

heckj:
- determined by reading the WebSocket source material

### DIFF
--- a/packages/automerge-repo-network-websocket/README.md
+++ b/packages/automerge-repo-network-websocket/README.md
@@ -140,6 +140,17 @@ Sent by the receiving peer in response to the join message in the
 }
 ```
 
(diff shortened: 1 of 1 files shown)


alexjg:
Whilst in practice this is sent by the initiating peer I don't think there's any constraint that it _has_ to be the initiating peer, we should probably just say "sent by a peer when they are planning to disconnect". I'm actually not sure what the receiver should do with this message, it seems to me like it should probably be an error if you receive a message after receiving the `leave` message, but in practice it's fine to just ignore it. You certainly can't rely on it for disconnect logic because the connection may be terminated abruptly. Maybe it's best to just say the `leave` message is advisory and has no effect on connection handling?

PR #307, opened 2024-03-08
Tests suite for storage adapters

bijela-gora:
This test suite is based on the @acurrieclark [comment](https://github.com/automerge/automerge-repo/pull/304#issuecomment-1977446032):
> the only reason we know the existing adapters work is because we are using them in production!

To get production data, I launched the React to-do app from the example directory. I also launched the Node.js server and recorded calls to each public method of the `NodeFSStorageAdapter` instance. That recordings found its place in the test suite. I made buffers smaller for simplicity.

The reason why the storage adapter test suite function has such an interface and is used in this specific way is because I think it should be possible to execute suite hooks. For example, I think it's a good idea to clean up temp files after the execution of each test. The suit hooks are a good place for this, because they will be called even if a test fails. But it seems that for testing the NodeFSStoradgeAdapter, calling the suite hook is not necessary.

In addition to all above I have added comments to explain the decision behind and reasons for a change.

### DIFF
--- a/packages/automerge-repo-storage-nodefs/test/NodeFSStorageAdapter.test.ts
+++ b/packages/automerge-repo-storage-nodefs/test/NodeFSStorageAdapter.test.ts
@@ -13,45 +13,26 @@ function cleanDir(dir: string) {
   } catch (e) {}
 }
 
-describe('NodeFSStorageAdapter', () => {
-  let baseDirectory: string;
-  let adapter: NodeFSStorageAdapter;
-
-  beforeEach(async () => {
-    baseDirectory = path.join(os.tmpdir(), crypto.randomUUID())
-    adapter = new NodeFSStorageAdapter(baseDirectory)
-  })
-
-  afterAll(async() => {
-    cleanDir(baseDirectory);
-  })
-
-  describe('getFilePath', () => {
-    it('should compose keys correctly', () => {
-      // @ts-ignore
-      const actual = adapter.getFilePath(["3xuJ5sVKdBaYS6uGgGJH1cGhBLiC","sync-state","d99d4820-fb1f-4f3a-a40f-d5997b2012cf"])
-      expect(actual).toStrictEqual(path.join(baseDirectory) + '/3x/uJ5sVKdBaYS6uGgGJH1cGhBLiC/sync-state/d99d4820-fb1f-4f3a-a40f-d5997b2012cf')
-    })
-  })
+function runStorageTests(sut: {adapter: NodeFSStorageAdapter}) {
 
   describe('load', () => {
     it('should return undefined if there is no data', async () => {
       expect(
-        await adapter.load(["3xuJ5sVKdBaYS6uGgGJH1cGhBLiC","sync-state","d99d4820-fb1f-4f3a-a40f-d5997b2012cf"])
+        await sut.adapter.load(["3xuJ5sVKdBaYS6uGgGJH1cGhBLiC","sync-state","d99d4820-fb1f-4f3a-a40f-d5997b2012cf"])
       ).toStrictEqual(undefined)
     })
   })
 
   describe('save and load', () => {
     it('should be possible to save and load', async () => {
-      await adapter.save(["storage-adapter-id"], new TextEncoder().encode('8a35c9b4-109e-4a7f-a35e-a5464121b6dd'));
-      const actual = await adapter.load(["storage-adapter-id"]);
+      await sut.adapter.save(["storage-adapter-id"], new TextEncoder().encode('8a35c9b4-109e-4a7f-a35e-a5464121b6dd'));
+      const actual = await sut.adapter.load(["storage-adapter-id"]);
       expect(actual).toStrictEqual(new TextEncoder().encode('8a35c9b4-109e-4a7f-a35e-a5464121b6dd'))
     })
 
     it('should work with composed keys', async () => {
-      await adapter.save(["pSq9fP9ekr1zembLzBJkgHTo7Wn","sync-state","3761c9f0-bb1d-44b6-88ac-f85072fc3273"], new Uint8Array([0, 1, 127, 99, 154, 235 ]))
-      const actual = await adapter.load(["pSq9fP9ekr1zembLzBJkgHTo7Wn","sync-state","3761c9f0-bb1d-44b6-88ac-f85072fc3273"]);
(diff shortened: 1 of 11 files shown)


bijela-gora:
I added some tests based on the issues I noticed while working on the LevelDbStoradgeAdapter in the [side branch](https://github.com/bijela-gora/automerge-repo/commit/0f8ef4efef1137e37c5e4d667e542e060c3499be#diff-e31e1647cb96497d61fa209a6250a7628722e8c4c6a1e8875ca2f9d8c8a29389R20).

bijela-gora:
While writing the SQLite storadge adapter. I encountered an issue that forces me to ask a question again. Is [this](https://github.com/automerge/automerge-repo/pull/307/files#diff-12e92867ad3c0285a974a0e32aa7379b453f211095e195d2f08a6edd1899ed1eR100) test which I added based on the NodeFSStorageAdapter behavior correct? Is this behavior expected?

bijela-gora:
@HerbCaudill in [this comment](https://github.com/automerge/automerge-repo/pull/304#issuecomment-1983968943) you wrote:
> i think it would be a good idea to restructure the storage adapter tests that currently live in [StorageSubsystem.test.ts](https://github.com/automerge/automerge-repo/blob/main/packages/automerge-repo/test/StorageSubsystem.test.ts) so that they work more like the network adapter acceptance tests [here](https://github.com/automerge/automerge-repo/blob/main/packages/automerge-repo/src/helpers/tests/network-adapter-tests.ts)

I can say that I did that in this PR. Could I ask you to review it?

bijela-gora:
Just a kind reminder: @HerbCaudill could you please review this PR?

HerbCaudill:
> While writing the SQLite storadge adapter. I encountered an issue that forces me to ask a question again. Is [this](https://github.com/automerge/automerge-repo/pull/307/files#diff-12e92867ad3c0285a974a0e32aa7379b453f211095e195d2f08a6edd1899ed1eR100) test which I added based on the NodeFSStorageAdapter behavior correct? Is this behavior expected?

I'm pretty sure that's what I would expect - @pvh @alexjg @alexcc thoughts?

HerbCaudill:
Hi, @bijela-gora - thanks for working on this, looks good and does what it needs to do. I'd like to organize things a little differently, can you [allow me to push commits](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to this PR branch?

HerbCaudill:
I also rebased on latest `main` branch so it can be merged. 

Let me know if this all looks good to you, and I'll merge this PR.

packages/automerge-repo-storage-nodefs/vitest.config.ts
```diff
@@ -1,8 +1,11 @@
 import { defineConfig, mergeConfig } from "vitest/config"
-import rootConfig from '../../vitest.config'
+import rootConfig from "../../vitest.config"
```
bijela-gora, line 2: Sorry, for missing styling fixes. Looks like I forgot to configure Prettier correctly.

packages/automerge-repo/src/helpers/tests/storage-adapter-tests.ts
```diff
@@ -28,185 +32,68 @@ export function runStorageAdapterTests(_setup: SetupFn, title?: string): void {
...
         const actual = await adapter.load(["storage-adapter-id"])
 
-        expect(actual).toStrictEqual(
-          new Uint8Array([
-            56, 97, 51, 53, 99, 57, 98, 52, 45, 49, 48, 57, 101, 45, 52, 97, 55,
-            102, 45, 97, 51, 53, 101, 45, 97, 53, 52, 54, 52, 49, 50, 49, 98,
-            54, 100, 100,
-          ])
-        )
+        expect(actual).toStrictEqual(PAYLOAD_A)
```
bijela-gora, line 37: A nitpick.

In most cases, it is not recommended to share any data between the Act and Assert steps. This can lead to errors. Because any change to an input object in the Act step changes the expectations object. There are couple ways to prevent the issue: 1) hardcoding an input and expectation values; 2) using a factory function like `PAYLOAD_A()`.

packages/automerge-repo/src/helpers/tests/storage-adapter-tests.ts
```diff
@@ -2,336 +2,376 @@ import { describe, expect, it } from "vitest"
...
+      it("should return undefined if there is no data", async () => {
+        const { adapter, teardown } = await setup()
+        expect(
+          await adapter.load([
+            "3xuJ5sVKdBaYS6uGgGJH1cGhBLiC",
+            "sync-state",
+            "d99d4820-fb1f-4f3a-a40f-d5997b2012cf",
+          ])
+        ).toStrictEqual(undefined)
+        teardown()
```
bijela-gora, line 27: The reason I intentionally utilized the `afterEach` hook is because it will be called even if a test fails. The `expect` function `throw`s an error. Maybe it is good to use [`onTestFinished`](https://vitest.dev/api/#ontestfinished) like this:
```javascript
const { adapter, teardown } = await setup()
onTestFinished(() => teardown())
```

packages/automerge-repo-storage-nodefs/test/NodeFSStorageAdapter.test.ts
```diff
@@ -1,38 +1,19 @@
...
-describe('NodeFSStorageAdapter', () => {
-  let baseDirectory: string = tempDir()
-  let sut: {adapter: NodeFSStorageAdapter} = {adapter: new NodeFSStorageAdapter(baseDirectory)}
-
-  beforeEach(async () => {
-    baseDirectory = tempDir()
-    sut.adapter = new NodeFSStorageAdapter(baseDirectory)
-  })
-
-  afterAll(async() => {
```
bijela-gora: It is mistake to use `afterAll` here. Should be `afterEach`.

packages/automerge-repo/src/helpers/tests/storage-adapter-tests.ts
bijela-gora: Ok, currently installed version of vitest is `1.2.2` and it doesn't support the onTestFinished.

bijela-gora:
@HerbCaudill let's merge?

packages/automerge-repo/src/helpers/tests/storage-adapter-tests.ts
HerbCaudill: That's a good point, thanks! I've changed PAYLOAD_A etc to functions.

PR #312, opened 2024-03-16
Fix event description in README.md

msakrejda:


packages/automerge-repo/README.md
```diff
@@ -52,7 +52,7 @@ A `Repo` exposes these methods:
 - `.on("document", ({handle: DocHandle}) => void)`  
   Registers a callback to be fired each time a new document is loaded or created.
 - `.on("delete-document", ({handle: DocHandle}) => void)`  
-  Registers a callback to be fired each time a new document is loaded or created.
+  Registers a callback to be fired each time a new document is deleted.
```
msakrejda, line 55: It might also be nice to clarify whether this is for deletions on this instance only or on peers or both.

acurrieclark: At the moment deletions are not triggered between peers and are only performed locally. I think this will suffice for now, but I admit we should revisit when we have a network `delete` message

acurrieclark:
Thank you, a good spot.

PR #313, opened 2024-03-17
Add @automerge/create-vite-app

alexjg:
When getting people up and running with automerge it's painful to go through the whole ceremony of setting up a vite app, installing automerge dependencies, and setting up the react hooks just to demonstrate simple functionality. This patch adds an `@automerge/create-vite-app` which allows you to run `npm create @automerge/vite-app` to quickly spin up a Vite + React + Typescript project with a simple counter app inside it.

Most of this is just template code in `packages/create-vite-app/template`. The code which actually sets up the new project is in `packages/create-vite-app/index.ts`.

If you want to have a play with this the easiest way I've found to test things is to run `npm pack` in the `create-vite-app` package, then `npm install <the tarball we just created>` in a new project and finally `./node_modules/@automerge/create-vite-app/dist/index.js` to actually run the helper.


packages/create-vite-app/template/vite.config.ts
```diff
@@ -0,0 +1,9 @@
+import { defineConfig } from 'vite'
+import react from '@vitejs/plugin-react'
+import wasm from "vite-plugin-wasm"
+import topLevelAwait from "vite-plugin-top-level-await"
```
pvh, line 4: I don't believe we need top-level-await anymore. Can we cut this?

packages/create-vite-app/template/src/main.tsx
```diff
@@ -0,0 +1,34 @@
+import { isValidAutomergeUrl, Repo } from '@automerge/automerge-repo'
+import { BrowserWebSocketClientAdapter } from '@automerge/automerge-repo-network-websocket'
+import { IndexedDBStorageAdapter } from "@automerge/automerge-repo-storage-indexeddb"
+import { RepoContext } from '@automerge/automerge-repo-react-hooks'
+import {next as A} from "@automerge/automerge" //why `next`? See the the "next" section of the conceptual overview
```
pvh, line 5: this is misformatted; probably you should make sure prettier has run over the lot of it. also, the comment is not very helpful: we should link directly to the docs.

packages/create-vite-app/template/src/main.tsx
```diff
@@ -0,0 +1,34 @@
...
+import React from 'react'
+import ReactDOM from 'react-dom/client'
+import App from './App.tsx'
+import './index.css'
+
+const repo = new Repo({
+  network: [new BrowserWebSocketClientAdapter("ws://sync.automerge.org")],
```
pvh, line 12: needs to be `wss`.

packages/create-vite-app/template/src/main.tsx
```diff
@@ -0,0 +1,34 @@
...
+  storage: new IndexedDBStorageAdapter(),
+})
+
+const rootDocUrl = `${document.location.hash.substr(1)}`
```
pvh, line 16: substr is deprecated in favour of substring (i just merged a PR that fixes this upstream)

packages/create-vite-app/template/src/main.tsx
```diff
@@ -0,0 +1,34 @@
...
+let handle
+if (isValidAutomergeUrl(rootDocUrl)) {
+    handle = repo.find(rootDocUrl)
+} else {
+    handle = repo.create<{counter?: A.Counter}>()
+    handle.change(d => d.counter = new A.Counter())
+}
+const docUrl = document.location.hash = handle.url
+// @ts-ignore
```
pvh, line 25: this doesn't throw lint complaints:
```
// @ts-expect-error -- we put the handle and the repo on window so you can experiment with them from the dev tools
```
also, we should include the repo on the window.

packages/create-vite-app/template/src/main.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { RepoContext } from '@automerge/automerge-repo-react-hooks'
+import {next as A} from "@automerge/automerge" //why `next`? See the the "next" section of the conceptual overview
+import React from 'react'
+import ReactDOM from 'react-dom/client'
+import App from './App.tsx'
+import './index.css'
+
+const repo = new Repo({
+  network: [new BrowserWebSocketClientAdapter("ws://sync.automerge.org")],
+  storage: new IndexedDBStorageAdapter(),
```
pvh, line 13: related to the above i was surprised when I didn't get local synchronization without sync server access; should we include BroadcastChannel by default? it's slower than MessageChannel but probably plenty good enough for experimentation and extremely convenient.

pvh, line 13: this line should also pass a database name in to avoid collisions

packages/create-vite-app/template/src/App.tsx
```diff
@@ -0,0 +1,41 @@
...
+          <img src={viteLogo} className="logo" alt="Vite logo" />
+        </a>
+        <a href="https://react.dev" target="_blank">
+          <img src={reactLogo} className="logo react" alt="React logo" />
+        </a>
+      </div>
+      <h1>Vite + React</h1>
+      <div className="card">
+        <button onClick={() => changeDoc((d) => d.counter.increment(1))}>
+          count is { doc && doc.counter.value }
```
pvh, line 28: I wonder if there's anything sufficiently small we could do that's more interesting/impressive than a counter...

packages/create-vite-app/template/README.md
```diff
@@ -0,0 +1,3 @@
+# Vite + React + Automerge
+
+This template provides a minimal vite app with React and Automerge
```
msakrejda, line 3:
```suggestion
This template provides a minimal Vite app with React and Automerge
```

msakrejda (commented):
Minor copy comment, but looks great. Thanks for working on this.

PR #318, opened 2024-03-19
Prevent returning undefined once for DocHandles that are ready

pvh:
Reported (and coauthored) by @msakrejda! This should reduce waterfall loading and save us a full frame of wasted rendering for docs that are already loaded.

### DIFF
--- a/packages/automerge-repo-react-hooks/src/useDocument.ts
+++ b/packages/automerge-repo-react-hooks/src/useDocument.ts
@@ -16,17 +16,20 @@ import { useRepo } from "./useRepo.js"
  * This requires a {@link RepoContext} to be provided by a parent component.
  * */
 export function useDocument<T>(id?: AnyDocumentId) {
-  const [doc, setDoc] = useState<Doc<T>>()
   const repo = useRepo()
 
   const handle = id ? repo.find<T>(id) : null
   const handleRef = useRef<DocHandle<T> | null>(null)
 
+  const [doc, setDoc] = useState<Doc<T> | undefined>(handle?.docSync())
+
   useEffect(() => {
-    // When the handle has changed, reset the doc to an empty state.
+    // When the handle has changed, reset the doc to the current value of docSync().
+    // For already-loaded documents this will be the last known value, for unloaded documents
+    // this will be undefined.
     // This ensures that if loading the doc takes a long time, the UI
     // shows a loading state during that time rather than a stale doc.
-    setDoc(undefined)
+    setDoc(handle?.docSync())
 
     if (!handle) return
 
--- a/packages/automerge-repo-react-hooks/test/useDocument.test.tsx
+++ b/packages/automerge-repo-react-hooks/test/useDocument.test.tsx
@@ -29,14 +29,21 @@ describe("useDocument", () => {
     const handleSlow = repo.create<ExampleDoc>()
     handleSlow.change(doc => (doc.foo = "slow"))
     const oldDoc = handleSlow.doc.bind(handleSlow)
+    const syncDoc = undefined
     handleSlow.doc = async () => {
       await new Promise(resolve => setTimeout(resolve, SLOW_DOC_LOAD_TIME_MS))
       const result = await oldDoc()
+      const syncDoc = result
       return result
+    }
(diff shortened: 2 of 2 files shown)


packages/automerge-repo-react-hooks/test/useDocument.test.tsx
```diff
@@ -61,15 +71,25 @@ describe("useDocument", () => {
     const { handleA, wrapper } = setup()
     const onDoc = vi.fn()
 
-    render(<Component url={handleA.url} onDoc={onDoc} />, {wrapper})
+    render(<Component url={handleA.url} onDoc={onDoc} />, { wrapper })
     await waitFor(() => expect(onDoc).toHaveBeenLastCalledWith({ foo: "A" }))
   })
 
+  it.only("should immediately return a document if it has already been loaded", async () => {
```
msakrejda, line 78: Tests are failing because of the `.only` here:
```suggestion
  it("should immediately return a document if it has already been loaded", async () => {
```

packages/automerge-repo-react-hooks/src/useDocument.ts
```diff
@@ -16,17 +16,20 @@ import { useRepo } from "./useRepo.js"
  * This requires a {@link RepoContext} to be provided by a parent component.
  * */
 export function useDocument<T>(id?: AnyDocumentId) {
-  const [doc, setDoc] = useState<Doc<T>>()
   const repo = useRepo()
 
   const handle = id ? repo.find<T>(id) : null
   const handleRef = useRef<DocHandle<T> | null>(null)
 
+  const [doc, setDoc] = useState<Doc<T> | undefined>(handle?.docSync())
```
msakrejda, line 24: I'm assuming that `docSync()` is pretty cheap, but if it's not, we can wrap this in a function to only call it when the component is first mounted:

```suggestion
  const [doc, setDoc] = useState<Doc<T> | undefined>(() => handle?.docSync())
```

packages/automerge-repo-react-hooks/test/useDocument.test.tsx
```diff
@@ -29,14 +29,21 @@ describe("useDocument", () => {
...
+    const syncDoc = undefined
     handleSlow.doc = async () => {
       await new Promise(resolve => setTimeout(resolve, SLOW_DOC_LOAD_TIME_MS))
       const result = await oldDoc()
+      const syncDoc = result
       return result
     }
+    handleSlow.docSync = () => {
+      return syncDoc
+    }
```
msakrejda, line 41: Just for readability, I think this would be clearer as
```suggestion
    const oldDoc = handleSlow.docSync.bind(handleSlow)
    const result = oldDoc()
    handleSlow.doc = async () => {
      await new Promise(resolve => setTimeout(resolve, SLOW_DOC_LOAD_TIME_MS))
      return result
    }
    handleSlow.docSync = () => {
      return result
    }
```
(I _think_ that has the same semantics.)

msakrejda (commented):
:+1:

pvh:
Thanks for the fixes @msakrejda

msakrejda:
Well, apparently they weren't really fixes :sweat_smile:. Taking a look.

msakrejda:
Ah, I think these may be other tests that were also failing when a whole bunch of other stuff was failing and then we added the `.only` and didn't notice they were still failing. I think I can fix them.

msakrejda:
Okay, see https://github.com/automerge/automerge-repo/compare/usedocs-avoid-undefined...msakrejda:automerge-repo:usedocs-avoid-undefined -- this passes the full suite locally and fails our new test with the old code. Our change to handleSlow broke some assumptions, but we can get away with a much simpler change.

geoffreylitt:
LGTM, good catch!

geoffreylitt (approved):

packages/automerge-repo-react-hooks/src/useDocument.ts
HerbCaudill: `docSync()` should be cheap but in a couple of experiments it gets called a lot of extra times the first way, so we might as well wrap in a function

PR #321, opened 2024-03-20
update react-use-awareness example

HerbCaudill:
- package.json was missing a dependency
- ran prettier throughout


alexjg (approved):

alexjg:
LGTM but could you tell me how you are running prettier? I tried running it yesterday and it changed almost every file.

HerbCaudill:
> LGTM but could you tell me how you are running prettier? I tried running it yesterday and it changed almost every file.

@alexjg I just opened those specific files and hit save in VS Code, it formats on save. 

It probably wouldn't be a bad idea to do a one-time pass on the whole monorepo with `prettier . -write`, as a standalone commit/PR.

PR #323, opened 2024-03-20
Run repo through prettier

HerbCaudill:
Adds a .prettierignore file so that we don't format automatically-generated files, then runs `prettier . -w` to format all files in the monorepo.


pvh (commented):
Looks good -- should we consider adding it to the automated linter to prevent future regression?

PR #325, opened 2024-03-21
Add prettier check to lint script

HerbCaudill:
This PR adds `prettier . --check` to the `lint` script at the monorepo root. This will ensure that code that's not consistently formatted will get flagged by the [lint action](https://github.com/automerge/automerge-repo/actions/workflows/lint.yaml), so PRs with wonky formatting will get the red X instead of the green check etc.


package.json
```diff
@@ -43,8 +44,8 @@
     "@types/react-dom": "^18.2.18",
     "@types/uuid": "^8.3.4",
     "@types/ws": "^8.5.10",
-    "@typescript-eslint/eslint-plugin": "^6.21.0",
-    "@typescript-eslint/parser": "^6.21.0",
+    "@typescript-eslint/eslint-plugin": "^7.3.1",
+    "@typescript-eslint/parser": "^7.3.1",
     "@vitejs/plugin-react": "^3.1.0",
```
HerbCaudill, line 49: I bumped this because we were getting a warning from typescript-eslint that it was only guaranteed to work with older versions of typescript.

alexjg (approved):

package.json
```diff
@@ -17,7 +17,8 @@
     "start:syncserver": "cross-env DEBUG='WebsocketServer' pnpm -F @automerge/example-sync-server start",
     "preinstall": "npx only-allow pnpm",
     "repocheck": "manypkg check",
-    "lint": "eslint --ext .ts .",
+    "format": "prettier --write . | dev-null",
```
HerbCaudill, line 20: `prettier --write` outputs to the console the name of every single file it looks at, whether it modifies it or not. That's a lot of noise, so we pipe that output to `dev/null` (or `NUL` on windows). 

Errors still are reported to the console.

PR #331, opened 2024-04-04
[2.0 release chore] remove useBootstrap

pvh:
`useBootstrap` was a good idea (maybe), poorly executed (by me).

Instead of this confusing and complicated bit of obfuscation that's really only fit for demo code, we now use a much shorter and more transparent stanza in examples:

```javascript
const rootDocUrl = `${document.location.hash.substring(1)}`
let handle
if (isValidAutomergeUrl(rootDocUrl)) {
  handle = repo.find(rootDocUrl)
} else {
  handle = repo.create<State>({ todos: [] })
}
const docUrl = (document.location.hash = handle.url)
```

Out of respect for semver, I've marked it as deprecated for now and made this PR to remember to finish it off later.


pvh:
@alexjg since we're changing the NetworkAdapter API and doing a major we should merge this

pvh:
Merged as part of 2.0

PR #333, opened 2024-04-05
fix test: "throws an error if we try to import an invalid document"

HerbCaudill:
The VS Code test runner was choking on this test - not sure why, because it's syntactically valid, but upon inspection it did seem like it wasn't intended to be written the way it was. I've fixed it in a way that makes sense to me and the test runner is happy. 

before: 

```ts
expect(() => {
  repo.import<TestDoc>(A.init<TestDoc> as unknown as Uint8Array)
}).toThrow()
````

after: 

```ts
expect(() => {
  repo.import<TestDoc>(new Uint8Array([1, 2, 3]))
}).toThrow()
```

### DIFF
--- a/packages/automerge-repo/test/Repo.test.ts
+++ b/packages/automerge-repo/test/Repo.test.ts
@@ -451,11 +451,29 @@ describe("Repo", () => {
       expect(A.getHistory(v)).toEqual(A.getHistory(updatedDoc))
     })
 
-    it("throws an error if we try to import an invalid document", async () => {
+    it("throws an error if we try to import a nonsensical byte array", async () => {
       const { repo } = setup()
       expect(() => {
         repo.import<TestDoc>(new Uint8Array([1, 2, 3]))
       }).toThrow()
+    })
+
+    // TODO: not sure if this is the desired behavior from `import`.
+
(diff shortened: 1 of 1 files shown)


acurrieclark:
From memory, the test was supposed to check that if a document (rather than a Uint8Array) was passed in, it would throw an error.

Admittedly, passing a document will show a type error, but that was the original plan.

HerbCaudill:
> From memory, the test was supposed to check that if a document (rather than a Uint8Array) was passed in, it would throw an error.
> 
> Admittedly, passing a document will show a type error, but that was the original plan.

ah ok. I'll update it.

HerbCaudill:
hm that actually doesn't throw...

image

HerbCaudill:
This is the actual behavior: 

image

HerbCaudill:
I've committed these tests, documenting the current behavior of `Repo.import({ foo: 123 })`and `Repo.import(A.from({ foo: 123 }))`. I'd suggest merging this fix and then deciding whether that's the behavior we want. (My 2 cents is that it's not a problem that it doesn't throw when you give it a parameter of the wrong type. We shouldn't be expected to do input validation in every function in the codebase, that's what TypeScript is for.)

PR #336, opened 2024-04-09
A simple heads API & optional Repo network config

pvh:
Was working on some related stuff and noticed we didn't have an API for .heads() so I added it. Didn't add a promise form... didn't feel necessary in the moment but I could do so if anyone thinks it makes sense. Tests included.

I also fixed a small annoyance en passant which I noticed while I was in there (separate patch but same PR). You can now instantiate a Repo with no config at all! No more passing in `network: []` to prevent complaints in tests.

I wanted to do a bit more about the syncState API (which is why I was in here in the first place) but I think there's a bit more of a conversation / design task to be done first so I'm going to file an issue about it and discuss there.

### DIFF
--- a/packages/automerge-repo/src/DocHandle.ts
+++ b/packages/automerge-repo/src/DocHandle.ts
@@ -328,6 +328,19 @@ export class DocHandle<T> //
     return this.#doc
   }
 
+  /**
+   * Returns the current "heads" of the document, akin to a git commit.
+   * This precisely defines the state of a document. In the event of multiple heads,
+   * there are unmerged concurrent changes to the document.
+   * @returns the current document's heads, or undefined if the document is not ready
+   */
+  heads(): A.Heads | undefined {
+    if (!this.isReady()) {
+      return undefined
+    }
+    return A.getHeads(this.#doc)
+  }
+
   /** `update` is called by the repo when we receive changes from the network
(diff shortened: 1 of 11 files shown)


packages/automerge-repo/src/DocHandle.ts
```diff
@@ -328,6 +328,19 @@ export class DocHandle<T> //
     return this.#doc
   }
 
+  /**
+   * Returns the current "heads" of the document, akin to a git commit.
+   * This precisely defines the state of a document. In the event of multiple heads,
+   * there are unmerged concurrent changes to the document.
```
HerbCaudill, line 334: This suggests that the document state is unresolved if there are multiple heads. My impression was that Automerge has no "merge commits", so having multiple heads just means that there were concurrent changes at some point, and there haven't been any changes since then; but the concurrent changes are all materialized in the document, so there are no differences between 1 head and 2+ heads that we need to care about. Is that understanding correct?

HerbCaudill (approved):

packages/automerge-repo/src/DocHandle.ts
pvh: Yes. I guess the word "unmerged" is misleading here. I'll  just remove it: it's accurate without.

PR #341, opened 2024-04-18
Fix: Fix race condition in `Repo.flush` method

mykola-vrmchk:
### Motivation
The `Flush` method waits for specific heads to be written on the disk, there was a race condition when you do change exactly after the `flush` heads on the disk would never be equal to the flushing heads.
This snippet shows what exactly was happening in the storage subsystem (and before this PR it would hang forever):
```ts
      it.only("Flush should resolve even if doc is getting changed afterwards", async () => {
        const storage = new StorageSubsystem(new NodeFSStorageAdapter())

        const doc = A.change(A.init<any>(), "first change", d => {
          d.foo = "bar"
        })
       
        // start flushing (with current heads).
        const key = parseAutomergeUrl(generateAutomergeUrl()).documentId
        const flush = storage.flush(key, doc)

        // make a change (changes heads).
        const changedDoc = A.change<any>(doc, "second change", d => {
          d.foo = "baz"
        })
        await storage.saveDoc(key, changedDoc)

        await flush // <- hangs here forever.
      }) 
 ```
 
 ### Fix
 Invoke save on flush method, because of how automerge storage works it will not break it, it could just make additional useless write, which is already pending, or save mutations which are already saved.

### DIFF
--- a/packages/automerge-repo/src/Repo.ts
+++ b/packages/automerge-repo/src/Repo.ts
@@ -507,10 +507,9 @@ export class Repo extends EventEmitter<RepoEvents> {
   }
 
   /**
-   * Waits for Repo to finish write changes to disk.
-   * @hidden this API is experimental and may change
-   * @param documents - if provided, only waits for the specified documents
-   * @param timeout - if provided, the maximum time to wait in milliseconds (rejects on timeout)
+   * Writes Documents to a disk.
+   * @hidden this API is experimental and may change.
+   * @param documents - if provided, only writes the specified documents.
    * @returns Promise<void>
    */
   async flush(documents?: DocumentId[]): Promise<void> {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: packages/automerge-repo/src/Repo.ts [hunk 1]
```diff
@@ -513,24 +513,22 @@ export class Repo extends EventEmitter<RepoEvents> {
    * @param timeout - if provided, the maximum time to wait in milliseconds (rejects on timeout)
    * @returns Promise<void>
    */
-  async flush(documents?: DocumentId[], timeout?: number): Promise<void> {
+  async flush(documents?: DocumentId[]): Promise<void> {
     if (!this.storageSubsystem) {
       return Promise.resolve()
```
dmaretskyi, line 518:
```suggestion
      return
```

packages/automerge-repo/src/Repo.ts
```diff
@@ -513,24 +513,22 @@ export class Repo extends EventEmitter<RepoEvents> {
    * @param timeout - if provided, the maximum time to wait in milliseconds (rejects on timeout)
```
dmaretskyi, line 513:
```suggestion
```

PR #350, opened 2024-05-08
Add syncPolicy callback for per-document authorization

max-dekock:
This PR adds a `syncPolicy` callback to the repo configuration. It will be called whenever a sync message is received from a peer, and if it returns false, syncing will be blocked, and the repo will respond with a `doc-unavailable` message.

Currently, if `sharePolicy` returns true, the repo will still send sync messages to peers even if `syncPolicy` would return false. If this is undesirable, another possibility is to only share if `syncPolicy` and `sharePolicy` both return true.

### DIFF
--- a/packages/automerge-repo/src/synchronizer/CollectionSynchronizer.ts
+++ b/packages/automerge-repo/src/synchronizer/CollectionSynchronizer.ts
@@ -85,6 +85,18 @@ export class CollectionSynchronizer extends Synchronizer {
     )
 
     const documentId = message.documentId
+    const senderId = message.senderId
+    const okToSync = await this.repo.syncPolicy(senderId, documentId)
+    if (!okToSync) {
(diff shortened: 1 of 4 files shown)


packages/automerge-repo/src/synchronizer/CollectionSynchronizer.ts
```diff
@@ -85,6 +85,18 @@ export class CollectionSynchronizer extends Synchronizer {
     )
 
     const documentId = message.documentId
+    const senderId = message.senderId
+    const okToSync = await this.repo.syncPolicy(senderId, documentId)
```
twop, line 89: was casually observing Discord and took a look at the PR, pardon my intrusion, just wanted to mention that `await somePromise()` will always put a new event into the event loop **at the very end**, thus even an innocent no-op (like `async () => true`) can create a lot of latency if the event loop is busy/saturated. So it is possible that it might be better to make it optional and thus not `await`ing as the default behavior.

PR #352, opened 2024-05-17
expose isValidDocumentId

akellbl4:
It's useful to have`isValidDocumentId` exposed so if we can check validity of an `id` when it comes from outside.


pvh:
Yeah, I agree. I'll get it into the next point release.

PR #353, opened 2024-05-25
fix(websocket): ProtocolV1 shouldn't be exported as type because it's a string

yurijmikhalevich:
Hi everyone! :) Thank you for the great project.

I have noticed that the example WebSocket code doesn't work with `ProtocolV1` being exported as type:
image

This PR fixes that.


pvh (approved):

pvh:
Thank you! Good catch.

yurijmikhalevich:
@pvh, thank you and the team for the great library 🔥

PR #355, opened 2024-06-21
New wasm packaging

alexjg:
This PR uses the new wasm packaging introduced in https://github.com/automerge/automerge/pull/938. The main change is that `automerge-repo` now only depends on `@automerge/automerge/slim`, which means that it does not automatically pull in the wasm blob, leaving initialization up to the application. This allows the use of automerge in environments where initializing the wasm as part of a module import is not viable.


packages/automerge-repo-react-hooks/test/useHandle.test.tsx
```diff
@@ -6,11 +7,14 @@ import {
...
 import { DummyStorageAdapter } from "@automerge/automerge-repo/test/helpers/DummyStorageAdapter"
 import { render, waitFor } from "@testing-library/react"
-import React from "react"
 import { describe, expect, it, vi } from "vitest"
 import { useHandle } from "../src/useHandle"
 import { RepoContext } from "../src/useRepo"
 
+import { next as A } from "@automerge/automerge"
+// Do something with 'next' so that it isn't compiled away
+const someDoc = A.init()
```
pvh, line 16: Uhhhh, we should talk about this!

alexjg: I have a solution! Unfortunately the solution (which involves subpath exports on the `@automerge/automerge-repo` package) has broken use of things like

```typescript
import { eventPromise } from "@automerge/automerge-repo/src/helpers/eventPromise"
```

which appears in lots of the tests . This means I need to find a proper place to put all of these helpers. Once I've figured that reorganization out I'll push an update.

alexjg:
Alright, I've updated how this works a little bit. Let me describe the problem a little before describing how we solve it. 

The new `@automerge/automerge` package exposes two subpath exports. One, the `./` export (i.e. `import am from "@automerge/automerge"}` is the same as the current version, it exposes a package layout which expects you to provide a bundler or other mechanism for natively loading the wasm module. The new thing is a `./slim` export (`import am from "@automerge/automerge/slim"`) which does not load the wasm module and leaves that up to the client.

The problem for `automerge-repo` is that we import `@automerge/automerge` but we don't want to, we're a library and we want to leave that up to the user. This seems to suggest that we should just import `@automerge/automerge/slim`. Unfortunately this would break any existing users because they _don't_ perform wasm initialization. 

So, what we need is a way for users to opt in to the wasm-less version of `automerge-repo`. To achieve this we expose an additional subpath export: `@automerge/automerge-repo/slim`, which doesn't perform wasm initialization.

There are now two subpath exports in the `automerge-repo` package, they look like this:

```json
"exports": {
    ".": "./dist/entrypoints/fullfat.js",
    "./slim": "./dist/entrypoints/slim.js",
}
```

The `fullfat` entry point imports automerge from `@automerge/automerge` before running `export * from `../index.js` whilst `/slim` doesn't. 

There is one possible concern here, any code we write which accidentally imports `@automerge/automerge` rather than `@automerge/automerge/slim` would trigger the wasm initialization unconditionally, breaking users who are relying on the `/slim` package. To combat this we should all just be really careful.

No that's a joke, I added a custom ESlint rule which throws an error if it sees a non-slim import of `@automerge/automerge` or `@automerge/automerge-repo`. Writing this rule in an ergonomic way required updating to ESLint 9, which has a slightly different config format. I've put that part in a separate commit.

PR #359, opened 2024-07-15
Network disconnection support & repo shutdown

pvh:
We've had reports for a while of problems with the websocket network client not actually supporting a proper disconnect. 

This PR works on that as well as working on a repo.shutdown().

I've removed the LeaveMessage because websocket has native support for a disconnection frame that gets sent when you close() a socket; in the event of a crash / timeout we still have the existing no-message fallback.

I also added a test to make sure that we can in fact reconnect based on existing tests. I'm not sure if that's the right approach: I'd prefer to have a test that actually makes sure messages get through. Maybe in the next revision.

I'd love @HerbCaudill or @acurrieclark to at least eyeball this and see if he can think of anything I missed but once this lands it should make a bunch of demo / test code happier as well as improve support for ephemeral things like CLI tooling.


pvh:
Okay! Lots of minor progress here (yes, I know the linter is mad, I gotta figure out why my prettier says it's running but isn't working anymore). 

I've replaced the `"ready"` event with `isReady()/whenReady()` on the NetworkAdapter, and plumbed through the appropriate bits out to implement repo.shutdown().

This code all works fine if the socket manages to connect before you call it, but I spent a good hour or two trying to figure out how to handle shutting down a websocket connection during its initial CONNECTING readyState without much success and now it's late and I'm going to call it a night. (If anyone has experience with this please chime in.)

My aim is to merge this tomorrow so with any luck I can track down the weird little gotchas in there.

pvh:
Okay, I think this branch looks pretty good but it *does* change the interface for a NetworkAdapter so we might want to think about that. Also, .shutdown() probably needs some proper tests before landing this.

pvh:
We should support disconnect properly for the other network adapter types: broadcast & messagechannel at least (and add a test to the shared suite!)

packages/automerge-repo/src/network/NetworkSubsystem.ts
```diff
@@ -157,26 +164,17 @@ export class NetworkSubsystem extends EventEmitter<NetworkSubsystemEvents> {
   }
 
   isReady = () => {
-    return this.#readyAdapterCount === this.#adapters.length
+    return this.adapters.length === 0 || this.adapters.every(a => a.isReady())
```
pvh, line 167: don't need .length check actually -- every should be true for []

packages/automerge-repo/src/network/NetworkSubsystem.ts
```diff
@@ -157,26 +164,17 @@ export class NetworkSubsystem extends EventEmitter<NetworkSubsystemEvents> {
...
   whenReady = async () => {
-    if (this.isReady()) {
-      return
-    } else {
-      return new Promise<void>(resolve => {
-        this.once("ready", () => {
-          resolve()
-        })
-      })
+    if (this.adapters.length == 0) {
```
pvh, line 171: again we don't actually need this

pvh:
Alright, I think this is good to go!

PR #361, opened 2024-07-17
Remove isNew from DocHandle

pvh:
The DocHandle state machine is too complicated and too intertwingled with other systems -- DocHandles have all this responsibility to understand the network, storage, and so on.

This is the start of a de-scoping of DocHandle's responsibilities to become a much simpler wrapper around the underlying Automerge.Doc's state by giving Repo some additional responsibilities (at least in the short term).

I intend to continue down this path and simplify the state machine so it has just four states:

LOADING ---------> READY -----> DELETED
    +--UNAVAILABLE+--^

This should also help with implementing an LRU cache. Alex and I discussed how we could transition a DocHandle back to LOADING and free up the doc inside. If the user requests the doc, we'd just reload / resync it (as required.) Probably more thinking to do around that but it feels directionally promising.

### DIFF
--- a/packages/automerge-repo/src/helpers/tests/network-adapter-tests.ts
+++ b/packages/automerge-repo/src/helpers/tests/network-adapter-tests.ts
@@ -31,7 +31,7 @@ export function runNetworkAdapterTests(_setup: SetupFn, title?: string): void {
   describe(`Network adapter acceptance tests ${
     title ? `(${title})` : ""
   }`, () => {
-    it.only("can sync 2 repos", async () => {
+    it("can sync 2 repos", async () => {
       const doTest = async (
         a: NetworkAdapterInterface[],
         b: NetworkAdapterInterface[]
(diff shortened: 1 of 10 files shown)


packages/automerge-repo/src/helpers/tests/network-adapter-tests.ts
```diff
@@ -31,7 +31,7 @@ export function runNetworkAdapterTests(_setup: SetupFn, title?: string): void {
   describe(`Network adapter acceptance tests ${
     title ? `(${title})` : ""
   }`, () => {
-    it("can sync 2 repos", async () => {
+    it.only("can sync 2 repos", async () => {
```
alexjg, line 34: :eyes:

alexjg:
LGTM once we remove the `.only` then everything should path and we should merge IMO.

alexjg:
I rebased on `main` and removed the `.only` calls. Unfortunately the removal of the `.only` calls exposed a bunch of failing tests. I haven't had a chance to look into them yet.

pvh:
Fixed most tests by moving `doneLoading()` to the correct spot. There are five tests left failing which I suspect are all caused by the storage engine not getting registered at the right moment (not sure if it's test-specific behaviour or what) but it's too late to keep poking at it. Feel free to have a look @alexjg but I'll plan to spend more time tomorrow.

PR #362, opened 2024-07-23
Document id / url type guards can accept unknown

marionauta:
- Type guards usually accept `unknown` value type to be as generic as possible.
- It allows more values to be checked, like `undefined`. The `isValidAutomergeUrl` was already accepting `null` and `undefined` values. This PR extends that.
- This shouldn't break existing code since users are already using the guards with `string`s.

```typescript
isValidDocumentId(undefined) // false
isValidDocumentId(1) // false
```


pvh:
Good point! I added a test for a previously unrequestable value and merged it. Thanks.

Issue #330, opened 2024-03-30
Error: recursive use of an object detected which would lead to unsafe aliasing in rust


georgewsu:
Hi, running in to this error when testing an automerge client with a 10MB test document.

Node.js v18.15.0

Steps to reproduce:
https://github.com/georgewsu/automerge-client-test/tree/error_recursive_use

```
git clone https://github.com/georgewsu/automerge-client-test.git
cd automerge-client-test
git checkout error_recursive_use
npm install
npm run client

```

code excerpt:
https://github.com/georgewsu/automerge-client-test/blob/error_recursive_use/src/client.ts
```
const stringArray: string[] = new Array(arrayLength).fill(stringWith100Characters);
const testJson = {
  stringArray: stringArray
};
const testDoc = A.from(testJson);
const handle1 = repo1.create(testDoc);
```

Result:
```
Error: recursive use of an object detected which would lead to unsafe aliasing in rust
    at module.exports.__wbindgen_throw (/Users/gsu/repo2/automerge-client-test/node_modules/@automerge/automerge-wasm/nodejs/automerge_wasm.cjs:1990:11)
    at wasm_bindgen::throw_str::hc0015d8a92fcec96 (wasm://wasm/00664e02:wasm-function[2065]:0x157683)
    at wasm_bindgen::__rt::borrow_fail::hd13a3345a14a3b49 (wasm://wasm/00664e02:wasm-function[2067]:0x15769f)
    at automerge_rollback (wasm://wasm/00664e02:wasm-function[1668]:0x1545ce)
    at Automerge.rollback (/Users/gsu/repo2/automerge-client-test/node_modules/@automerge/automerge-wasm/nodejs/automerge_wasm.cjs:544:26)
    at _change (file:///Users/gsu/repo2/automerge-client-test/node_modules/@automerge/automerge/dist/mjs/stable.js:378:22)
    at Module.from (file:///Users/gsu/repo2/automerge-client-test/node_modules/@automerge/automerge/dist/mjs/stable.js:176:12)
    at Module.from (file:///Users/gsu/repo2/automerge-client-test/node_modules/@automerge/automerge/dist/mjs/next.js:98:19)
    at new DocHandle (file:///Users/gsu/repo2/automerge-client-test/node_modules/@automerge/automerge-repo/dist/DocHandle.js:51:21)
    at #getHandle (file:///Users/gsu/repo2/automerge-client-test/node_modules/@automerge/automerge-repo/dist/Repo.js:244:24)

[...]

Issue #358, opened 2024-07-11
Inactive `DocHandle`s are not evicted from `handleCache`


yarolegovich:
Hi, wanted to share my view on the DocHandle eviction problem and hear your thoughts. If the suggested approach sounds reasonable, I can start working on it this or early next week.

The problem:
DocHandles are hard-referenced in `Repo#handleCache` and never destroyed. The longer the program runs the more "idle" documents there are. 

Constraints:
* We definitely don't want to unload a document if client code is working with it (holds a reference).
* We don't want to unload a document another peer is actively replicating to us, so that we don't need to load it back to memory on every new message.
* There might be >10k handles if objects are small and each reside in their own document, unlikely >100k (?). 

Suggested solution:
1. Rename `DocHandle` to  `InternalDocHandle`, all the private logic is going to stay here.
2. Add `DocHandle` that'll hold a hard-ref to `InternalDocHandle`. Public API goes here.
3. `InternalDocHandle` holds a weak-ref to `DocHandle` that hard-references it. We need this to avoid unloading documents referenced by client code.
4. Maintain a lastChange timestamp on `InternalDocHandle`. We need it to avoid unloading documents which are being replicated.
5. `setInterval` to scan all handles, removing those where `DocHandle` weak-ref is undefined and it's been more than a minute (?) since the last document update. Should probably be user-configurable.
6. Scan can be incremental, yielding after 10k items not to hold event-loop for too long, we're ok with eventual eviction. We can continue the scan from where we left on the next interval trigger or just schedule a macro-task to continue after other callbacks had a go. I'd probably go with this approach.

Alternatives:
1. Maintain an LRU-cache, moving handles in response to `change` events, when garbage-collecting we can iterate from least to most recently used and stop iteration as soon as time passed since `lastChange` doesn't exceed eviction threshold. Not a bad idea, but we don't gain much if user code holds references to a lot of objects which rarely change and sit in the tail. We'll need to go through all of them first. In addition with our data set size reference walking can be worse than linear array scan even if the latter checks more elements.
2. Don't introduce `InternalDocHandle`. Let's say we maintain an LRU cache as in alternative#1 and based on `lastChange` timestamp we move handles from strong-ref `active` cache to `inactive: Map<DocumentId, WeakMap<DocHandle>>` cache. On sync message or `find` we can look up in both and return an existing handle which was still in-use by user-code. On `change` we can bring handles back to `activeCache` if they weren't there. Don't like this solution because: 
[...]

PR #365, opened 2024-07-30
Add support for clearing doc from DocHandle

georgewsu:
Summary:
Add support for clearing doc from DocHandle so that reference can be released and memory freed, without deleting document from storage

Issue:
https://github.com/automerge/automerge-repo/issues/330

Background:
After creating enough documents in a repo, out of memory errors will cause a synch server to crash because memory isn't being freed. Repo holds references to each DocHandle in handleCache and each DocHandle holds a reference to the automerge doc. 

Proposed solution:
https://github.com/georgewsu/automerge-repo/tree/gsu/cache-eviction-test2 has code for an initial implementation of cache eviction and releasing document reference so that memory can be freed. I've tested this with a sync server running locally.

Part 1:
This PR only covers introducing a new state and event to DocHandle to allow clearing the document.


georgewsu:
I've discussed this with @pvh but he is on holiday now

alexjg:
Thanks for putting a bunch of work into this and the follow up PRs. I'm going to post my comments here but this feedback also relates to the linked work.

The addition of a new state to the `DocHandle` makes me uneasy. I've been thinking about why and I think it's because from the perspective of the application, the `CLEARED` state is not useful. I think instead we should just re-use the `LOADING` state. This is important because we need to figure out what to do if we evict a document which the application has a live `DocHandle` for (as discussed in #358 ). There seems to me to be not much difference between a handle which is loading because it is waiting for storage or network and a handle which is loading because the `Repo` has decided to evict it.

As to how to the time based cache expiration policy. I think there are some tactical problems with it - the `accessTime` seems to me like it would be based on when the handle was created, not on the last time the application used it. Maybe that works for some usecases but it doesn't seem very general. This feels more like something which should be configurable in some way, maybe by passing a function to the `Repo.clearCache` method.

georgewsu:
> Thanks for putting a bunch of work into this and the follow up PRs. I'm going to post my comments here but this feedback also relates to the linked work.
> 
> The addition of a new state to the `DocHandle` makes me uneasy. I've been thinking about why and I think it's because from the perspective of the application, the `CLEARED` state is not useful. I think instead we should just re-use the `LOADING` state. This is important because we need to figure out what to do if we evict a document which the application has a live `DocHandle` for (as discussed in #358 ). There seems to me to be not much difference between a handle which is loading because it is waiting for storage or network and a handle which is loading because the `Repo` has decided to evict it.
> 
> As to how to the time based cache expiration policy. I think there are some tactical problems with it - the `accessTime` seems to me like it would be based on when the handle was created, not on the last time the application used it. Maybe that works for some usecases but it doesn't seem very general. This feels more like something which should be configurable in some way, maybe by passing a function to the `Repo.clearCache` method.

Sounds good, agree with not having another state in `DocHandle` if it isn't necessary. @pvh had originally suggested reusing the `IDLE` state but I wasn't sure how that would work since the constructor sends the `BEGIN` event to transition to `LOADING` state. I'll work on removing the added state and reusing the `LOADING` state instead and will test it out.

The intention of `accessTime` is to track the last time the handle was used - will follow up with you on that approach / implementation details / changes need to support that.

Thanks!

pvh:
Just a note that I'm going to let @alexjg be the reviewer for this one (unless he asks for help) since he's already up to speed. I'm mostly done refactoring in this area -- I am considering retiring xstate just to reduce dependencies (cc/ @HerbCaudill), but I'll try and wait until after this lands.

packages/automerge-repo/test/DocHandle.test.ts
```diff
@@ -325,6 +326,52 @@ describe("DocHandle", () => {
...
+    handle.change(doc => {
+      doc.foo = "bar"
+    })
+    const doc = await handle.doc()
+    assert.equal(doc?.foo, "bar")
+
+    handle.reset()
+
+    // remain idle beyond timeoutDelay
+    await pause(10)
```
alexjg, line 357: I think we should assert that we are in the IDLE state after this call, otherwise this test passes even if `reset` is a no-op.

packages/automerge-repo/src/DocHandle.ts
```diff
@@ -426,6 +434,11 @@ export class DocHandle<T> extends EventEmitter<DocHandleEvents<T>> {
     this.#machine.send({ type: DELETE })
   }
 
+  /** Called by the repo to free memory used by the document. */
+  reset() {
```
alexjg, line 438: I think that rather than calling this `reset` we should all it `idle` as I think that more clearly describes what it actally does.

alexjg (commented):
I've left a few small comments. I have a bigger question which is bothering me. Currently it's quite common for people to depend on this idea that when a `DocHandle` is ready then the application can assume it will always be ready. For example, code like this is quite common:

```typescript
const handle = repo.find(...)
await handle.whenReady()
console.log(handle.docSync()) // <-- this line depends on the document never becoming unready
```

Introducing a transition back to `IDLE` breaks this assumption and it's the reason that in #358 @yarolegovich spent a lot of time trying to find a way to avoid evicting handles which the application is known to be using.

I _think_ that finding a solution to this problem is orthogonal to having a way to transition in the first place though so I think that - once the comments I've made are addressed - we can do that later.

pvh:
> I've left a few small comments. I have a bigger question which is bothering me. Currently it's quite common for people to depend on this idea that when a `DocHandle` is ready then the application can assume it will always be ready. For example, code like this is quite common:
> 
> ```ts
> const handle = repo.find(...)
> await handle.whenReady()
> console.log(handle.docSync()) // <-- this line depends on the document never becoming unready
> ```
> 

JS is single threaded -- as long as you don't yield, the above code is indeed safe. This is also a benefit of the promise-oriented API. Overall though, your point that introducing non-monotonic loading behaviour could have pitfalls is true but unavoidable. 

I think ideally, most people will never encounter this behaviour, but we might in fact want a new handle state (or to rename the state to perhaps UNLOADED) so that we can give better error messages if it does happen.

georgewsu:
Thanks for the reviews @alexjg @pvh 

Agree about the need to determine which handles are in use, will follow up with questions I had in existing threads.

For the handle state, will follow up with you about going with `IDLE` or `UNLOADED`.

alexjg:
It does seem like we need a new state if we want to provide good error messages, apologies for making you go around in circles. I think we should go for a new state called `UNLOADED` (distinct from `IDLE`) and rename `reset() -> unload()`.

georgewsu:
> It does seem like we need a new state if we want to provide good error messages, apologies for making you go around in circles. I think we should go for a new state called `UNLOADED` (distinct from `IDLE`) and rename `reset() -> unload()`.

Great - that sounds good, I'll make that change. Do you think `UNLOADED` should be a terminal state, or should it allow reusing the handle and reloading, either via a new `RELOAD` event or reusing the `BEGIN` event? If reloading is not needed yet, I could start with the terminal state and leave it to be refactored later.

alexjg:
I think we should definitely support reloading via `BEGIN` yeah but I'm happy to leave that for a future PR of it's complicated.

georgewsu:
> I think we should definitely support reloading via `BEGIN` yeah but I'm happy to leave that for a future PR if it's complicated.

Cool, @alexjg just saw your comment - I actually just pushed 3 versions:
1. [234f838](https://github.com/automerge/automerge-repo/pull/365/files/234f83890049222f399fc1f5556c168edb49c48c) which has unloaded state as final
2. [42706f7](https://github.com/automerge/automerge-repo/pull/365/files/42706f72f43ebcd48130e0a3ee59cdb4669838c5) which allows calling `begin()` again
3. [8f0f839](https://github.com/automerge/automerge-repo/pull/365/files/8f0f839082daf39abe7e7d46564d20055635c2dc) which adds `reload()`

Happy to go with any of these

alexjg:
@georgewsu I likke `8f0f`! If you can fix the lint I'll merge.

PR #367, opened 2024-07-31
make sure we start synchronising & save documents even if initially loaded from storage

diederich:
We found using `IndexedDBStorageAdapter` storage adapter, documents wouldn't start syncing in case they were found in the storage. Equally, they wouldn't get stored to the storage system. 
I *think* there's a `document` event missing, which in the end adds the doc to both the storage & sync machinery.

This PR adds the missing event and a test.

I'm sure there's a better way for testing than hooking into the `document` event? Happy to adjust!

### DIFF
--- a/packages/automerge-repo/src/Repo.ts
+++ b/packages/automerge-repo/src/Repo.ts
@@ -441,8 +441,8 @@ export class Repo extends EventEmitter<RepoEvents> {
             .catch(err => {
               this.#log("error waiting for network", { err })
             })
-          this.emit("document", { handle })
         }
+        this.emit("document", { handle })
       })
     } else {
       handle.request()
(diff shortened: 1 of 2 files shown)


BrianHung:
Encountered bug of indexeddb not persisting on page refresh when upgrading automerge-repo from 1.2.1 to 2.0.0.

pvh:
Hey thanks so much! That's why it's an alpha still... Sorry I've been out on holiday the last couple of weeks and left you hanging.

I probably won't take this patch as-is (I agree it should have a different test) but this absolutely helped me and @alexjg figure out what the problem is and it is greatly appreciated.

diederich:
No worries at all! And thanks for looking into this. Feel free to either push on top or just close this! 🙏

pvh:
Okay, @diederich, take a look here: https://github.com/automerge/automerge-repo/pull/374

I think this is a better test of the bug you describe but it's passing. Do you have any idea why you might be seeing different behaviour?

diederich:
Closing this, as fixed on `main` with https://github.com/automerge/automerge-repo/commit/b9a212e3048186fa91f90c988fcf37d8c9855129 
Thank you! 🙏

PR #369, opened 2024-08-05
failing test: loading empty doc from network

geoffreylitt:
In Jacquard, we have a use case where we sometimes try to load an empty doc from network. It's a metadata doc which sometimes has some stuff in it, but other times it starts out just being empty.

It turns out that if you do this, automerge-repo will never resolve the `.doc()` promise, so our app hangs. (This PR has a failing test for that)

The underlying cause is this code in DocHandle, which uses heads to decide whether to do a state transition:

```
 #checkForChanges(before: A.Doc<T>, after: A.Doc<T>) {
    const beforeHeads = A.getHeads(before)
    const afterHeads = A.getHeads(after)
    const docChanged = !headsAreSame(afterHeads, beforeHeads)
    if (docChanged) {
      this.emit("heads-changed", { handle: this, doc: after })
```

The "empty" handle pre-load has heads `[]`, and the doc loaded from network also has heads `[]`, so the state transition doesn't happen.

@joshuahhh and I got this far and we weren't sure what made sense to do next, thoughts @alexjg ?

For now we were able to workaround in our app by just making a change in the empty docs to avoid this case.

### DIFF
--- a/packages/automerge-repo/test/DocHandle.test.ts
+++ b/packages/automerge-repo/test/DocHandle.test.ts
@@ -153,6 +153,22 @@ describe("DocHandle", () => {
     assert.equal(doc?.foo, "bar")
   })
 
+  it("should become ready if the network returns an empty document", async () => {
+    const handle = new DocHandle<TestDoc>(TEST_ID)
+
+    // we don't have it in storage, so we request it from the network
+    handle.request()
+
+    // simulate updating from the network with an empty document
+    handle.update(doc => {
+      return A.init()
+    })
+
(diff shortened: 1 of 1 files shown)


alexjg:
Interesting, in the latest alpha of `automerge-repo` I think we may fix this. We made the decision to always create a document with an empty change in it to distinguish a new document from one we are requesting. This is effectively what you are doing as well.

geoffreylitt:
Great, sounds like that'll fix it! I'll just close this issue and then we can confirm fixed once we upgrade to that.

PR #374, opened 2024-08-13
Sync bug test

pvh:
Here's a test that I think exercises the bug you describe (as well as a PNPM update)... but it's passing!

I wonder if it's because IndexedDB is slower, somehow? Did I not understand the description of the bug you reported?

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
   "engines": {
     "node": ">= 18.x"
   },
-  "packageManager": "pnpm@8.11.0",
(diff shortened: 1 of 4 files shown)


pvh:
Okay... right, so looking at this more closely I realize that when I see if the data is being synced to charlie... it causes it to be synced to charlie, which is why this test was passing. I just pushed a fix to main which exposes the synchronizer for testing / debugging purposes but flags them as hidden since I'm not promising to maintain that API.

PR #375, opened 2024-08-15
Allow useDocuments to immediately return already-loaded documents

msakrejda:
This is essentially the useDocuments counterpart of #318.


packages/automerge-repo-react-hooks/test/useDocuments.test.tsx
```diff
@@ -62,6 +62,18 @@ describe("useDocuments", () => {
     )
   })
 
+  it("returns a collection of loaded documents immediately, given a list of ids", async () => {
```
msakrejda, line 65: This sort of duplicates the above test. I wasn't sure if this should be its own test or be folded into the above.

PR #379, opened 2024-08-27
Update useDocuments to return map keyed by AutomergeUrl

msakrejda:
The recommended mechanism seems to be to store Automerge URLs when referencing other documents, not just plain document ids, so it seems more natural to key the map returned from useDocuments by URL rather than by id.

See discussion in https://discord.com/channels/1200006940210757672/1202325266937159690/1277333966575374356


pvh:
Closing this because although I didn't use this patch I did update useDocuments to work this way as part of integrating your suspense work as well!

PR #380, opened 2024-08-28
Add support for removing doc handle from handleCache

georgewsu:
**Summary:**
Add support for removing doc handle from handleCache to free memory, without deleting document from storage.

**Issue:**
https://github.com/automerge/automerge-repo/issues/330

**Background:**
After creating enough documents in a repo, out of memory errors will cause a synch server to crash because memory isn't being freed. Repo holds references to each DocHandle in handleCache and each DocHandle holds a reference to the automerge doc.

**Next steps:**
Once `removeFromCache` is added, it can be called from the Repo and/or synch server to free memory usage (that code must determine that the doc handle is safe to be removed from cache at that time).

### DIFF
--- a/packages/automerge-repo/src/Repo.ts
+++ b/packages/automerge-repo/src/Repo.ts
@@ -6,7 +6,14 @@ import {
   interpretAsDocumentId,
   parseAutomergeUrl,
 } from "./AutomergeUrl.js"
-import { DocHandle, DocHandleEncodedChangePayload } from "./DocHandle.js"
+import {
+  DELETED,
+  DocHandle,
+  DocHandleEncodedChangePayload,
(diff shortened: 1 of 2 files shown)


alexjg:
This looks good but could you add a typedoc `@experimental` tag to the `removeFromCache` method as I think it's quite likely we'll change this API a bit before we're sure of how the cache eviction stuff should work.

georgewsu:
> This looks good but could you add a typedoc `@experimental` tag to the `removeFromCache` method as I think it's quite likely we'll change this API a bit before we're sure of how the cache eviction stuff should work.

Sounds good, also added a check for `documentId` being present before proceeding to remove it.

PR #382, opened 2024-09-03
rename poorly named WebSocket packages

pvh:
The old names are preserved for backwards compatibility via hidden exports. I didn't add "Network" to the name (which would be more consistent) but WebSocketClientNetworkAdapter is such a finger-full... does anyone have opinions?


HerbCaudill:
omg thank you this has bugged me forever.

This looks fine to me, you could even just call them `ClientAdapter` and `ServerAdapter`.

HerbCaudill (approved):

pvh:
Landed, at long long last.

PR #384, opened 2024-09-06
Remove unused `@ibm/plex` dependency

### LINKED ISSUE #383: Remove `@ibm/plex` as dependency from `react-todo` example
During my initial `pnpm install` I noticed a fair amount of time was spent downloading the [`@ibm/plex`](https://www.npmjs.com/package/@ibm/plex) package (a 187 mb package with the IBM Plex typeface). I thought to myself, this must be an amazing font - I look forward to seeing it in the example app only to find that it's never actually imported.

The tailwind config does reference it by name: https://github.com/automerge/automerge-repo/blame/main/examples/react-todo/tailwind.config.cjs but there's no `@font-face` nor `@import` rules in the projects CSS files to actually load in the font.

As this is effectively an unused dependency (and it's "just an example"), I'd suggest just getting rid of the dependency all together instead of fixing the missing `@import`.

kraenhansen:
This fixes #383.

NOTE: For some reason the `pnpm-lock.yaml` version gets upgraded from 6 to 9, although I'm using `pnpm@9.7.0` as declared in the root `package.json`:

https://github.com/automerge/automerge-repo/blob/7c486df03d5936ab0d4258040e3cc8f590967a2f/package.json#L31

### DIFF
--- a/examples/react-todo/tailwind.config.cjs
+++ b/examples/react-todo/tailwind.config.cjs
@@ -1,22 +1,10 @@
 /** @type {import('tailwindcss').Config} */
 const colors = require("tailwindcss/colors")
 
-const emoji = "Segoe UI Emoji"
-const mono = "IBM Plex Mono"
-const sans = "IBM Plex Sans"
-const condensed = "IBM Plex Sans Condensed"
-const serif = "IBM Plex Serif"
-
 module.exports = {
   content: ["./**/*.{html,tsx}"],
   theme: {
     extend: {
-      fontFamily: {
-        mono: [mono, emoji, "monospace"],
-        sans: [sans, emoji, "sans-serif"],
-        condensed: [condensed, emoji, "sans-serif"],
-        serif: [serif, emoji, "serif"],
-      },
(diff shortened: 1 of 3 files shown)


examples/react-todo/tailwind.config.cjs
```diff
@@ -11,12 +11,6 @@ module.exports = {
   content: ["./**/*.{html,tsx}"],
   theme: {
     extend: {
-      fontFamily: {
-        mono: [mono, emoji, "monospace"],
-        sans: [sans, emoji, "sans-serif"],
-        condensed: [condensed, emoji, "sans-serif"],
-        serif: [serif, emoji, "serif"],
-      },
```
kraenhansen, line 19: It seems fine to rely on [the defaults](https://tailwindcss.com/docs/font-family).

kraenhansen:
cc @HerbCaudill as you're the original author on this: https://github.com/automerge/automerge-repo/pull/14

pvh:
This seems very much worth the fix. Thanks!

PR #385, opened 2024-09-06
Refactor the storage adapter tests to use a `beforeEach` hook

kraenhansen:
To implement tests of a storage which use async APIs, I need the `teardown` function to be able to return a promise which gets awaited. I also took the liberty to refactor the tests a bit to use a `beforeEach` hook instead of calling `setup` and `teardown` in each of the tests. This ensures the `teardown` gets called even if the test fails. In the current state, this isn't the case since something throwing in the test could prevent the call to `teardown`.

### DIFF
--- a/packages/automerge-repo/src/helpers/tests/storage-adapter-tests.ts
+++ b/packages/automerge-repo/src/helpers/tests/storage-adapter-tests.ts
@@ -1,4 +1,4 @@
-import { describe, expect, it } from "vitest"
+import { describe, expect, beforeEach, test } from "vitest"
 
 import type { StorageAdapterInterface } from "../../storage/StorageAdapterInterface.js"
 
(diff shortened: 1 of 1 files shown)


kraenhansen:
cc @HerbCaudill as it seems you've worked on this file too 👍

packages/automerge-repo/src/helpers/tests/storage-adapter-tests.ts
```diff
@@ -8,72 +8,61 @@ const PAYLOAD_C = () => new Uint8Array([2, 111, 74, 131, 236, 96, 142, 193])
...
-export function runStorageAdapterTests(_setup: SetupFn, title?: string): void {
-  const setup = async () => {
-    const { adapter, teardown = NO_OP } = await _setup()
-    return { adapter, teardown }
-  }
+type AdapterTestContext = {
+  adapter: StorageAdapterInterface
+}
+
+const it = test<AdapterTestContext>
```
HerbCaudill, line 15: small nitpick - I'd prefer 
```ts
import { ..., it as _it } from "vitest"

...

const it = _it<AdapterTestContext>
```

HerbCaudill (changes requested):

HerbCaudill:
Hi @kraenhansen. Thanks for working on this, and for introducing me to Vitest's test context API! I've leaned heavily on the `setup()` pattern to avoid `beforeEach` + global mutable state, but this is much cleaner. 

(For anyone following along, it took me a while to find this in the docs - it's not referenced in the [API reference](https://vitest.dev/api/#beforeeach), but it's documented in a standalone guide to [Test Context](https://vitest.dev/guide/test-context.html#beforeeach-and-aftereach).)

packages/automerge-repo/src/helpers/tests/storage-adapter-tests.ts
kraenhansen: Done 👍

HerbCaudill (approved):

PR #386, opened 2024-09-09
Storage adapter for MongoDB

kraenhansen:
Implements the `StorageAdapterInterface` targeting MongoDB.

The constructor is overloaded to take either a `MongoDBClient` instance or a MongoDB server `url`.

### Key storage and query strategy

Chunk keys are stored and queried as arrays of strings and I've prepared the adapter to take an option to change this to storing keys as strings (joined by some separator). I believe using array of strings is "good enough" and the simplest of the two, since the range query would have to be implemented using the [$regex operator](https://www.mongodb.com/docs/manual/reference/operator/query/regex/) which would require sanitizing the key segments (and making assumptions on valid characters) before injecting them into the regular expression.

### Why is it draft?

This contains changes from bb489067a1b0d23000a4f10f95015928596e688d which should not be considered when reviewing this PR and as such, this branch needs a rebase onto `main` once https://github.com/automerge/automerge-repo/pull/385 gets merged 🤞

### DIFF
--- a/packages/automerge-repo-storage-mongodb/test/MongoDBStorageAdapter.test.ts
+++ b/packages/automerge-repo-storage-mongodb/test/MongoDBStorageAdapter.test.ts
@@ -1,0 +1,70 @@
+import cp from "node:child_process"
+import { beforeAll, describe } from "vitest"
+import { MongoClient } from "mongodb"
+import { runStorageAdapterTests } from "../../automerge-repo/src/helpers/tests/storage-adapter-tests"
+import { MongoDBStorageAdapter } from "../src"
+
+const MONGODB_URL = "mongodb://localhost:27017"
+const MONGODB_DB_NAME = "automerge-repo-test-db"
+const MONGODB_COLLECTION_NAME = "automerge-repo-storage"
+const MONGODB_DOCKER_CONTAINER_NAME = "automerge-repo-test-mongo"
+
+/**
+ * @returns true if the docker cli is available
+ */
+function hasDocker() {
+  try {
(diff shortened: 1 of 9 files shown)


alexjg:
This looks like a solid implementation and I agree with your design decisions. We will be _slightly_ changing the storage interface in future to explicitly disallow a separator character from the keys (probably `"/"`) which would make it possible to use a regex query I think, though I think your concerns about sanitizing the input remain.

However, I would prefer not to add more storage adapters to this repository. In my view this repository should contain only the most basic and universal storage adapters (i.e. the file system and indexeddb ones) and anything else should live in other repositories maintained by their users. This reduces maintenance burden on the core repository and also allows storage adapters to adapt to the specific needs of the people using them. The storage interface is deliberately designed to be simple with very few requirements to make this mode of use easier.

I hope you don't see this as a snub, this is great work and I think lots of people would benefit from it if you are willing to maintain it in another repository!

One thing I think we should do is add links to community maintained storage adapters in the README.

packages/automerge-repo-storage-mongodb/test/MongoDBStorageAdapter.test.ts
```diff
@@ -0,0 +1,70 @@
...
+    cp.execFileSync("docker", ["--version"], { encoding: "utf8" })
+    return true
+  } catch {
+    return false
+  }
+}
+
+let client: MongoClient
+
+describe.skipIf(!hasDocker())("MongoDBStorageAdapter", () => {
```
kraenhansen, line 26: Skipping tests if `docker` is not available on the machine.

PR #387, opened 2024-09-23
Slightly improved tests for loadRange method

bijela-gora:
This PR is more a question then a suggestion.

The question: should we care about an order in which `StorageAdapterInterface.loadRange` returns data?

When I utilized `expect.arrayContaining` in this commit https://github.com/bijela-gora/automerge-repo/commit/3238d7a6e2a7f8563aed17b554996c3c329894a8 to automerge-repo I'm sure I didn't consider that. Now, I'm not even sure why I used that asymmetric matcher. The `expect.arrayContaining` does not care about an order of elements inside arrays:
```javascript
describe('expect.arrayContaining', () => {
	it("doesn't care about order", () => {
		expect([0, 1, 2]).toEqual(expect.arrayContaining([2, 0]))
	})
})
```

I noticed that a day ago when I was trying to make a storage adapter based on `node:sqlite`. For some reasons `better-sqlite3` and `node:sqlite` for the same SQL statement returns results ordered differently. I tested that with latest version of better-sqlite3 and with latest Node.js because latest versions uses [the](https://github.com/WiseLibs/better-sqlite3/commit/8b584ef6a15caecf2b121a0891aa70d6e56f4f8f) [same](https://github.com/nodejs/node/commit/3caf29ea8869d4668193111ebdd390b63d7942dc) SQLite version

### DIFF
--- a/packages/automerge-repo/src/helpers/tests/storage-adapter-tests.ts
+++ b/packages/automerge-repo/src/helpers/tests/storage-adapter-tests.ts
@@ -67,20 +67,16 @@ export function runStorageAdapterTests(setup: SetupFn, title?: string): void {
         await adapter.save(["AAAAA", "snapshot", "yyyyy"], PAYLOAD_B())
         await adapter.save(["AAAAA", "sync-state", "zzzzz"], PAYLOAD_C())
 
-        expect(await adapter.loadRange(["AAAAA"])).toStrictEqual(
-          expect.arrayContaining([
-            { key: ["AAAAA", "sync-state", "xxxxx"], data: PAYLOAD_A() },
-            { key: ["AAAAA", "snapshot", "yyyyy"], data: PAYLOAD_B() },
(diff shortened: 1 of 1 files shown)


pvh:
Hi @bijela-gora -- sorry for the delay in getting back to you on this but to answer your question, I think we should *not* insist on ordering range data unless a serious need arises. If we guarantee an order then providers that aren't ordered will have to sort their keys before returning them and right now Automerge doesn't actually require the responses to be ordered (though it does go faster if they're at least semi-ordered).

pvh:
(Landing this stricter test.)

bijela-gora:
Thank you for the response!
If the ordering is not important, probably [this change](https://github.com/automerge/automerge-repo/pull/387/files#diff-12e92867ad3c0285a974a0e32aa7379b453f211095e195d2f08a6edd1899ed1eL70-R79) should be reverted.

- the `toStrictEqual([])` does care about order.
- the `toStrictEqual(expect.arrayContaining()` doesn't.

PR #388, opened 2024-09-24
allow for initial id in repo create

BrianHung:
Allows for specified id in `repo.create`.

One thing: in the conversation about [pre-defined ids](https://discord.com/channels/1200006940210757672/1230453235207110666/1231199773571158068) and about [loading from binary](https://automerge.org/docs/cookbook/modeling-data/#setting-up-an-initial-document-structure), does `automerge.from` clone the document that is passed in? 

I'm wondering if this allows for the following:
```ts
const copy = {
  content: new Uint8Array([133, 111, 74, 131, ...]),
  id: ...
}

const doc = am.load(copy.content)
const handle = repo.create(doc, copy.id)
```
so that the initial value in `repo.create` shares ancestry with the binary that is passed in.


adrianduke:
Nudge for this functionality! I was just about to write this PR before I saw this... Can provide a hand if needed.

BrianHung:
Feel free to jump into this discord thread: https://discord.com/channels/1200006940210757672/1287965416689238047/1288223592936374323 The main blocker is how should we handle conflicting ids.

adrianduke:
I don't appear to be able to access from that link, I'm not currently apart of the discord either so that might be why!

PR #394, opened 2024-10-14
Add a denylist of automerge URLs to Repo constructor

alexjg:
Problem: some documents can consume so much memory at runtime and require so long to load that they render a server unresponsive.

Solution: add a denylist argument to Repo which contains a list of documents that the repo will not attempt to load when receiving sync messages. This is a fairly hacky stop-gap solution, but it is _a_ solution.


pvh:
Thanks, I hate it!

PR #399, opened 2024-11-30
Fix memoization of ids in useDocuments.ts

msakrejda:
Right now, useDocuments tries to memoize the list of ids/urls argument as it normalizes it to ids to avoid doing unnecessary work if the list is not changing. However, it does this by passing the ids as a member of the useMemo dependencies array. If callers are passing the array as a literal, this will create a new array on every render, and the memoization won't accomplish anything.

To avoid this, we could spread the array to add its members (the real dependencies) to our useMemo dependencies, but since there are no other dependencies, we can pass the ids array itself instead.

### DIFF
--- a/packages/automerge-repo-react-hooks/src/useDocuments.ts
+++ b/packages/automerge-repo-react-hooks/src/useDocuments.ts
@@ -28,7 +28,7 @@ export const useDocuments = <T>(idsOrUrls?: DocId[]) => {
           return idOrUrl as DocumentId
         }
       }) ?? [],
-    idsOrUrls
+    idsOrUrls ?? []
   )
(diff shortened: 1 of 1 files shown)


FAKHRUL12345:
Dearest hope well.At present  i dont comment due to stay in the world  for kindess of some humans.I am totally hopeless everywhere. At now dont think little bit my life.I am totally  unable to fill up  minimum  demand of my two small babies. 

But i always salute GitHub to  always nearest me.Thsbks to all/Fakhrul

FAKHRUL12345 (approved):

FAKHRUL12345:
> Best of luck

PR #401, opened 2024-12-07
Clean up useDocument hook

msakrejda:
- drop unnecessary useRef
 - drop unnecessary check for id (a null id cannot yield a non-null handle)
 - drop unnecessary function wrapping a rerender callback


pvh:
obviated by the rewrite -- thanks for all your help with this so far

PR #402, opened 2024-12-30
The end of undefined docSync() / await doc()

pvh:
After many long conversations with automerge-repo users, I have finally seen the error of my ways. I see now that `repo.find()` should not be synchronous and should instead return a promise to a handle that resolves when the handle is actually loaded.

This change should integrate better with suspense and radically reduce the amount of annoying awaits, undefined checks, and other irritations when consuming documents in automerge-repo.

Here's an initial draft that passes most tests. I as I continue to work through the consequences and simplify the resulting library, I'll update it here.

Comments & questions welcome.

(Note that this PR probably currently breaks a few features such as unload() but we'll work through that stuff before landing.)


packages/automerge-repo/test/Repo.test.ts
```diff
@@ -1516,9 +1471,9 @@ describe("Repo", () => {
         eventPromise(client.networkSubsystem, "peer"),
       ])
 
-      const clientDoc = client.find(doc.url)
-      await pause(100)
-      assert.strictEqual(clientDoc.docSync(), undefined)
+      expect(async () => {
```
kraenhansen, line 1474: When passing an async function to expect the return value becomes a promise itself which needs to be awaited.

```suggestion
      await expect(async () => {
```

packages/automerge-repo/test/Repo.test.ts
```diff
@@ -1335,16 +1290,16 @@ describe("Repo", () => {
     const aliceDoc = aliceRepo.create()
     aliceDoc.change((doc: any) => (doc.text = "Hello world"))
 
-    const bobDoc = bobRepo.find(aliceDoc.url)
-    await eventPromise(bobDoc, "unavailable")
+    expect(async () => {
```
kraenhansen, line 1293:
```suggestion
    await expect(async () => {
```

packages/automerge-repo/test/Repo.test.ts
```diff
@@ -621,13 +610,13 @@ describe("Repo", () => {
...
         expect(
-          (await repo.find<{ foo: string }>(handle.documentId).doc()).foo
+          (await repo.find<{ foo: string }>(handle.documentId)).docSync().foo
         ).toEqual("first")
         // Really, it's okay if the second one is also flushed but I'm forcing the issue
         // in the test storage engine above to make sure the behaviour is as documented
-        expect(
-          await repo.find<{ foo: string }>(handle2.documentId).doc()
-        ).toEqual(undefined)
+        expect(async () => {
```
kraenhansen, line 617:
```suggestion
        await expect(async () => {
```

packages/automerge-repo/test/Repo.test.ts
```diff
@@ -706,7 +695,7 @@ describe("Repo", () => {
         }
 
         await connectedPromise
-
+        console.log("All connected")
```
kraenhansen, line 698:
```suggestion
```

packages/automerge-repo/test/Repo.test.ts
```diff
@@ -574,9 +562,10 @@ describe("Repo", () => {
       })
 
       // Could not find the document that is not yet saved because of slow storage.
-      const reloadedHandle = repo2.find<{ foo: string }>(handle.url)
+      expect(async () => {
```
kraenhansen, line 565:
```suggestion
      await expect(async () => {
```

packages/automerge-repo/test/Repo.test.ts
```diff
@@ -192,48 +192,36 @@ describe("Repo", () => {
...
-      assert.equal(doc, undefined)
-    })
-
-    it("emits an unavailable event when you don't have the document locally and are not connected to anyone", async () => {
-      const { repo } = setup()
-      const url = generateAutomergeUrl()
-      const handle = repo.find<TestDoc>(url)
-      assert.equal(handle.isReady(), false)
-      await eventPromise(handle, "unavailable")
+      expect(async () => {
```
kraenhansen, line 202:
```suggestion
      await expect(async () => {
```

packages/automerge-repo/test/Repo.test.ts
```diff
@@ -192,48 +192,36 @@ describe("Repo", () => {
 
     it("throws an error if we try to find a handle with an invalid AutomergeUrl", async () => {
       const { repo } = setup()
-      try {
-        repo.find<TestDoc>("invalid-url" as unknown as AutomergeUrl)
-      } catch (e: any) {
-        assert.equal(e.message, "Invalid AutomergeUrl: 'invalid-url'")
-      }
+      expect(async () => {
```
kraenhansen, line 195:
```suggestion
      await expect(async () => {
```

packages/automerge-repo/src/synchronizer/DocSynchronizer.ts
```diff
@@ -351,7 +351,13 @@ export class DocSynchronizer extends Synchronizer {
 
     this.#withSyncState(message.senderId, syncState => {
       this.#handle.update(doc => {
+        console.log(
+          "received sync message in state",
+          this.#handle.documentId,
+          this.#handle.state
+        )
```
kraenhansen, line 358: This probably need removal before merging too?

kraenhansen:
I see you're generally not awaiting the `expect` calls some places. It seems the pattern is when you chain `.rejects.toThrow`. If there's a reason for this, feel free to disregard my comments, but it looks like an oversight to me. If the promise resolves "slowly" I suspect the tests won't fail.

msakrejda:
Nice, excited to see this revisited. Can you talk about how this affects [this diagram](https://automerge.org/docs/repositories/dochandles/#waiting-for-a-dochandle)?

pvh:
> Nice, excited to see this revisited. Can you talk about how this affects [this diagram](https://automerge.org/docs/repositories/dochandles/#waiting-for-a-dochandle)?

Ideally, all of those states except READY and DELETED will go away from a public consumer point of view. If you call find() and it doesn't reach a useful state, you'll get an exception.

That might happen because you're offline and trying to open a document you have heard of but didn't fetch yet for whatever reason, or because the peers you're connected to don't have the data.

That said, we won't give you the handle back from the promise unless and until you can actually use it! Long-term, I think some kind of signal-based model here would be preferable: you might want to be able to see incremental loading details to show some kind of progress bar. That said, we don't have that yet, so it feels premature to worry too much about it.

packages/automerge-repo-react-hooks/test/useDocument.test.tsx
```diff
@@ -64,150 +49,268 @@ describe("useDocument", () => {
...
+        <Component url={handleA.url} onDoc={onDoc} />
+      </Suspense>,
+      { wrapper }
+    )
 
-  it("should immediately return a document if it has already been loaded", async () => {
-    const { handleA, wrapper } = setup()
-    const onDoc = vi.fn()
+    // First we should see the loading state
+    expect(screen.getByTestId("loading")).toBeInTheDocument()
```
msakrejda, line 72: We expect to see a loading state even if we're dealing with a locally-created document?

packages/automerge-repo-react-hooks/test/useDocument.test.tsx
```diff
@@ -64,150 +49,268 @@ describe("useDocument", () => {
...
+    await waitFor(() => {
+      expect(screen.getByTestId("content")).toHaveTextContent("C")
+    })
+    expect(onDoc).toHaveBeenCalledWith({ foo: "C" })
   })
 
-  it("avoids showing stale data", async () => {
-    const { handleA, handleSlow, wrapper } = setup()
+  // Test document changes during loading
+  it("should handle document changes while loading", async () => {
```
msakrejda, line 252: Nice, good test idea.

packages/automerge-repo-react-hooks/test/useDocuments.test.tsx
```diff
@@ -1,250 +1,490 @@
...
 
-  it("returns a collection of loaded documents immediately, given a list of ids", async () => {
-    const { documentIds, wrapper } = setup()
-    const onDocs = vi.fn()
+    await act(async () => {
+      render(<Wrapped />, { wrapper })
+    })
+
+    await act(async () => {
+      await Promise.resolve()
```
msakrejda, line 78: Is this waiting for `useEffect`? It might be nice to have a comment.

packages/automerge-repo-react-hooks/test/useDocuments.test.tsx
```diff
@@ -1,250 +1,490 @@
...
-    // checking the number of listeners on the handle at the end.
-    const numMounts = 5 // arbitrary number here
-    for (let i = 0; i < numMounts; i++) {
-      const { unmount } = render(
-        <Component idsOrUrls={documentIds} onDocs={onDocs} />,
-        { wrapper }
-      )
-      await waitFor(() => unmount())
-    }
+  it("should handle multiple documents and parallel changes", async () => {
```
msakrejda, line 98: Nice.

packages/automerge-repo-react-hooks/test/useDocuments.test.tsx
msakrejda: Ah, this is waiting for the doc to load, right?

msakrejda (commented):
Nice, I'm excited to see this direction. I left a bunch of comments and questions, but overall it looks good. You had asked me privately about testing, but I don't see anything missing: test coverage looks great to me.

My main concern: in the React hooks, is it possible to load local documents _without_ suspending? That would be really nice. Is Suspense always triggered because `find` returns a promise? If that's the case, would a `findSync` work, that returns undefined if the handle's document is not loaded? I think that could be used to build hooks that don't suspend unless necessary.

packages/automerge-repo-react-hooks/src/useDocHandle.ts
```diff
@@ -0,0 +1,67 @@
...
+    controllerRef.current = new AbortController()
+
+    const promise = repo.find<T>(id, { signal: controllerRef.current.signal })
+    wrapper = wrapPromise(promise)
+    wrapperCache.set(id, wrapper)
+  }
+
+  useEffect(() => {
+    if (suspense === false) {
+      void wrapper.promise
```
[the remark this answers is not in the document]
msakrejda: Ah, this is to deal with floating promises? But you have a `then` and `catch` here, which should take care of that, no?

msakrejda:
Oh, also, does this approach to hooks work on React 19? I _think_ it should, but given the state of the Suspense APIs, it might be good to check.

examples/react-todo/src/main.tsx
```diff
@@ -38,7 +39,11 @@ window.repo = repo
 ReactDOM.createRoot(document.getElementById("root") as HTMLElement).render(
   <RepoContext.Provider value={repo}>
     <React.StrictMode>
-      <App url={docUrl} />
+      <ErrorBoundary fallback={<div>Something went wrong</div>}>
```
pvh, line 42: this officialish error boundary is chatty in the logs. i should silence it and/or replace it with one that is better for tests.

packages/automerge-repo-react-hooks/src/useDocHandle.ts
```diff
@@ -0,0 +1,67 @@
+import { AnyDocumentId, DocHandle } from "@automerge/automerge-repo/slim"
+import { wrapPromise } from "./wrapPromise.js"
+import { useRepo } from "./useRepo.js"
+import { useEffect, useRef, useState } from "react"
+
+// Shared with useDocHandles
+export const wrapperCache = new Map<
+  AnyDocumentId,
+  ReturnType<typeof wrapPromise>
+>()
```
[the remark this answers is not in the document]
pvh: oooooh. that might be better. it would also be a bit more complicated... my immediate instinct is to defer this until someone asks for it. the failure mode would be that if you had the same document in two different repos in your react app they'd share a cache of their state.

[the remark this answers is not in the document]
pvh: that would be hard to debug and understand for sure. should this be part of the repo context? i seem to remember in one version of this patch trying to rely on the consistency of the promises from repo.find() ... but then you can't resolve it in two places so i guess actually we need the cache here. hmmm.

i think i'd prefer to leave this as-is because the failure mode seems pretty remote and a robust solution seems complicated but i could be persuaded this is the wrong idea.

packages/automerge-repo-react-hooks/src/useDocHandle.ts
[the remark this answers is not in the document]
pvh: eslint was likely complaining. i'll see if i can remove it

packages/automerge-repo-react-hooks/src/useDocHandle.ts
```diff
@@ -0,0 +1,67 @@
...
+    controllerRef.current?.abort()
+    controllerRef.current = new AbortController()
+
+    const promise = repo.find<T>(id, { signal: controllerRef.current.signal })
+    wrapper = wrapPromise(promise)
+    wrapperCache.set(id, wrapper)
+  }
+
+  useEffect(() => {
+    if (suspense === false) {
```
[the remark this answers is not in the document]
pvh: i can fix that, yeah

packages/automerge-repo-react-hooks/src/useDocHandle.ts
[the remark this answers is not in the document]
pvh: my weak compromise was to add the comment below. if you find this too odious, i'm open to improving on it.

packages/automerge-repo-react-hooks/src/useDocHandles.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  // Check if we need any new wrappers
+  for (const id of ids) {
+    let wrapper = wrapperCache.get(id)!
+    if (!wrapper) {
+      try {
+        const promise = repo.find<T>(id)
+        wrapper = wrapPromise(promise)
+        wrapperCache.set(id, wrapper)
+      } catch (e) {
```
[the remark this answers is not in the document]
pvh: unavailable. this should probably have a test, now i think about it. i'm not sure what the behaviour would be without thinking hard which suggests they're probably wrong.

packages/automerge-repo-react-hooks/src/useDocHandles.ts
```diff
@@ -0,0 +1,75 @@
...
+              nextHandleMap.set(h.url, h)
+            }
+          })
+          setHandleMap(nextHandleMap)
+        }
+      )
+    } else {
+      setHandleMap(nextHandleMap)
+    }
+  }, [suspense, ids])
```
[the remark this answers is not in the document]
pvh: having that plugin would have saved me a couple stupid hours the other day, so i guess i will install it and fix the gripes!

packages/automerge-repo-react-hooks/src/useDocument.ts
```diff
@@ -1,63 +1,72 @@
...
+ * // Must be wrapped in Suspense boundary
+ * <Suspense fallback={<Loading />}>
+ *   <Counter />
+ * </Suspense>
+ * ```
+ */
+
+export interface UseDocumentParams {
+  suspense?: boolean
+}
```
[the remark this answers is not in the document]
pvh: it's used below. might be a stale comment.

packages/automerge-repo-react-hooks/src/useDocument.ts
```diff
@@ -1,63 +1,72 @@
...
+  const [doc, setDoc] = useState<Doc<T>>(() => handle.doc())
+  const [deleteError, setDeleteError] = useState<Error>()
 
-  // a state value we use to trigger a re-render
-  const [, setGeneration] = useState(0)
-  const rerender = () => setGeneration(v => v + 1)
+  // Reinitialize doc when handle changes
+  useEffect(() => {
+    setDoc(handle.doc())
+  }, [handle])
```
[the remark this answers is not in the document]
pvh: we could check reference equality before setting, do you think that would fix it?

packages/automerge-repo-svelte-store/package.json
```diff
@@ -1,13 +1,13 @@
...
   "version": "2.0.0-alpha.20",
-  "description": "A Svelte store containing your automerge documentsj",
+  "description": "A Svelte store containing your automerge documents",
   "repository": "https://github.com/automerge/automerge-repo/tree/master/packages/automerge-repo-svelte-store",
   "license": "MIT",
   "type": "module",
   "main": "dist/index.js",
   "scripts": {
-    "build": "tsc",
+    "build": "",
```
pvh, line 10: right, i forgot i need some svelte help

pvh: i think it's fine to ship an alpha but i need to either get help or learn some svelte i suppose

packages/automerge-repo-svelte-store/src/index.ts
```diff
@@ -87,7 +87,7 @@ export function setContextRepo(repo: Repo) {
 export function document<T>(documentId: AutomergeUrl, repo?: Repo) {
   repo = repo ?? getContextRepo()
   const handle = repo.find<T>(documentId)
```
pvh, line 89: this line here is the problem for svelte!

packages/automerge-repo/src/Repo.ts
```diff
@@ -397,7 +388,7 @@ export class Repo extends EventEmitter<RepoEvents> {
    * be notified of the newly created DocHandle.
    *
    * @throws if the cloned handle is not yet ready or if
-   * `clonedHandle.docSync()` returns `undefined` (i.e. the handle is unavailable).
+   * `clonedHandle.doc()` returns `undefined` (i.e. the handle is unavailable).
```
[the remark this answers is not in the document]
pvh: yep, so it should never throw unless you're doing something outside the API, so I'll remove the warning

packages/automerge-repo/src/Repo.ts
```diff
@@ -407,7 +398,7 @@ export class Repo extends EventEmitter<RepoEvents> {
       )
     }
 
-    const sourceDoc = clonedHandle.docSync()
+    const sourceDoc = clonedHandle.doc()
     if (!sourceDoc) {
```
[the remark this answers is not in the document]
pvh: fixed

packages/automerge-repo/test/DocHandle.test.ts
```diff
@@ -260,7 +259,7 @@ describe("DocHandle", () => {
     const handle = new DocHandle<TestDoc>(TEST_ID)
     assert.equal(handle.isReady(), false)
 
-    handle.doc()
+    handle.legacyAsyncDoc()
```
msakrejda, line 262: And here...

packages/automerge-repo/test/remoteHeads.test.ts
```diff
@@ -209,19 +218,21 @@ describe("DocHandle.remoteHeads", () => {
...
-          until: alice2DocA.whenReady(),
+          until: alice2DocAPromise,
         })
       ).filter(({ type }) => type === "remote-heads-changed")
 
+      const alice2DocA = await alice2DocAPromise
+
       // we should only be notified of the head changes of doc A
-      assert.strictEqual(remoteHeadsChangedMessages.length, 2)
+      assert.strictEqual(remoteHeadsChangedMessages.length, 1)
```
msakrejda, line 235: Why did this change?

msakrejda (approved):
Nice, I left some questions and comments, but broadly this looks great.

packages/automerge-repo-react-hooks/test/useDocument.test.tsx
```diff
@@ -110,7 +110,10 @@ describe("useDocument", () => {
     expect(onDoc).toHaveBeenCalledWith({ foo: "new value" })
   })
 
-  it("should throw error if the doc is deleted", async () => {
+  it.only("should throw error if the doc is deleted", async () => {
```
msakrejda, line 113: Stray `.only`

packages/automerge-repo/test/remoteHeads.test.ts
pvh: i don't know why. i decided i don't need to

packages/automerge-repo/src/helpers/tests/network-adapter-tests.ts
```diff
@@ -49,9 +49,10 @@ export function runNetworkAdapterTests(_setup: SetupFn, title?: string): void {
         // Alice creates a document
         const aliceHandle = aliceRepo.create<TestDoc>()
 
-        // Bob receives the document
-        await eventPromise(bobRepo, "document")
-        const bobHandle = bobRepo.find<TestDoc>(aliceHandle.url)
+        // TODO: ... let connections complete. this shouldn't be necessary.
+        await pause(50)
```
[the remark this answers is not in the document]
pvh: yeah, i needed it. i don't know why... investigating it is beyond my motivation right now.

PR #403, opened 2025-01-03
Support for heads in Automerge urls

pvh:
Long promised, finally implemented. Here's a patch that adds support for heads in URLs. They look a bit like this right now:

`automerge:2B7BMuPYXrj5wA9P2mjnKDtWhDwD#3h6NoQHYgGneicxivqye6vA4PqSA`

You can find, clone, and diff DocHandles that include their heads URLs. Changing documents identified by their heads is not supported (use changeAt on the main document handle without heads.)

Some design notes: the individual URL components are each bs58check encoded. This is wasteful but felt consistent. Designing a way to encode an array of heads into one bs58check encoded block was beyond my creativity today, and I felt like being able to see the individual components is worthwhile.

Internally, automerge uses hex-encoded hashes for heads. For consistency with documentIds (and compactness), I encode the *binary* form of the heads, not the string. Alas, this means I need to decode and re-encode them to hex strings every time I pass them into Automerge core. It's... fine, but it might be nice to support passing in the Uint8Array forms if it's not too much hassle.

I think before landing this (aside from any review comments) I ought to work up a little demo that shows time travel in action. If anyone else would like to do so, it would now be quite a bit easier! Contributions are welcome.

### DIFF
--- a/packages/automerge-repo/src/AutomergeUrl.ts
+++ b/packages/automerge-repo/src/AutomergeUrl.ts
@@ -5,11 +5,20 @@ import type {
   DocumentId,
   AnyDocumentId,
 } from "./types.js"
-import type { Heads } from "@automerge/automerge/slim"
+
 import * as Uuid from "uuid"
 import bs58check from "bs58check"
+import {
+  uint8ArrayFromHexString,
+  uint8ArrayToHexString,
+} from "./helpers/bufferFromHex.js"
+
+import type { Heads as AutomergeHeads } from "@automerge/automerge/slim"
 
 export const urlPrefix = "automerge:"
+
+// We need to define our own version of heads because the AutomergeHeads type is not bs58check encoded
+export type UrlHeads = string[] & { __automergeUrlHeadsBrand: unknown }
 
 interface ParsedAutomergeUrl {
   /** unencoded DocumentId */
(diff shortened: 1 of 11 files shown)


pvh:
Oh yeah, the type names. I did't love calling the type "UrlHeads" but couldn't think of anything better. I tried renaming Automerge.Heads to AutomergeHeads and just calling them Heads but that felt even more confusing.

Also, fun fact, because the Automerge.Heads is just type `string[]`, it's very easy to mishandle them. Oh well!

pvh:
A thought after spending a bit more time looking at improving useDocuments. I think we should probably add support for terminating a .find() via the AbortController API. I'm working on a revision to the patch for that.

alexjg (approved):
This looks good to me. As you mentioned in the discord, it would be nice to do away with the SHA256 check in bs58check and use something lighter now that we have multiple bs58checked things in the URL but that's a much bigger change.

pvh:
> This looks good to me. As you mentioned in the discord, it would be nice to do away with the SHA256 check in bs58check and use something lighter now that we have multiple bs58checked things in the URL but that's a much bigger change.

I could bs58encode the whole array with separators instead? I think this is fine for now and we can always revisit in the future.

PR #413, opened 2025-03-24
Export missing types from NetworkSubsystem

aidenfoxx:
I bumped into a situation where I needed the `PeerPayload` type for the "peer" network subsystem event.


pvh:
Thanks Aiiden, I'll include this in the next release.

PR #416, opened 2025-04-15
Clean up abort behavior to fix tests

pvh:


pvh (commented):
LGTM (but for real, thanks for finishing this)

pvh:
(Now you've written this patch twice, @msakrejda)

PR #427, opened 2025-05-20
fix: Fix return type of `DocHandle.changeAt` type

mykola-vrmchk:


alexjg:
Good spot!

pvh:
Thanks, @mykola-vrmchk!

PR #415, opened 2025-04-01
Let's make find synchronous again, why not?

pvh:
When we already have the document loaded, ideally we don't want to suspend. This patch is close to accomplishing that but has an unexplained exception leaking from abortable().

PR #429, opened 2025-05-27
automerge-repo-react-hooks: avoid Suspense where possible for useDocuments

msakrejda:
Allow useDocuments and useDocHandles to immediately return
already-loaded documents without invoking Suspense (in their Suspense
variants).

We should avoid unnecessary Suspense since it can lead to clumsy UX
with flashing loaders.

See also #415.


msakrejda:
I still need to add tests, but this fixes the problem for me locally.

msakrejda:
I think CI is not running because this is on top of #430 instead of against `main`. It should trigger once #430 lands.

msakrejda:
Hmm, well now CI is not running because I guess it does not run when the merge base is changed. Let me see if I can fix that (in a separate PR).

PR #430, opened 2025-05-27
Refactor automerge-repo-react-hooks test setup code

msakrejda:
In updating the tests for #429, I found that we have very similar
setup code in three hook test files. There are some benefits to
localized setup like that, but I can't imagine our tests for this
package will ever care very much about the actual document contents,
so instead of copying the paired repo setup code to yet another place,
I decided to try refactoring.

Thoughts?


msakrejda:
I plan to merge this in a day or two unless someone would like more time to review.

pvh:
That’s a nice catch, thanks!

PR #453, opened 2025-07-03
State of Dochandle should be "ready" when change event is triggered

paulsonnentag:
When you subscribe to changes on a doc handle the first time the document loads the state on the handle is still "loading" in  the change callback even though the document is already loaded.

```js
handle.on("change", () => {
   console.log(handle.state) // logs "loading" when initially loading instead of ready
})
```

Why is this happening? On each state transition we call `#checkForChanges` this function which has two responsibilities:

- checks if we need to emit a change event 
- transitions the state machine from "loading" to "ready" when the document was loaded for the first time

Right now when the document is loaded for the first time `#checkForChanges` emits a change event and then transitions the state machine from "loading" to "ready". That's why in the callback the state field on the handle is still loading. This pull request fixes this that we first transition into the ready state and then trigger a change event.

### DIFF
--- a/packages/automerge-repo/src/DocHandle.ts
+++ b/packages/automerge-repo/src/DocHandle.ts
@@ -227,8 +227,18 @@ export class DocHandle<T> extends EventEmitter<DocHandleEvents<T>> {
       encodeHeads(afterHeads),
       encodeHeads(beforeHeads)
     )
+
     if (docChanged) {
       this.emit("heads-changed", { handle: this, doc: after })
+
+      // If we didn't have the document yet, signal that we now do and return
+      // checkForChanges will then be triggered again by the state machine
+      // This ensures that when change is triggered the state is set to ready
+      if (!this.isReady()) {
+        console.log(this.state)
+        this.#machine.send({ type: DOC_READY })
+        return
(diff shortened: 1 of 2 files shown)


msakrejda (approved):
Thanks for the bug report and the fix! I have some minor comments, but overall it looks great.

msakrejda:
I'm pretty sure the CI failure is unrelated: https://github.com/automerge/automerge-repo/actions/runs/16046435722/job/45280996045?pr=453#step:6:797 -- it's an OOM. I'll try restarting the test and see if it's a recurring issue.

msakrejda:
Hmm, so, I think this does break some existing assumptions somewhere. I was able to reproduce the CI failure, partly: the tests appear to hang. I'm guessing maybe we're leaking memory somehow during the hang and eventually we OOM. I do see a slow leak locally. I can take a look at it next week.

darcyparker:
Looks superseded by #618 (Remove xstate, add DocumentQuery), which replaced the DocHandle state machine.
 
 Probably safe to close as resolved there.

msakrejda:
Ah, good call.

Issue #456, opened 2025-07-08
Sync server calls save multiple times on document create


georgewsu:
When calling `repo.create(doc)` from a client, the sync server calls save multiple times.

In Repo.ts, the issue is that `#registerHandleWithSubsystems` is called multiple times, but doesn't check for existing listeners before adding the save function:
`handle.on("heads-changed", throttle(saveFn, this.saveDebounceRate))`

eventemitter3 doesn't do any deduplication on listeners.

PR #457, opened 2025-07-08
Prevent duplicate save listeners

georgewsu:
Problem:
https://github.com/automerge/automerge-repo/issues/456

Proposed solution:
In Repo.ts, since `#registerHandleWithSubsystems` can be called multiple times, check existing listeners on the handle before adding a listener for the save function.

### DIFF
--- a/packages/automerge-repo/src/Repo.ts
+++ b/packages/automerge-repo/src/Repo.ts
@@ -279,13 +279,21 @@ export class Repo extends EventEmitter<RepoEvents> {
       }
       const throttledSaveFn = throttle(saveFn, this.saveDebounceRate)
       // Use a named function to make it easier to identify
-      const wrappedSaveFn = function wrappedSaveFn(payload: DocHandleEncodedChangePayload<any>) {
+      const wrappedSaveFn = function wrappedSaveFn(
+        payload: DocHandleEncodedChangePayload<any>
+      ) {
         throttledSaveFn(payload)
       }
 
       // Add save function as a listener if it's not already registered
       const existingListeners = handle.listeners("heads-changed")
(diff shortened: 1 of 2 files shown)


packages/automerge-repo/src/Repo.ts
```diff
@@ -277,7 +277,25 @@ export class Repo extends EventEmitter<RepoEvents> {
...
+      ) {
+        throttledSaveFn(payload)
+      }
+
+      // Add save function as a listener if it's not already registered
+      const existingListeners = handle.listeners("heads-changed")
+      if (
+        !existingListeners.some(
+          listener =>
+            listener.name === wrappedSaveFn.name &&
```
alexjg, line 293: Is it necessary to do this dance with the name or can we use `listener == wrappedSaveFn`?

georgewsu: Yeah, `listener == wrappedSaveFn` still doesn't work because they are different instances. Likewise, `Object.is(listener, wrappedSaveFn)` returns false (they each have a different prototype).

I originally just had this code check for `handle.listenerCount("heads-changed") === 0` but that wouldn't allow for multiple unique listeners if the code changes.

alexjg: Ah I see, because the functions are declared in the scope of a surrounding function. I feel like a cleaner solution would be to hoist those functions into module scope so that the same function is used everywhere and we can then use `==` for checking rather than the name checking.

georgewsu: Ah yes, good call! We can rely on the fact that `storageSubsystem` and `saveDebounceRate` are set once and never mutated after initial construction. Refactored accordingly, so that the save function is a single instance defined in the constructor (after `storageSubsystem` and `saveDebounceRate` are set).

PR #468, opened 2025-07-19
wip: add sedimentree

alexjg:
This PR is an attempt to find a minimal way to add  a new sedimentree based method of synchronizing documents to automerge-repo. The overall strategy is to add a second network subsystem which speaks sedimentree rather than speaking the automerge sync protocol. We are planning to implement the actual sedimentree logic in a separate rust-compiled-to-wasm package, so this PR is mostly wiring up a stubbed out version to automerge-repo internals. Specifically:

* Add an `Sedimentree` interface describing the interface we expect a sedimentree based network subsystem to implement
* Add a new `sedimentreeImplementation` and `sedimentreeAdapters` argument to the `RepoConfig` constructor. The idea behind this is that whilst we are developing it is quicker to inject the implementation into Repo than to import it internally - we don't have to keep rebuilding automerge-repo that way. The `sedimentreeAdapters` are the network adapters which are intended for use with the new sedimentree subsystem. `Repo` passes these adpaters to `sedimentreeImplementation.start` on boot.
* Split the "unavailalable" xstate machine into two parallel states tracking unavailability from the existing network adadpter and from the sedimentree network adapter
* If a `sedimentreeImplementation` is passed then use it to request `DocHandle`s.

### DIFF
--- a/packages/automerge-repo/src/DocHandle.ts
+++ b/packages/automerge-repo/src/DocHandle.ts
@@ -120,11 +120,41 @@ export class DocHandle<T> extends EventEmitter<DocHandleEvents<T>> {
           after: { [delay]: "unavailable" },
         },
         requesting: {
+          type: "parallel",
           on: {
-            DOC_UNAVAILABLE: "unavailable",
             DOC_READY: "ready",
           },
-          after: { [delay]: "unavailable" },
+          states: {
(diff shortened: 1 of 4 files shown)


expede:
Preeeeeetty sure that we can close this one

PR #471, opened 2025-07-24
Fix document saves made within 100ms being discarded

71:
`saveFn()` would only be called with the last set of parameters given to it, which would discard documents saved within 100ms of each other, except for one.

It's fairly trivial to reproduce:

```js
const a = repo.create()
const b = repo.create()
const c = repo.create()
```

Only `a` will be saved.

Fixing this could be approached in multiple ways. The mapping from key to throttled function could be done in `Repo.ts`, but after shortly working on it it seems it would have made things more complex. The current approach is quite simple IMO.

### DIFF
--- a/packages/automerge-repo/src/Repo.ts
+++ b/packages/automerge-repo/src/Repo.ts
@@ -18,7 +18,7 @@ import {
 } from "./DocHandle.js"
 import { RemoteHeadsSubscriptions } from "./RemoteHeadsSubscriptions.js"
 import { headsAreSame } from "./helpers/headsAreSame.js"
-import { throttle } from "./helpers/throttle.js"
+import { throttle, throttleByKey } from "./helpers/throttle.js"
 import {
   NetworkAdapterInterface,
   type PeerMetadata,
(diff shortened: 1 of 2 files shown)


packages/automerge-repo/src/Repo.ts
```diff
@@ -158,8 +158,10 @@ export class Repo extends EventEmitter<RepoEvents> {
       const saveFn = ({ handle, doc }: DocHandleEncodedChangePayload<any>) => {
         void this.storageSubsystem!.saveDoc(handle.documentId, doc)
       }
+      const keyFn = ({ handle }: DocHandleEncodedChangePayload<any>) => handle
```
71, line 161: This assumes that the `handle` object is stable per document ID. If that's not true, we can modify `throttleByKey` to use a regular map rather than a `WeakMap`, and use the document ID as a key.

msakrejda:
Thanks for reporting the issue, and thanks for the patch!

However, I wonder if we could instead handle this with per-document throttled save functions? Having document-specific throttling within a general save function feels awkward. That approach might be less performant, but I'm not sure if that matters here. Maybe it's worth benchmarking.

Thoughts?

packages/automerge-repo/src/Repo.ts
msakrejda: Looking at how handles are managed in this class, I think that's a safe assumption.

msakrejda:
No problem! Thanks for reporting the issue. I can add a test and finish the patch.

PR #476, opened 2025-07-31
Add the ability to export a bundle of documents from a Repo

alexjg:
Problem: there are many usecases where it would be useful to be able to transfer a set of documents from one repo to another without having a direct connection between them. Some important usecases are initializing an application with some shared document history and sync-via-email. At the moment this is not possible. Even though the underlying automerge document can be saved to a binary format, automerge-repo provides no APIs for importing such a binary file into an existing document without first finding that document somewhere.

The reason we have not added this import feature to date is that allowing the user to set the document ID for some arbitrary document is a giant footgun. It allows users to create documents which are unrelated to each other but have the same ID, which generally results in extremely surprising merge behavior.

Solution: add a new "bundle" concept. A bundle is a set of document IDs and their associated automerge documents serialized to a binary format. A bundle is created by calling `Repo.exportBundle(<list of DocHandles>)` and then imported into another repo by calling `Repo.importBundle`. This mechanism means that document IDs are still created by the repo - which means that there is no risk of a user accidentally creating two documents with the same ID.

Bundles can be encoded to a binary format using `Bundle.encode` and decoded using `Bundle.decode` - though this latter function is not necessary most of the time as you can also pass a `Uint8Array` directly to `Repo.importBundle`. The binary format is intended to be forwards compatible so that bundles can be kept around for a long time.

### DIFF
--- a/packages/automerge-repo-bundles/README.md
+++ b/packages/automerge-repo-bundles/README.md
@@ -1,0 +1,48 @@
+# `automerge-repo-bundles`
+
+A library for exporting and importing "bundles" of documents from
+`automerge-repo`
+
+## Why?
+
+There are two important usecases for this package:
+
+1. Exporting some kind of "initial" document state for users to collaborate on
+2. Exporting documents for asynchronous collaboration (e.g. email)
+
+`automerge-repo` doesn't provide any way to obtain a `DocHandle` without having
+some network connection to a peer who has that document. This makes asynchronous
+operations complicated. For example, if I want to write an application which
+enables people to collaborate over email, what do I actually give my users to
+send to each other?
+
+`automerge-repo` does expose an `Repo.import` method, which can be used to
+import a raw automerge document and you can pass a document ID to this method.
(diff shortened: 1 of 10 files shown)


packages/automerge-repo/src/bundle.ts
```diff
@@ -0,0 +1,348 @@
...
+
+    docHeaders.push({ docId, heads, deps, dataLen, bundleOffset })
+  }
+
+  // Read document data
+  const docs = new Map<DocumentId, DocBundle>()
+  const dataStart = offset
+
+  for (const header of docHeaders) {
+    const docData = data.slice(
```
dmaretskyi, line 327: consider using [TypedArray.subarray](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/subarray) to avoid memcopy

dmaretskyi (approved):
This is exactly what we need, thanks!

PR #483, opened 2025-08-13
fix: Fix imported document is not being lazy replicated

mykola-vrmchk:


alexjg:
This looks like an entirely excellent fix. Would you mind also adding a test?

msakrejda:
I pushed a test fix (I _think_ that's the right change) and a minor lint fix. This should be good to go now.

PR #484, opened 2025-08-15
Avoid re-evaluating useDocHandles hook unnecessarily

### LINKED ISSUE #479: [BUG] - useDocuments causes infinite re-renders
I'm encountering an issue where the useDocuments hook in automerge-repo triggers an infinite render loop in React.

🔍 Context:
When I subscribe to documents using:

```tsx
const docs = useDocuments(docUrls);
```

The component keeps re-rendering indefinitely. It appears that the return value from useDocuments might be unstable or causing state updates even when nothing has changed.

✅ Expected behavior:
The hook should only trigger a re-render when the actual content of the documents changes - not on every render cycle.

🧪 Environment:
```
  {
    "dependencies": {
     "@automerge/react": "^2.2.0",
     "@automerge/automerge-repo": "^2.2.0",
     "@automerge/automerge-repo-storage-indexeddb": "^2.2.0"
    }
  }
```

- React version: ^19.1.1
- Framework: Vite
- Browser: Chrome

msakrejda:
We're currently updating the hook's handle map whenever the list of ids changes, causing it to trigger a re-render. This is intended, but we are passing the ids Array directly in as a useEffect dependency, meaning that any usage of the hook with a literal array, like

  useDocHandles([id1, id2, id3])

will create a new array on each render, re-triggering the useEffect hook, causing a setState and another render, indefinitely.

This code could use some refactoring (it's not obeying React's rules of hooks), but for now, work around the issue by adding a useSet hook that gives a stable reference for the useEffect dependency array as long as the contents of the id list passed to useDocHandles don't change.

Fixes #479.

### DIFF
--- a/packages/automerge-repo-react-hooks/src/helpers/useSet.ts
+++ b/packages/automerge-repo-react-hooks/src/helpers/useSet.ts
@@ -1,0 +1,19 @@
+import { useEffect, useState } from "react"
+
+export function useSet<T>(items: T[]): Set<T> {
+  const [set, setSet] = useState<Set<T>>(() => {
+    return new Set<T>(items)
+  })
+  useEffect(() => {
+    const newSet = new Set(items)
+    if (identical(set, newSet)) {
+      return
+    }
+    setSet(newSet)
+  }, [set, items])
+  return set
+}
+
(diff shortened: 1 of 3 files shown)


msakrejda:
Still needs testing.

PR #487, opened 2025-08-20
Fix useDocHandle unnecessary re-renders

simleb:
With `suspense: false`, `useDocHandle` would thrash the state by calling `setHandle` even when the `currentHandle` was already available.

In some project, this lead to a dramatic performance degradation as many components re-rendered unnecessarily.

The rules of hooks force us to place hooks before early returns so we just need to short-circuit inside the `useEffect`, like we already do for suspense.

I've verified that the new test fails before the change (and passes after).

### DIFF
--- a/packages/automerge-repo-react-hooks/src/useDocHandle.ts
+++ b/packages/automerge-repo-react-hooks/src/useDocHandle.ts
@@ -65,7 +65,7 @@ export function useDocHandle<T>(
    * re-running this function until it succeeds, whereas the synchronous
    * form uses a setState to track the value. */
   useEffect(() => {
-    if (suspense || !wrapper) {
+    if (currentHandle || suspense || !wrapper) {
       return
     }
     wrapper.promise
--- a/packages/automerge-repo-react-hooks/test/useDocHandle.test.tsx
+++ b/packages/automerge-repo-react-hooks/test/useDocHandle.test.tsx
@@ -1,4 +1,8 @@
-import React, { Suspense } from "react"
+/**
+ * Copyright (c) 2025 - Present. Blend Labs, Inc. All rights reserved
+ * Blend Confidential - Restricted
+ */
+import React, { act, Suspense } from "react"
 import {
   AutomergeUrl,
   DocHandle,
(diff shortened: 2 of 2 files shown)


packages/automerge-repo-react-hooks/test/useDocHandle.test.tsx
```diff
@@ -281,5 +282,34 @@ describe("useDocHandle", () => {
...
+        <NonSuspenseComponent url={handleA.url} onHandle={onHandle} />,
+        { wrapper }
+      )
+
+      expect(onHandle).toHaveBeenCalledTimes(1)
+
+      rerender(<NonSuspenseComponent url={handleA.url} onHandle={onHandle} />)
+      await act(pause)
+
+      expect(onHandle).toHaveBeenCalledTimes(2)
```
simleb, line 312: Without the change, this test fails because `onHandle` is called 3 times.

I don't love having to use a pause but it's necessary for the extra re-render to happen.

msakrejda: I think the pause is okay, I can't think of a great way to test this either. But this could maybe use a comment in the code (noting the desired behavior), rather than just here on the PR.

Also, it looks like if I move the pause to line 306 and drop the rerender, the test fails on line 307 without the fix (with two renders instead of one). Do we need the rerender?

msakrejda (commented):
Nice, thanks for the fix. A couple of minor comments, but looks great.

packages/automerge-repo-react-hooks/test/useDocHandle.test.tsx
simleb: Good point, I'll try to simplify.

packages/automerge-repo-react-hooks/src/useDocHandle.ts
```diff
@@ -65,7 +65,7 @@ export function useDocHandle<T>(
    * re-running this function until it succeeds, whereas the synchronous
    * form uses a setState to track the value. */
   useEffect(() => {
-    if (suspense || !wrapper) {
+    if (currentHandle || suspense || !wrapper) {
```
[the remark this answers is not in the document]
simleb: Arg, I'm so used to having the linter have my back that I forgot.

packages/automerge-repo-react-hooks/test/useDocHandle.test.tsx
simleb: I ended up keeping the second render to hammer the point but I did add a pause after the initial render and added some comments. Let me know if that works.

simleb:
Also let me know if I should rebase into a single clean commit of if you squash and merge.

Issue #488, opened 2025-08-21
Possible bug when initializing a doc with an empty object


sergeuz:
I encountered what looks like a bug in `automerge-repo` when upgrading my project to Automerge 3: when the initial contents of a new doc created by one peer is an empty object, other peers fail to synchronize it:

```ts
doc = repo.create({}); // Doesn't work
doc = repo.create({ a: 1 }); // Works
doc = repo.create(undefined); // Works
```

The problem manifests itself when a peer that needs the affected doc calls `repo.find()` with its document ID (shared out of band) for the first time – the call never resolves, supposedly because the doc doesn't transition to a ready or unavailable state in time. I'm using automerge-repo 2.2.0 and the problem is not reproducible with automerge 1.2.1.

I don't have a minimal example as of right now that would reproduce this as the project uses custom network adapters, just figured this might be something on the radar already or can be easily tested/debugged on your side.

svdvonde:
I am also facing this issue (see referred issue #538, which I closed since it's a duplicate of this). The behaviour that empty documents do not sync is weird and unexpected.

pawelsmolaga:
Hi @sergeuz, @svdvonde this issue was addressed in this pull request https://github.com/automerge/automerge-repo/pull/631

CC: @alexjg @msakrejda please close this issue

PR #489, opened 2025-08-22
WIP: Fix synchronization of empty documents

msakrejda:
As a first step, add a reproduction of the issue reported in #488.

This test is not in the right place, but I'm not sure where the
problem is yet.


msakrejda:
I tracked this down to [this line](https://github.com/automerge/automerge/blob/01590990918e55009cd7ffd0c4334844322f1a2f/javascript/src/implementation.ts#L1097) in the core automerge package--it returns a null `syncMesssage` when the doc is empty. I tried to follow through to the Rust code to understand why, but I got lost.

darcyparker:
See https://github.com/automerge/automerge-repo/issues/488#issuecomment-4647074972 This was probably addressed in #631

PR #493, opened 2025-09-10
Add option to choose doc IDs in Repo.create

alexjg:


jtfmumm (approved):

alexjg:
Yeah I don't like it either but this is the minimal API we need to allow keyhive to manage document IDs. One mitigation is that the factory function doesn't get access to the document contents and it's not an argument to `create` so it's not overridable on an ad hoc basis. I hope that this would make it tougher to blow your foot off.

Really though, this is a hidden API which is here just to expedite keyhive integration and we should replace it with something better once we have it working. Fundamentally we need to unforgeably associate each Automerge document with a keyhive ID, other ways to do this would be more complex I think but I would be very very open to suggestions.

alexjg:
I think there is an argument here that this functionality should be on a `keyhive` branch and release until we've figured it out.

PR #496, opened 2025-09-12
Make state management more explicit

alexjg:
Currently the lifecycle of a document is a little tricky to follow. We have an xstate state machine in `DocHandle`, but the events which modify that state machine are emitted by various asynchronous processes which have a reference to the document scattered throughout the system. This makes it hard to answer questions like "why is this document unavailable" because we don't have a single place in the codebase where we have all the information used to make that decision at the same time.

This commit is an attempt to reorganise state management to make these kinds of questions easier to answer. Roughly speaking it looks like this:

* Replace the `CollectionSynchronizer` and `DocSynchronizer` classes with a single `DocumentPhasor` class, which manages the lifecycle of a single document.
* Move all the state logic out of `DocHandle`. `DocHandle`s are now just a place for users to register interest in a document and get notified of changes. All actual state changes are routed through the `DocumentPhasor`

The "DocumentPhasor" state machine manages a set of `Phase`s, each of which represents different stages of the document lifecycle ("loading", "requesting", "ready", "unavailable"). Modifications to the phasor are achieved by passing events to the `DocumentPhasor.handleEvent` method, which returns an object describing what has changed.

This PR is still in-progress because

a) I think the state machine logic is pretty verbose and can probably be made more succinct
b) I need to write up a few things which are not covered by tests relating to the metrics we use for monitoring the sync server
c) I need to go through in detail looking for things which are not covered by tests in the original implementation as this is a large change.


msakrejda (commented):
I didn't have time to look at the full thing this morning, but I made it through most of DocumentPhasor. In general, I really like this approach: I think it achieves the goals you laid out in the overview. I left some nitpicks and a bunch of questions, and I think I need to read it over another couple of times to fully understand it, but this is a great direction.

packages/automerge-repo/src/DocumentPhasor.ts
```diff
@@ -0,0 +1,829 @@
...
+// it is necessary to load the document from storage. When an event returns
+// `DocEffects.beginLoad: true` then the caller should load the document. When
+// the load completes the caller should provide the result of the load via the
+// "load_complete" event.
+//
+// ### Network Ready
+//
+// Initially a phasor is in a "network not ready" state. This means that it will not mark
+// documents as unavailable until the "network_ready" event has been received. Callers
+// should dispatch this event as soon as the network subsystem says it is ready.
```
msakrejda, line 64: I don't think this refactor should change the state machine, but in the long term, it would be nice to support the network going in and out of the "ready" state. Have you considered how that fits in here?

msakrejda (commented):
I took another look at this, focusing on the phases. I still think it broadly makes sense, though the interface between `DocumentPhasor` and the phases seems pretty intricate. Maybe that's necessary complexity, but I wonder if we can avoid some of it. E.g., `loadRunning` seems like a quirk.

I also feel like I'm not quite clear on the sync protocol: are there docs on that?

alexjg:
superceded by #618

PR #508, opened 2025-09-30
feat: add exponential backoff to WebSocketClientAdapter

whythecode:


msakrejda:
Thanks for the contribution! Can you say more about the motivation for this change? In general, exponential back-off is a good idea, but the 5s retry period we have seems unlikely to be a problem. Are you running into an actual issue?

whythecode:
no issue, just seemed like an easy improvement to make - for me personally it was a bit annoying when testing intermittent connections between peers, but I can live without it

PR #528, opened 2025-11-15
Rework awareness

msakrejda:
This aims to move the awareness/presence functionality out of the
useLocalAwareness/useRemoteAwareness React hooks and make it available
as a core part of automerge-repo. The hooks can then be rewritten to
be simple wrappers around the lower-level code, and users of other
frameworks (or no frameworks at all) can benefit as well.

This approach introduces an Awareness EventEmitter class that adds
some structure around ephemeral-message events on a doc handle. It
also allows checking state by peerId or by caller-supplied userId.

It does not attempt to address the existing issue that userIds can be
spoofed.

### DIFF
--- a/packages/automerge-repo/src/Presence.ts
+++ b/packages/automerge-repo/src/Presence.ts
@@ -154,6 +154,7 @@ export class Presence<
     if (this.#running) {
       return
     }
+    this.#running = true
 
     this.#heartbeatMs = heartbeatMs ?? DEFAULT_HEARTBEAT_INTERVAL_MS
     this.#peers = new PeerPresenceInfo(peerTtlMs ?? DEFAULT_PEER_TTL_MS)
@@ -239,7 +240,6 @@ export class Presence<
 
     this.broadcastLocalState() // also starts heartbeats
     this.startPruningPeers()
-    this.#running = true
   }
 
   /**
@@ -347,6 +347,9 @@ export class Presence<
     type: PresenceMessageType,
     extra?: Record<string, unknown>
   ) {
+    if (!this.#running) {
+      return
+    }
     this.#handle.broadcast({
       [PRESENCE_MESSAGE_MARKER]: {
         userId: this.userId,
--- a/packages/automerge-repo/test/Presence.test.ts
+++ b/packages/automerge-repo/test/Presence.test.ts
@@ -10,14 +10,16 @@ import { wait } from "./helpers/wait.js"
 type PresenceState = { position: number }
 
 describe("Presence", () => {
-  async function setup() {
+  async function setup(opts?: { skipAnnounce?: boolean }) {
     const alice = new Repo({ peerId: "alice" as PeerId })
     const bob = new Repo({ peerId: "bob" as PeerId })
     const [aliceToBob, bobToAlice] = DummyNetworkAdapter.createConnectedPair()
     alice.networkSubsystem.addNetworkAdapter(aliceToBob)
     bob.networkSubsystem.addNetworkAdapter(bobToAlice)
-    aliceToBob.peerCandidate("bob" as PeerId)
-    bobToAlice.peerCandidate("alice" as PeerId)
+    if (!opts?.skipAnnounce) {
+      aliceToBob.peerCandidate("bob" as PeerId)
+      bobToAlice.peerCandidate("alice" as PeerId)
+    }
     await Promise.all([
       alice.networkSubsystem.whenReady(),
       bob.networkSubsystem.whenReady(),
@@ -44,17 +46,19 @@ describe("Presence", () => {
         repo: alice,
         handle: aliceHandle,
         presence: alicePresence,
+        network: aliceToBob,
       },
       bob: {
         repo: bob,
         handle: bobHandle,
         presence: bobPresence,
+        network: bobToAlice,
       },
     }
   }
 
   describe("start", () => {
-    it("shares initial state", async () => {
+    it("activates presence and shares initial state", async () => {
       const { alice, bob } = await setup()
 
       alice.presence.start({
@@ -62,12 +66,14 @@ describe("Presence", () => {
           position: 123,
         },
       })
+      expect(alice.presence.running).toBe(true)
 
       bob.presence.start({
         initialState: {
           position: 456,
         },
       })
+      expect(bob.presence.running).toBe(true)
 
       await waitFor(() => {
         const bobPeerStates = bob.presence.getPeerStates()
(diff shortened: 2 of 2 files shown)


OrionReed:
This is good. In fact, this afternoon I spent a couple hours implementing almost the exact same design with the same structure, heartbeat system, and events just with slightly different names. So I take that as a good sign.

The only difference with the thing I tried is that I added a 'direct message' system which orders peer ID pairs, hashes them, then bs58check encodes them to get a unique message channel between 2 peers (as handshakes are pairwise and a non-trivial amount of data, sending these to everyone felt like something worth avoiding)

I wonder what your thoughts are there. But as for the overall API design I like it. I'll pull this into my experiment and see how it feels. I think for React this is good, @chee can speak more to solid-js.

I'm also personally in favor of 'presence' over 'awareness' and I think it is a bit more widely used.

msakrejda:
@OrionReed excellent, thanks for the feedback, and glad we've arrived at more or less the same design. No preference on "presence" versus "awareness" as terminology, but it probably makes sense to go with common usage.

The DM idea also makes sense. Is that still going through handle ephemeral messages? That still broadcasts it to all peers and they just ignore everything? DocSynchronizer doesn't seem to have a public way to send a message to a single peer. Or should we add that to support this use case?

OrionReed:
@msakrejda my approach was to make a new doc that only 2 peers share (a little hacky, making a new automerge URL based on the pair of peer IDs)

I think a better approach probably would be to use the same ephemeral messaging as the rest of the system, and leave more direct connections to a future performance optimisation (Keyhive is also a consideration here, so feels reasonable to punt)

I'm using your system right now, happy to chat tomorrow or Monday and let you know how it went and anything I learn along the way.

OrionReed:
- Seems like dispose() should broadcast a goodbye?
- Should there be a heartbeat timeout to remove peers if they've been gone a long time? I could see the case for leaving them forever if certain apps want that. In my current experiment I want a mostly 'live' peer list. perhaps this should be configurable?

msakrejda:
> - Seems like dispose() should broadcast a goodbye?

Yeah, good point. Will update.

> - Should there be a heartbeat timeout to remove peers if they've been gone a long time?

We may want to leave that as a higher-level concern, but I could also see taking an `idle_timeout` option to prune these. I'll explore that.

OrionReed:
My takes after a few days of use:
- call it 'Presence' as a slightly more idiomatic name.
- consider [private elements](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_elements) instead of TS private
- add some docstrings (including mention of events used to say goodbye when closing the page on the .dispose() method)
- add the missing `goodbye` broadcast on dispose
- personally, I think an idle_timeout is worthwhile (and I think will save at least a few people from doing it wrong in React). I'm not sure if the default should be a timeout (say, 3x the heartbeat rate) or `null`, but I lean towards a default timeout and let people disable it if they really want to. I think that's what users will expect.

Aside from that this all looks nice!

examples/react-use-awareness/src/App.jsx
```diff
@@ -1,30 +1,21 @@
-import {
-  useDocument,
-  useLocalAwareness,
-  useRemoteAwareness,
-} from "@automerge/react"
+import { useDocument, usePresence } from "@automerge/react"
```
msakrejda: Note: I combined the two hooks into one, since they need to communicate (local awareness needs to broadcast when a new peer is seen). The current code does that with a shared global, but that has a bunch of shortcomings: e.g., it could lead to unnecessary messages if working with presence on two unrelated handles (or even repos). We could use React Context instead, but I don't think that's better.

packages/automerge-repo/src/Presence.ts
```diff
@@ -0,0 +1,596 @@
...
+    this.#userId = userId
+    this.#deviceId = deviceId
+
+    this.#localState = initialState
+    this.#peers = new PeerPresenceInfo(opts?.peerTtlMs ?? DEFAULT_PEER_TTL_MS)
+
+    if (opts?.skipAutoInit) {
+      return
+    }
+    this.initialize()
```
msakrejda: I had a really hard time getting React to play nice with Presence's non-pure constructor, especially when in Strict mode. I ended up moving initialization out to a separate method. I don't love that this complicates the interface, but otherwise, Presence initialization in React needs to happen in a `useEffect`, so we'd need to call the constructor there. Since `useEffect` happens _after_ a render, that means the initial `usePresence` render would return `undefined`, screwing up types and making the interface more complicated. 

If this is totally unpalatable, we can instead use a wrapper to initially delegate to a dummy Presence stub and then to a real Presence once initialized, but that complicates the React integration quite a bit (plus any other system with similar constraints). Alternately, we could drop the `skipAutoInit` option and always require a manual call to `initialize`. I don't have strong feelings on that. I like that it avoids side effects in the constructor, but it is a more verbose API.

packages/automerge-repo/src/Presence.ts
```diff
@@ -0,0 +1,592 @@
...
+  private stopHeartbeats() {
+    clearInterval(this.#heartbeatInterval)
+  }
+}
+
+/**
+ * A summary of the latest Presence information for the set of peers who have
+ * reported a Presence status to us.
+ */
+export class PeerPresenceView<State> {
```
msakrejda: I'm exposing the current peer Presence state through this class, since (especially with users and devices) some things would be awkward to do with just a raw data structure.

packages/automerge-repo/src/Presence.ts
```diff
@@ -0,0 +1,592 @@
...
+    const peer = this.getFreshestPeer(peers)
+    if (!peer) {
+      return undefined
+    }
+
+    return this.getPeerState(peer, channel)
+  }
+}
+
+class PeerPresenceInfo<State> extends EventEmitter<PresenceEvents> {
```
msakrejda: This class is used internally by Presence to manage the state that's exposed to users through `PeerPresenceView` above.

msakrejda:
I addressed some of the feedback. I also decided to drop the automatic initialization (for now: I'm not pushing for that as a final mechanism, it just makes it easier to explore some API options), and renamed `initialize` / `dispose` to `start` / `stop` and made sure they are idempotent. I think this is easier to reason about.

I also attempted to implement my idea above re: DocHandle owning Presence to avoid conflicts, but I ran into an issue: then DocHandle either needs to know about all the things the Presence constructor needs: either in its constructor (and anywhere that's called, like `Repo#getHandle()` and all its callers) or a `getPresence()` method. The former is clearly not a good idea. The latter means that users could pass a different `userId` each time they call `getPresence()` even though the intent is to have that return a persistent Presence object that lives as a member property of the DocHandle. Do we allow those changing parameters? Ignore it? Reject it?

The other thing I'm considering is moving most of those parameters from the Presence constructor to the `start()` method. That still maintains the dubious feature of changing `userId` and `deviceId` in the middle of a session, but otherwise seems generally simpler.

msakrejda:
>The other thing I'm considering is moving most of those parameters from the Presence constructor to the start() method. That still maintains the dubious feature of changing userId and deviceId in the middle of a session, but otherwise seems generally simpler.

Concretely, as sketched out in 42b5803249fbf7c957001ce5b3644f2309c113b3. The downside here is that users need to supply the presence state type parameter manually when getting the Presence instance from the DocHandle. I don't see a great way around that. I'll sleep on this some more and see if another approach springs to mind.

alexjg:
I think we definitely shouldn't own the ephemeral message channel. We don't have much control over what people are sending (messages are gossiped quite prolifically). We shouldn't break on unrecognized messages, but I think we should also make an effort to let people use the ephemeral channel without us stepping on our toes. The approach that makes sense to me there is to stick everything we are sending inside a top level key in the message, something like `handle.broadcast({__presence: { .. }})`. That way other code sharing the dochandle can follow similar conventions in order to interoperate.

PR #530, opened 2025-11-25
Fix formatting issues in README.md

2400032286:
Added missing periods to Storage adapters and Network adapters sections for consistent punctuation throughout the documentation.


msakrejda (commented):
Thanks for the contribution, but these are not full sentences, so they don't really need periods.

PR #533, opened 2025-12-02
Add end-to-end testing with Playwright

msakrejda:
This fixes the react-use-awareness example bug I found in the course
of #528 and adds a Playwright test to have some end-to-end
coverage. We can add other examples and
https://github.com/automerge/automerge-repo-quickstart as follow-ups
if this looks good.


.github/workflows/playwright.yaml
```diff
@@ -0,0 +1,49 @@
...
+        with:
+          fetch-depth: 0
+          ref: ${{ github.ref }}
+      - uses: pnpm/action-setup@v4
+        with:
+          run_install: false
+      - name: Install and Build
+        run: |
+          pnpm install
+          pnpm build
```
msakrejda: I copied this from the neighboring `tests.yaml`. I'm not sure if there's a better way to share this setup.

msakrejda:
I think that last test failure was unrelated. I'll try restarting.

PR #540, opened 2025-12-09
Presence API tweaks

msakrejda:
In trying to integrate the new Presence APIs, I found it's tricky to
handle changes to peer states in React because our PeerPresenceView
class retains its object identity and is just mutated for
updates. Because React relies on object identity for things like
useEffect, this can be awkward to work with.

Instead, use a plain JS object to manage peer information, and add
some helper functions to access it.


msakrejda:
Updated based on discussion with pvh. I ended up wrapping the peer state in a class that provides the helpers as member functions. This is not especially elegant, but it maintains the distinct-object-identity-for-distinct-states property we want for React and gives a reasonably convenient place to put the helpers. I'll review the helpers again before merging: right now accessing the presence state can involve a potentially awkward mix of helper calls and accessing the raw state `.value`. It'd be nice to make that more consistent.

PR #549, opened 2026-01-17
Add usePresence for solid primitives

matfire:
- Adds a usePresence hook to the solid-primitives package (mimics the one present in the react-hooks package)
- Exposes start/stop methods from the underlying Presence instance

### DIFF
--- a/packages/automerge-repo-solid-primitives/src/usePresence.ts
+++ b/packages/automerge-repo-solid-primitives/src/usePresence.ts
@@ -23,6 +23,8 @@ export type UsePresenceResult<State extends PresenceState> = {
     channel: Channel,
     value: State[Channel]
   ) => void
+  start: () => void
+  stop: () => void
 }
 
 export function usePresence<State extends PresenceState>({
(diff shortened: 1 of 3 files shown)


neftaly:
Can I trivially switch between fast and slow heartbeat rate, while the app is running? If this can be done just by changing a prop or arg then it should be good. I am very unfamiliar with these codebases :)

msakrejda:
> hey @msakrejda, quick question: when you mention the update interval you're talking about the heartbeatMs attribute in the Presence class, right? Or am I missing something?

Sorry I missed that: yeah, just the heartbeat interval.

neftaly:
A test for the switching scenario would be really nice :) I'll make an issue

matfire:
Got it!

Since there's a separate issue, should the tests be in a separate pr or (at least for solid) is it fine to add it in here ?

matfire:
added a few more tests to also set a custom heartbeat on "mount"

packages/automerge-repo-solid-primitives/src/usePresence.ts
```diff
@@ -0,0 +1,109 @@
...
+  deviceId,
+  initialState,
+  heartbeatMs,
+  peerTtlMs,
+}: UsePresenceConfig<State>): UsePresenceResult<State> {
+  const [localState, setLocalState] = createSignal<State>(initialState)
+  const [peerStates, setPeerStates] = createSignal(new PeerStateView<State>({}))
+  const [presence] = createSignal(
+    new Presence<State>({ handle, userId, deviceId })
+  )
```
chee: i don't think we get anything from this being a signal, as it isn't tracking and doesn't change.

msakrejda:
Thanks for the update, @matfire. @chee (or anyone else), any other feedback?

msakrejda:
Any reason not to merge this?

PR #586, opened 2026-04-01
Fix `subductionjs` infinite loop

expede:
PvH reported an infinite loop edge case and very helpfully provided the minimal logs needed to diagnose! 🙏


packages/automerge-repo/src/subduction/source.ts
```diff
@@ -575,8 +570,7 @@ export class SubductionSource implements DocumentSource {
         `[subduction] #save ${sid}: setting needsResync=true (sync in flight)`
       )
       entry.needsResync = true
-    } else {
-      console.log(`[subduction] #save ${sid}: setting lastSyncResult=null`)
+    } else if (entry.lastSyncResult !== "no-peers") {
```
expede, line 575: The actual fix

PR #593, opened 2026-04-10
Wire up SubductionSource to handle unknown docs and retry failed syncs.

jtfmumm:


packages/automerge-repo/src/subduction/source.ts
```diff
@@ -98,6 +98,12 @@ export interface SubductionSourceOptions {
   onEphemeral?: OnEphemeral
   onHealExhausted?: (documentId: DocumentId) => void
 
+  /**
+   * Called when subduction receives data for a document it doesn't have
+   * an entry for. The Repo creates the query/handle and calls attach.
+   */
+  ensureHandle?: (documentId: DocumentId) => void
```
alexjg: Is there any reason for this to be optional?

alexjg (commented):
Can we add some tests for this functionality? There should be sufficient machinery in the existing subduction integration tests.

packages/automerge-repo/src/subduction/source.ts
jtfmumm: Good point. Updated.

jtfmumm:
> Can we add some tests for this functionality? There should be sufficient machinery in the existing subduction integration tests.

Added

packages/automerge-repo/src/subduction/source.ts
```diff
@@ -98,6 +98,12 @@ export interface SubductionSourceOptions {
   onEphemeral?: OnEphemeral
   onHealExhausted?: (documentId: DocumentId) => void
 
+  /**
+   * Called when subduction receives data for a document it doesn't have
+   * an entry for. The Repo creates the query/handle and calls attach.
+   */
+  ensureHandle: (documentId: DocumentId) => void
```
expede, line 105: Minor naming nit: the convention that I generally see for `ensure*` is that it's something that does extra work to ensure that you acquire some resource, _or fails_. I think the intention here is more for "I heard about a thing that I wasn't aware of before, please optionally add it to the store". Have you considered `receiveDocument`, `reactiveHandle`, `upsertHandle`, or similar?

expede: Alternately, if there's more general use in the `ensure*` pattern, we could make the comment reflect that generality of use case. Maybe something like:

>  Idempotently ensure the Repo has a handle for the given document. Called when we discover a document that doesn't have an entry for — either via incoming data or a newly persisted ID. The Repo can then create the query/handle and calls attach if needed.

expede: Also I could be misunderstanding the code, but in the current state, can this be used to (adversarially) force remote nodes to load docs into memory rather than just what they've asked for? The current setup for new docs in e.g. `react-todo` is to have a top level doc that tracks subdocuments and lets the application decide what to load.

expede: Ran it past Opus

---

Yes, this is a real concern. Let me lay out the attack surface this PR introduces.

Pre-PR behavior
- #handleDataFound: if data arrives for a sedimentree with no entry in #entries, it returns early — if (!entry) return. The data is persisted to storage by the Wasm layer, but no handle is created and nothing is loaded into memory.
- sedimentree-id-saved: this event didn't exist.
Post-PR behavior
Both new paths unconditionally call ensureHandle, which creates a DocumentQuery + DocHandle, attaches all sources, and loads data into memory:
Remote peer pushes commits for doc X
  → Wasm engine: authorizePut(requestor, author, sedimentreeId)
  → [if allowed] saveSedimentreeId → emit("sedimentree-id-saved")
  → #handleSedimentreeIdSaved → ensureHandle(docX)
  → Repo creates query, handle, attaches all sources, loads blobs
The issue
If authorizePut allows the data (which it does by default — no policy = allow-all), a remote peer can push commits for arbitrary document IDs. Each one triggers:
1. A DocumentQuery + DocHandle allocation in Repo.#queries
2. source.attach(query) on every registered DocumentSource
3. Blob loading into an Automerge doc in memory
There is no:
- Rate limiting on handle creation
- Cap on the number of handles
- Lazy/eviction strategy for handles created this way
- Distinction between "I asked for this doc" vs "a peer pushed this to me unsolicited"

A malicious (or buggy) peer could push thousands of small documents, each causing a handle + Automerge doc to be allocated. This is a memory exhaustion vector. [BZ: and a replication vector for malicious content]

expede: That said, I _have_ been seeing handles that are `undefined` in the wild which I'm told should not be possible, so you've 100% identified a real thing that needs fixing. I think we should think slight more on this one Monday though

PR #594, opened 2026-04-12
Retry failed syncs on SubductionSource on shareConfigChanged()

jtfmumm:
This PR adds `retryFailedSyncs()` to `SubductionSource`, which resets all "all-failed" entries to null and clears their heal state, triggering an immediate re-sync on the next `#recompute`. `Repo.shareConfigChanged()` now calls it, so subduction policy updates (e.g., because of a new keyhive delegation) lead to successful syncing.

### DIFF
--- a/packages/automerge-repo/src/Repo.ts
+++ b/packages/automerge-repo/src/Repo.ts
@@ -753,6 +753,7 @@ export class Repo extends EventEmitter<RepoEvents> {
 
   shareConfigChanged() {
     this.synchronizer.reevaluateDocumentShare()
+    this.#subductionSource?.retryFailedSyncs()
   }
 }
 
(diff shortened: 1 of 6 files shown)


expede (approved):
Makes sense to me 👍

packages/automerge-repo/src/Repo.ts
```diff
@@ -753,6 +753,7 @@ export class Repo extends EventEmitter<RepoEvents> {
 
   shareConfigChanged() {
     this.synchronizer.reevaluateDocumentShare()
+    this.#subductionSource?.retryFailedSyncs()
```
alexjg: These two methods are basically doing the same thing. I think we should add a `shareConfigChanged` method to `DocumentSource` and then do `for (const source of this.#sources) { source.shareConfigChanged() }`.

jtfmumm: Done

PR #595, opened 2026-04-14
Fix import cycles introduced in v2.5.2

msakrejda:
The import cycles trip up esm.sh's module rewriter and lead it to
generate a broken stub for @automerge/vanillajs/slim?bundle-deps.

Add the ref constructor as an injected dependency (and clean up some
other minor issues) to avoid the circular imports.

### DIFF
--- a/packages/automerge-repo/src/DocHandle.ts
+++ b/packages/automerge-repo/src/DocHandle.ts
@@ -58,7 +58,7 @@ export class DocHandle<T> extends EventEmitter<DocHandleEvents<T>> {
   #refCache: Map<string, WeakRef<Ref<any>>> = new Map()
 
   /** Factory for creating Ref instances, injected by Repo to avoid circular imports */
-  #refConstructor?: <TDoc, TPath extends readonly PathInput[]>(
+  #refConstructor: <TDoc, TPath extends readonly PathInput[]>(
     handle: DocHandle<TDoc>,
     path: [...TPath]
   ) => Ref<any>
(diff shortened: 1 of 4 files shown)


packages/automerge-repo/src/DocHandle.ts
```diff
@@ -58,13 +57,23 @@ export class DocHandle<T> extends EventEmitter<DocHandleEvents<T>> {
   /** Cache for ref instances, keyed by serialized path */
   #refCache: Map<string, WeakRef<Ref<any>>> = new Map()
 
+  /** Factory for creating Ref instances, injected by Repo to avoid circular imports */
+  #refConstructor?: <TDoc, TPath extends readonly PathInput[]>(
```
alexjg: Is there any reason to make this optional? It seems to me like we would want to always pass it in.

msakrejda: The main reason is to avoid having to pass the option in all the places in tests where we build handles manually. But it also seems weird to have a required option, since the `options` argument itself is optional. I don't think we want to make options required, right?

alexjg: Yeah agreed it should not be a part of the options if we make it required. I personally think we should make this another constructor argument alongside the document ID. It's not really an option in the sense that the other options are, the refs functionality just doesn't work without it. I think we should bite the bullet on updating the tests if that's the only problem.

msakrejda: Sounds good; will do.

PR #598, opened 2026-04-18
Read-atomic/crash-atomic NodeFS & Subduction bridge

expede:
The NodeFS storage backend can fail and leave torn writes. This PR improves the NodeFS write safety guarantees for POSIX and Windows targets, and updates the Subduction storage bridge to write in the correct order given how this new atomic API works. Possibly controversial is that we explicitly `fsync` on POSIX, which adds up to 100us per write in exchange for durability guarantees.

### DIFF
--- a/packages/automerge-repo-storage-nodefs/src/index.ts
+++ b/packages/automerge-repo-storage-nodefs/src/index.ts
@@ -216,11 +216,21 @@ const atomicWrite = async (
   } finally {
     await fh.close()
     if (!wroteTmp) {
-      // Best-effort cleanup if writeFile/sync threw.
+      // Best-effort cleanup if writeFile/sync threw. The caller is
+      // about to see the outer writeFile/sync error; the cleanup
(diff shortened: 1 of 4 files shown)


alexjg:
My main worry with this PR is that it breaks old storage adapters which don't know to filter out the temporary files. We can fix that by putting the temporary files under a top level directory in the data directory the storage adapter is pointing at. 

I also think we should make this a PR to `main` and rebase `subductionjs` on top of it

expede:
@alexjg both good points. Will do!

expede:
Just FYI because I want to ensure that everything will work together, the plan is:

1. Get to a good place in this PR, then close it
2. Create a new branch off `main`
3. Copy the relevant changes for NodeFSStorageAdapter
4. Create PR
5. Rebase `subductionjs` over new `main`
6. Create new branch off `subductionjs`
7. Copy the bridge changes
8. Create PR

expede:
Superseded by #600

Issue #389, opened 2024-09-28
Storage errors cause node to exit


FauxFaux:
If the storage system throws an error, the exception unwinds back to here:

https://github.com/automerge/automerge-repo/blob/d3616cd6f8c3a7cdd511624b1cfe0b478dcffe5f/packages/automerge-repo/src/Repo.ts#L34-L38

It's not valid to return a rejected promise from this event handler, as nothing handles that rejected promise. The default `unhandledRejection` handler is to exit with status 1, so node exits.

I would rather that node didn't exit when there was an error, and some kind of error handler was called, so I can handle or retry said error.

Issue #297, opened 2024-02-28
WebSocket Ready flag should only be set after Open event


justin0mcateer:
Currently, the WebSocket Client Transport will mark itself as 'ready' once it has received a PeerMessage from the server, otherwise one second after construction.

However, if for some reason the WebSocket takes more than one second to fully establish, the Transport will be marked ready and will emit the 'ready' event prior to the WebSocket transport actually being ready to accept 'send' calls. This could be resolved by moving the 'auto-ready' setTimeout call into the 'onOpen' callback function.

We are periodically seeing 'Websocket not ready' messages emitted by this line:
https://github.com/automerge/automerge-repo/blob/main/packages/automerge-repo-network-websocket/src/BrowserWebSocketClientAdapter.ts#L141

Since Automerge Repo is fully responsible for sending messages, either there is an issue with the 'ready' handling, or the Transport is being marked 'ready' before the WebSocket is in the OPEN state. It should be noted that we were seeing this issue with the public/test Websocket server, which may be slower/more latent.

pvh:
Thanks for the bug report, @justin0mcateer! I see the comment indicates that this choice was deliberate by @alexjg so before I change this let's see if he has any comment on why he set it up that way.

alexjg:
The reason for the `setTimeout` which marks the adpater as ready is because we don't do _anything_ until all the network adapters have said they are ready. The idea of the "ready" callback was not to wait for peers who might never appear (because for example they are overloaded and will never respond), but just to avoid situations where an adapter will take a few miliseconds to set up (such as for the MessageChannel adapter) and consequently races with a `Repo.find` call on application startup. Frankly, this is a bit of a hack, a more principled approach probably means introducing some additional states in the `NetworkSubsystem` to track the state of an adapter (connecting, handshaking, ready etc.).

In this case however, we can probably continue hacking. The problem is that the `setTimeout` is in the `connect` handler, which means that if we take longer than a second to connect then we end up telling the network adapter we are ready for messages when we aren't. Probably the easiest way around this is to make the readiness timeout configurable - with the option to have no timeout whatsoever. It should be noted though that until we mark the adapter as ready the `Repo` will not do any networking at all, which means you'll force all requests for a document to wait on the adapter being ready.

Issue #275, opened 2024-01-22
Intermittent CI test failures


HerbCaudill:
The test action failed twice on #274, and passed the third time I ran it: 

❌ [Attempt #1](https://github.com/automerge/automerge-repo/actions/runs/7603968031/job/20706319740)
❌ [Attempt #2](https://github.com/automerge/automerge-repo/actions/runs/7603968031/job/20719542932)
✅ [Attempt #3](https://github.com/automerge/automerge-repo/actions/runs/7603968031/job/20719597238)

In all cases the tests are all passing, but Vitest fails the test run if any errors are thrown after the test environment is torn down. 

```
⎯⎯⎯⎯⎯⎯ Unhandled Errors ⎯⎯⎯⎯⎯⎯

Vitest caught 1 unhandled error during the test run.
This might cause false positive tests. Resolve unhandled errors to make sure your tests are not affected.

⎯⎯⎯⎯ Unhandled Rejection ⎯⎯⎯⎯⎯
Unknown Error: bad theirHave: error loading have at index 0: bad bloom: the value was not a Uint8Array
This error originated in "packages/automerge-repo/test/remoteHeads.test.ts" test file. It doesn't mean the error was thrown inside the file itself, but while it was running.
This error was caught after test environment was torn down. Make sure to cancel any running tasks before test finishes:
- cancel timeouts using clearTimeout and clearInterval
- wait for promises to resolve using the await keyword
⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯

 Test Files  15 passed (15)
      Tests  164 passed (164)
     Errors  1 error
```

I haven't been able to reproduce the failure locally, so it may have something to do with the CI runner being a bit slower than my laptop. This bit in the setup function looks suspicious to me: 

https://github.com/automerge/automerge-repo/blob/b87785a01e0b2944feea49b2342c960e09cf1c32/packages/automerge-repo/test/remoteHeads.test.ts#L89-L96

There are similarly arbitrary pauses in the tests themselves:

https://github.com/automerge/automerge-repo/blob/b87785a01e0b2944feea49b2342c960e09cf1c32/packages/automerge-repo/test/remoteHeads.test.ts#L112-L116

Issue #264, opened 2023-12-31
Data durability question


sroze:
Looking at the code, I'm trying to figure out how much I can trust automerge-repo from a durability question and I'd love to have a few pointers, for pieces I don't understand.

1. When peers send messages to each other, there doesn't seem to be any sort of acknowledgment. As such, is it fair to say that if a server fails to process a given message (i.e. storage fails or power outage at this moment for instance), other peers won't know about it? If so, wouldn't they continue to send further messages and fail to do so given the de-sync of heads?
2. [`Repo`'s use of the storage subsystem is out-of-band (in other to debounce).](https://github.com/automerge/automerge-repo/blob/5269a79f4f472a0653f39888b8941e49a4586226/packages/automerge-repo/src/Repo.ts#L78-L84). As such, does it mean that once the message has been sent and ack'ed by the Repo, there is a risk of data loss, because it might not have been persisted by the server?

Thank you so much 🙏

alexjg:
Good questions. 

1. There isn't an acknowledgement as such, although there is logic to make sure we don't send messages unless either we have new changes or the other end has responded to the last message we sent, so we don't send messages indefinitely. More importantly, the transport is assumed to be reliable and `NetworkAdapter`s should emit `"peer-disconnected"` when a connection is lost, which will notify the `Repo` to stop sending messages on that connection. Obviousy disconnection often requires waiting for a timeout in the underlying transport. 
2. True, currently it is possible for a peer to acknowledge changes but then to fall over before the change makes it to storage. In this case when the peer comes back up again then the sender would reconnect and resend their changes (this is handled by the sync protocol). 

To the second point, I have played with various implementations of "flush before ack" type things, but I'm still uncertain as to whether it's worth the performance hit and implementation complexity. One of the great things about local-first architectures is that you can be less dependent on the reliability of a single machine. Say the receiving peer is a sync server, if that peer falls over before saving to storage then as long as some peer has the data then they will resend it to the sync server when it comes back up. This is not to say that we _shouldn't_ implement some kind of enhanced durability mode, but I would like to have a motivating usecase for it because so far it's not been needed so I don't know what the API should look like. In that vein would you be able to describe your architecture and what kind of data loss you are concerned about?

sroze:
Very clear, thank you @alexjg. Just so I understand your point though, what would happen in the following scenario:

1. "client peer" sends an update, "server peer" receives it and stores it (successfully) in its storage.
[...]

PR #600, opened 2026-04-19
Read-atomic/crash-atomic NodeFSStorageAdapter

expede:
The NodeFS storage backend can fail and leave torn writes. This PR improves the NodeFS write safety guarantees for POSIX and Windows targets. Possibly controversial is that we explicitly `fsync` on POSIX, which adds up to 100us-1ms per write (depending on the specific SSD hardware) in exchange for durability guarantees. This doesn't get us to fully transactional writes (with CAS before/after gates, WAL, etc etc), but SIGNIFICANTLY improves atomic write reliability with read-/crash-atomicity.

Anecdotally: we were getting lots of torn writes in pushwork (due to an early exit bug) — we switched to this before fixing pushwork and haven't seen any torn writes since.

We've rebased `subductionjs` over this PR. That branch adds a `saveBatch` to the interface for performance (e.g. hitting IDB many times but then need to update all instances) — this PR can be applied to those semantics, too

### DIFF
--- a/packages/automerge-repo-storage-nodefs/src/index.ts
+++ b/packages/automerge-repo-storage-nodefs/src/index.ts
@@ -1,6 +1,27 @@
 /**
  * @packageDocumentation
- * A `StorageAdapter` which stores data in the local filesystem
+ * A `StorageAdapter` which stores data in the local filesystem.
+ *
+ * ## Durability and atomicity
+ *
+ * Writes use the standard POSIX "write-to-temporary + fsync + rename"
+ * pattern so that a reader or a crash never observes a half-written file:
+ *
+ *   1. The payload is written to `<baseDirectory>/.tmp/<pid>.<uuid>`.
+ *   2. The temporary file is `fsync`ed so its bytes reach disk.
+ *   3. `rename(2)` replaces the target atomically (POSIX) or via
+ *      `MoveFileExW(MOVEFILE_REPLACE_EXISTING)` (Windows NTFS).
+ *   4. On POSIX, the parent directory is `fsync`ed so the rename itself
+ *      is durable across a crash. Windows does not expose directory
+ *      fsync from user-space Node; directory metadata durability falls
+ *      back to the operating system's own guarantees.
+ *
+ * Temporary files live in a dedicated `<baseDirectory>/.tmp/` directory
+ * rather than as siblings of their target files. This keeps them on the
+ * same filesystem as their targets (required for atomic rename) while
+ * making them invisible to older adapters and to {@link loadRange} /
+ * {@link load}, which are always prefix-scoped and are unlikely to walk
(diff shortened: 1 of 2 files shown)


darcyparker:
@alexjg @msakrejda @expede, #600 lines up with a cluster of open issues and PRs that all look like one failure class, and it seemed worth connecting them in one place.

**The class:** async work from storage or sync (a rejected save/load, a throw while decoding a peer message) that is neither awaited nor caught. It surfaces as an unhandled rejection or uncaught exception, so in Node the process exits, and
any in-flight write is torn or lost.

You can reproduce the mechanism in isolation: #673 includes a self-contained script (eventemitter3 + a real socket `data` handler) where a throw in a listener escapes `emit()`, reaches the event loop, and exits the process with a non-zero
code. That is the same shape behind every PR below.

**Existing issues that are this class:**
- #389 "Storage errors cause node to exit": a `saveDoc` rejection from a  fire-and-forget listener becomes an `unhandledRejection` and exits. Still open, with a corroborating report.
- #275 "Intermittent CI test failures": the flake is an unhandled rejection (`bad bloom: ... not a Uint8Array`, a malformed inbound sync message) firing after teardown.
- #264 "Data durability question": the durability motivation behind this PR.

**The PRs, by layer:**
- Disk atomicity: #600 (write-temp + fsync + rename) and #602 (write order). The "even if we crash mid-write, don't tear the file" layer.
- Catching the error at the listener / fire-and-forget boundary. These are all the same shape (an event listener or un-awaited call whose async work rejects or throws), so the useful split is by what fails:
  - Storage I/O: #671 (`StorageSource` load), #676 (`StorageSource` save, on the  `heads-changed` listener, i.e. the save side of #389), and #672 (`StorageSubsystem` compaction).
  - Untrusted peer input: #673 (`Repo` inbound-message listener) and #674 (`WebSocketServerAdapter` decode). A malformed peer message can currently exit a Node sync server; #297 was an earlier instance on the send path, fixed by not throwing.

**Bottom line:** one failure class at different layers (uncaught async from storage/sync, leading to unhandled rejection / process exit / torn or lost writes). #600 makes a write survive a crash; the rejection-handling PRs stop the crash and stop dropping the error. They are complementary.

The deeper "do it properly" direction (make storage/network I/O abort-aware and properly awaited rather than fire-and-forget) is sketched in a WIP branch, [`darcy/promisifications_phase0`](https://github.com/darcyparker/automerge-repo/tree/darcy/promisifications_phase0): an optional `AbortSignal` on the adapter interfaces, `withAbort`/`withTimeout` helpers, and a `dev-docs/abort-patterns.md` write-up. It is not a PR, and I will likely break it into focused PRs the way #671 and #676 already are; sharing it for the direction.

Happy to fold the cross-references into #389 as an umbrella, or open a short tracking issue, if that is easier to follow.

PR #602, opened 2026-04-19
Add `saveBatch` and safer write order for Subduction bridge

expede:
This PR:

1. Adds a `saveBatch` to the storage interface to provide better semantic guarantees and speed up many backends (e.g. IDB)
2. changes the write order of storage in the Subduction bridge so that it provides better guarantees on failure and leads to fewer torn writes. This depends on changes in #600.

Writing blobs before metadata means that for backends that can observe partial failure, Subduction will only pick up on a value _if_ the metadata exists. At this stage, we may orphan failed writes. We can come back and clean them up later but at least naively it requires checking for bad data on lookup which slows things down.

We may still end up with partial failure, but the windows for that are MUCH narrower thanks to the 2-phase "stage each, then quickly `rename(2)` all and `fsync`" strategy.


expede:
@copilot thoughts?

expede:
feeling good about code state. I want to test some of the example apps manually before merging. Will be back in a bit

PR #604, opened 2026-04-20
add refFromObject function which constructs a Ref from a doc sub-object

OrionReed:
```ts
/**
 * Create a ref from a sub-object of an Automerge document.
 *
 * This lets you take a value read out of the document (a map or list) and
 * recover a {@link Ref} pointing at that location, without having to know or
 * re-construct the path by hand.
 *
 * ```ts
 * const doc = handle.doc()
 * const foo = doc.bar.foo
 * const ref = refFromObject(handle, foo)
 * ref.value()          // same object
 * ref.change(f => { f.x = 42 })
 * ```
 *
 * Works both on the materialized document returned by `handle.doc()` and on
 * the live proxy passed into `handle.change(d => ...)` callbacks.
 *
 * Limitations:
 * - The value must be a map or list sub-object of the document. Primitives
 *   (numbers, booleans), text strings, {@link Counter}s,
 *   {@link ImmutableString}s, and {@link Date} values do not carry path
 *   information and cannot be used. For those, use `handle.ref(...)` or
 *   {@link refFromString} with an explicit path.
 * - Array element refs are position-based, just like `handle.ref('list', 0)`.
 *   If you need stability across concurrent inserts, build a pattern ref
 *   instead: `handle.ref('list', { id })`.
 * - The object must belong to the same document as the provided handle.
 *
 * @experimental This API is experimental and may change in future versions.
 *
 * @throws Error if `value` is not an Automerge map or list sub-object.
 * @throws Error if the object does not belong to `handle`'s document.
 */
 function refFromObject<TValue = unknown>(
  handle: DocHandle<any>,
  value: TValue
): Ref<TValue>
```

### DIFF
--- a/packages/automerge-repo/src/refs/utils.ts
+++ b/packages/automerge-repo/src/refs/utils.ts
@@ -16,8 +16,6 @@ import { CURSOR_MARKER } from "./types.js"
 import { RefImpl } from "./ref.js"
 import { parseRefUrl } from "./parser.js"
 
-const AUTOMERGE_STATE = Symbol.for("_am_meta")
-
 /**
  * Create a cursor-based range segment for stable text selection.
  *
@@ -138,8 +136,9 @@ export async function findRef<TValue = unknown>(
  * ref.change(f => { f.x = 42 })
  * ```
  *
- * Works both on the materialized document returned by `handle.doc()` and on
- * the live proxy passed into `handle.change(d => ...)` callbacks.
+ * Works on the materialized document returned by `handle.doc()`, the live
+ * proxy passed into `handle.change(d => ...)` callbacks, and objects read
+ * from any view of the same document.
  *
  * Limitations:
  * - The value must be a map or list sub-object of the document. Primitives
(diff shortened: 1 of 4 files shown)


packages/automerge-repo/src/refs/utils.ts
```diff
@@ -119,6 +123,98 @@ export async function findRef<TValue = unknown>(
...
+    return new RefImpl(handle, [] as unknown as AnyPathInput[]) as Ref<TValue>
+  }
+
+  let info: { path?: Automerge.Prop[] }
+  try {
+    info = docBackend.objInfo(objectId)
+  } catch (err) {
+    throw new Error(
+      "refFromObject: object is not present in the current document. " +
+        "It may belong to a different document or a stale view. " +
```
alexjg: I don't think a stale view should cause problems.

OrionReed: Agreed, pushed a change

msakrejda:
Is this still relevant after #640 ?

OrionReed:
@msakrejda I think not. The desire for that functionality is still there, but:
- the code is now wholly out of date
- users can do this themselves if needed

I'll close it.

PR #605, opened 2026-04-20
Unify DocHandle and Ref into a single path-aware handle

pvh:
A "ref" is now just a DocHandle scoped to a path. Sub-handles delegate lifecycle and network state to a root handle but carry their own #path (and optionally a cursor #range), so working with part of a document feels just like working with a whole document: same doc(), value(), history(), view()/viewAt(), change(), on("change"), url, remove(), equals/contains/overlaps, etc.

- DocHandle.url returns an AutomergeUrl for root handles and a RefUrl (automerge:docId/path#heads) for sub-handles, via stringifyRefUrl.
- DocHandle.on("change") and onChange() filter patches to the sub-handle's path; onChange callbacks receive the scoped value.
- DocHandle.history() on a sub-handle is filtered to heads where the sub-path actually changed, using A.topoHistoryTraversal + A.diff.
- viewAt(heads) accepts both UrlHeads and raw Automerge heads.
- Repo.find() now accepts RefUrls: it resolves the root document, views at heads if present, and scopes into the requested path.
- refs/types.ts: Ref<T> is a deprecated alias for DocHandle<T>.
- refs/ref.ts (RefImpl) is removed; refs/utils.ts helpers create handles via handle.ref(...). refs/parser.ts exports isValidRefUrl.

Includes a new test/refs/unified.test.ts covering URL round-trips, repo.find(refUrl), scoped change events, and history filtering, plus a runnable examples/sub-doc-handles.mjs that tours the API end-to-end.

Made-with: Cursor


OrionReed:
## Urls
Right now heads are encoded two different ways depending on the handle, this was a mistake in the original implementation of Refs which has been inherited here. The root branch passes `#fixedHeads` (bs58check-encoded UrlHeads) straight through. The sub-handle branch `decodeHeads` back to raw hex and emits those. So:
```
handle.view(h).url → automerge:docId#<bs58check>
handle.ref("x").viewAt(h).url → automerge:docId/x#<hex>
```
We should extend `AutomergeUrl` to have an optional `segments` and extend `parseAutomergeUrl` to accept a `/path` after documentID and extend the regex from `^automerge:(\w+)$` to `^automerge:(\w+)(?:/(.*))?$` (ish, needs testing)

`isValidAutomergeUrl` should accept `AutomergeUrl` with path too.

This also avoids a lot of cascading type churn of `AutomergeUrl | RefUrl`

## Event Dispatch
the current design is N listeners per change, but I think a ~1 trie walk is a much faster option that allows folks to go ham with the number of refs they use. Currently, each sub-handle subscribes **five listeners** on the root and two of them run on every change, walking the patch list and/or the sub-handle's own path independently.

If the root owns the dispatch for all sub-handles we can speed this up dramatically and I suspect in the end it'll also be more maintainable.

## `if (this.#root)` scattering
Not sure if this is the best answer but we could extract a small `SubHandleDelegate`. Private module with applyScopedChange, applyScopedRemove, scopedValue, wireForwarding, propPath so that DocHandle method bodies stop branching by moving the #isSubHandle decision to the top and calling this.#delegate.x or this.#rootMachine.x. This way we lose the scattered if/else without changing public types.

## Misc
- `#viewCache` is a `Map<string, DocHandle<T>>` of strong references, this means time travel UIs and stuff like that will slowly leak, this could be a `WeakRef`-based like `#refCache` already is?
- note that `valueOf()` changes DocHandle comparison semantics
- need to check we still get nice type inference

msakrejda:
+1 on Orion's feedback, but otherwise the new example looks good to me. It's nice to have a more unified API here.

msakrejda:
Closing since this was superseded by #640.

PR #606, opened 2026-04-21
Add asyncThrottle() to prevent races, strengthen abortable() and add consts to keep code DRY 

darcyparker:
- Adds a new `asyncThrottle()` helper to replace the current sync `throttle()`
   - This prevents potential race between the throttle period and the previous calls of an async function
   - Added tests to demonstrate the race issues and integration tests
- strengthen `abortable()`
    - handles minor case where signal is already aborted when `abortable()` is called
    - ensure signal event listener is cleaned up when the signal is never aborted
    - add tests
- adds some consts to keep code DRY
   - `foreverPromise`; (used in Repo to prevent possible leak of new promises that never settle)
   - `noop()`
   - `truePromiseFactory()` and `falsePromiseFactory()`
- refactors the `pause()` helper to support `signal: AbortSignal` option 
    - This is helpful for future if we want to use pause() in method with abort signal.
    - And add tests
- minor fix of `import` of `UrlHeads` type in `AutomergeUrl.test.ts`

### DIFF
--- a/packages/automerge-repo/src/Repo.ts
+++ b/packages/automerge-repo/src/Repo.ts
@@ -19,7 +19,7 @@ import {
 } from "./DocHandle.js"
 import { RemoteHeadsSubscriptions } from "./RemoteHeadsSubscriptions.js"
 import { headsAreSame } from "./helpers/headsAreSame.js"
-import { throttle } from "./helpers/throttle.js"
+import { asyncThrottle } from "./helpers/throttle.js"
 import {
   NetworkAdapterInterface,
   type PeerMetadata,
@@ -102,7 +102,7 @@ export class Repo extends EventEmitter<RepoEvents> {
   #progressCache: Record<DocumentId, FindProgress<any>> = {}
   #saveFns: Record<
     DocumentId,
-    (payload: DocHandleEncodedChangePayload<any>) => void
+    (payload: DocHandleEncodedChangePayload<any>) => Promise<void>
   > = {}
   #idFactory: ((initialHeads: Heads) => Promise<Uint8Array>) | null
 
@@ -190,13 +190,14 @@ export class Repo extends EventEmitter<RepoEvents> {
       this.#saveFn = ({ handle, doc }: DocHandleEncodedChangePayload<any>) => {
         let fn = this.#saveFns[handle.documentId]
         if (!fn) {
-          fn = throttle(
-            ({ doc, handle }: DocHandleEncodedChangePayload<any>) => {
-              void this.storageSubsystem!.saveDoc(handle.documentId, doc)
-            },
+          fn = this.#saveFns[handle.documentId] = asyncThrottle(
+            ({
+              doc,
+              handle,
+            }: DocHandleEncodedChangePayload<any>): Promise<void> =>
+              this.storageSubsystem!.saveDoc(handle.documentId, doc),
             this.#saveDebounceRate
           )
-          this.#saveFns[handle.documentId] = fn
         }
         fn({ handle, doc })
       }
@@ -364,7 +365,7 @@ export class Repo extends EventEmitter<RepoEvents> {
 
   #throttledSaveSyncStateHandlers: Record<
     StorageId,
-    (payload: SyncStatePayload) => void
+    (payload: SyncStatePayload) => Promise<void>
   > = {}
 
   /** saves sync state throttled per storage id, if a peer doesn't have a storage id it's sync state is not persisted */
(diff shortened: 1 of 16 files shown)


packages/automerge-repo/src/helpers/pause.ts
```diff
@@ -1,6 +1,72 @@
...
+ * ```typescript
+ * const signal1 = new AbortController().signal;
+ * const signal2 = new AbortController().signal;
+ * const combinedSignal = AbortSignal.any([signal1, signal2]);
+ * await pause(5000, combinedSignal); // Pause until timeout or either signal aborts
+ * ```
+ */
+export const pause = (
+  milliseconds: number = 0,
+  options?: { signal?: AbortSignal }
```
darcyparker, line 47: When possible, it is helpful when async fns support abort signals. I'd like to see the interfaces for storage and network adapter to support signals.  But there is some async work that developers probably want to continue (and not be abortable).  

Your `Promise.race()` and `abortable()` are a better mechanism sometimes.

I have not been able to update the storage and network adapter interfaces and get unit tests working... but when I or someone comes back to this, it will be helpful to have this option for abort signal.   In the mean time, if you want me to remove this update from PR, I can do that.

packages/automerge-repo/src/Repo.ts
```diff
@@ -642,7 +645,8 @@ export class Repo extends EventEmitter<RepoEvents> {
       documentId,
       handle,
       progressSignal,
-      signal ? abortable(new Promise(() => {}), signal) : new Promise(() => {})
+      //IMPORTANT: Reuse foreverPromise. Don't keep re-creating promises that never settle like `new Promise(() => {})`
```
msakrejda:
```suggestion
```
I think the change itself makes sense, but I'm skeptical about the value of this comment: it's not likely someone will revert the change, and someone doing it elsewhere probably won't see the comment. Maybe we can add linting to avoid literal `new Promise(() => {})` outside of the `foreverPromise` definition instead?

packages/automerge-repo/test/DocSynchronizer.test.ts
```diff
@@ -81,6 +82,51 @@ describe("DocSynchronizer", () => {
...
+
+    // Fire rapid changes spaced so later ones land while a prior run is still
+    // in #syncWithPeers (waiting on the slow whenReady).
+    for (let i = 0; i < 6; i++) {
+      handle.change(doc => {
+        doc.foo = `v${i}`
+      })
+      await pause(30)
+    }
+    await pause(600)
```
msakrejda: I'm wary of all the `pause`s with arbitrary times here. Is there a way we can avoid that and progress based on state changes instead?

packages/automerge-repo/test/pause.test.ts
```diff
@@ -0,0 +1,73 @@
...
+    it("rejects immediately with AbortError when the signal is already aborted", async () => {
+      const controller = new AbortController()
+      controller.abort()
+
+      const start = Date.now()
+      await expect(
+        pause(1000, { signal: controller.signal })
+      ).rejects.toBeInstanceOf(AbortError)
+      // Fast-path: reject without waiting for the 1000ms timer.
+      expect(Date.now() - start).toBeLessThan(50)
```
msakrejda: I don't see a way to avoid this pause, but maybe we should make the pause timeout much larger and be more lenient in the expectation? I don't know if this is actually likely to be a problem, but I've seen some weird clock issues in CI, and in the happy path, there's no actual delay with the suggested change. If you think I'm being paranoid, feel free to push back.

msakrejda (commented):
Nice, thanks for the contribution. I'm a little wary of the pause-heavy tests and whether they'll prove stable in CI, but otherwise this looks great. Thoughts?

packages/automerge-repo/test/pause.test.ts
darcyparker: Good point...  I refactored to use `vitest`s `vi.fakeTimers()`.  This allows assertions on what was scheduled rather than measuring wall clock elapsed time.  I see this is in `DocHandle.test.ts` already https://github.com/automerge/automerge-repo/blob/main/packages/automerge-repo/test/DocHandle.test.ts#L264 and I should have noticed this before.

The tests run faster now too.  See e55e2064e31de6dccc875e91d705bfcd8308682d

packages/automerge-repo/test/DocSynchronizer.test.ts
darcyparker: Good point. (I looked at this after the updates to `pause.test.ts` below.)

Converted the throttle-serialization test to vitest fake timers (vi.useFakeTimers + advanceTimersByTimeAsync). Time advances are deterministic now — no wall clock, no CI clock flakiness, and runtime drops from  ~800ms to ~5ms. See 1411f753e55ab5c766f1af12b3621429ffd17c9d

Note: It's not fully state/event-driven: `asyncThrottle` uses `setTimeout` internally, so the test still advances the fake clock to fire that scheduling rather than driving purely on observed `whenReady` calls. A pure event-driven version would need a gate-released `whenReady` plus fake timers for the throttle internals — more machinery for marginal gain over the current maxConcurrent === 1 assertion.

Hope you agree that the CI clock flakiness is resolved now... and not worth the additional machinery.  But let me know if you think otherwise.

darcyparker:
> Nice, thanks for the contribution. I'm a little wary of the pause-heavy tests and whether they'll prove stable in CI, but otherwise this looks great. Thoughts?

@msakrejda Thanks for review. It was good insight about the pause-heavy tests and CI flakiness concern. See last few commits. I refactored and replied to your comments. Please `Resolve conversation` if you think they are addressed properly, or let me know otherwise.

msakrejda:
Awesome, looks good. I'll leave this open for a bit in case anyone else has feedback and will merge soon if not.

PR #612, opened 2026-05-06
feat: add WeakValueCache and refactor existing #refCache and #viewCache to use it

darcyparker:
- `#refCache` previously used `Map<string, WeakRef>`, which is essentially same as `WeakValueMap` (included in PR).  But `WeakValueMap` has additional benefit of cleaning up dead string keys.  So there is no real functional change other than deleting dead string keys. (which I think is a good thing)
- `#viewCache` previously used strong refs and each entry pined a full `A.clone(doc)`. 
    - I think it was an oversight to not use a similar `Map<string, WeakRef>` for the `#viewCache` like `#refCache`
    - The functional change is that the `#viewCache` no longer prevents garbage collection if the user stops holding a strong ref to the view handle. (This is a good thing. As a user, I want garbage collection.)
     - view handles are never registered with storage or sync subsystems, and nothing holds a strong ref unless the automerge-repo user does, so the previous behavior (of holding onto the view handles strongly) meant garbage collection didn't happen without `Repo.removeFromCache()`
     - new behavior: `view()` has no side effects beyond cloning. A re-clone on cache miss is a performance hit (full doc clone) but not a correctness hit.  I think the re-clone on cache miss is acceptable in order to benefit from garbage collection when the automerge-repo users stop holding strong references.
- To keep code DRY for the 2 caches (`#refCache` and `#viewCache`), and to add benefit of cleaning up dead key strings (lazily), I implemented `WeakValueMap`.

### DIFF
--- a/packages/automerge-repo/src/DocHandle.ts
+++ b/packages/automerge-repo/src/DocHandle.ts
@@ -18,6 +18,7 @@ import {
   isAbortErrorLike,
 } from "./helpers/abortable.js"
 import type { PathInput, InferRefType, Ref } from "./refs/types.js"
+import { WeakValueMap } from "./helpers/WeakValueMap.js"
 
 /**
  * A DocHandle is a wrapper around a single Automerge document that lets us listen for changes and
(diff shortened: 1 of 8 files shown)


darcyparker:
Note: Another cache `#handleCache` is another cache I considered for `WeakValueCache`.  But I would like some opinions from others before doing this refactor.

Why?: Right now `#handleCache()` grows w/o bounds.  And users of automerge-repo have to manually use `Repo.removeFromCache()`.  

Thoughts on how to deal with this:
- There has been consideration for an LRU or similar store to ensure memory doesn't grow w/o bounds.  But this adds some complexities
- But another option is to use a `WeakValueCache`.
   - My mental model as an automerge-repo user: "I drop my handle; the repo should follow suit'.
   - When a downstream user of automerge-repo drops their handle reference, then the following become collectable (garbage collection) as a single unit:
      1) the handle
      2) the synchronizer state
      3) progress entry and 
      4) saver throttle 
   - This is what it might look like:
        1. `WeakValueMap` for `#handleCache`
        2. `WeakRef` for `DocSynchronizer.#handle` + `.deref()` guards at every site
        3. `FinalizationRegistry` for the per-document Repo state
       4. Cache-miss no-op in the two sync-message routing sites
       5. Alive-iteration on WeakValueMap for repo.handles / repo.flush()
       6. Tests: a "drop reference + flush + GC → handle reclaimed" test, and a "sync message arrives for dropped handle → no crash, no panic" test
    - If we do this, then `removeFromCache()` becomes redundant, and we don't need to consider an LRU

At this time, I think this PR is worth merging... and this follow up on `#handleCache` is a good future PR consideration.

darcyparker:
Reviewed https://github.com/automerge/automerge-repo/pull/618.  This PR has no design risk if 618 merges first.  There will be a simple low effort rebase.  So I recommend having a quick look at the gist of this PR, and when 618 merges, I'll rebase. 

Update: I have `darcy/weakCaches_on_PR618` https://github.com/darcyparker/automerge-repo/tree/darcy/weakCaches_on_PR618 branch rebased and ready to replace this PR if PR 618 merges.

darcyparker:
https://github.com/automerge/automerge-repo/pull/637 replaces this PR

PR #615, opened 2026-05-06
Add opt-in idle eviction for handleCache (#358)

### LINKED ISSUE #358: Inactive `DocHandle`s are not evicted from `handleCache`

caseprince:
## Summary

Fixes #358.

`Repo#handleCache` grows without bound: every doc that gets opened stays in memory until the process exits. On a long-running sync server with many docs and reconnecting clients, that's a memory leak in practice.

Adds two opt-in `RepoConfig` options:

- `idleEvictionMs`: handle is eligible for eviction after this many ms with no activity.
- `idleScanIntervalMs`: period of the eviction scan.

Both unset (the default) disables eviction entirely. No listeners are attached, no timestamps recorded, no `setInterval` scheduled. Existing consumers pay nothing.

When enabled, activity is tracked on four signals:

- `#getHandle` access (`create` / imports / cache misses)
- cached `find` / `findWithProgress` / `findClassic` access
- synchronizer `sync-state` events (incoming peer traffic)
- `DocHandle` `heads-changed` events (covers offline local writes)

The scan walks `#handleCache`, skips handles touched within `idleEvictionMs`, skips handles that still have at least one peer, and calls the existing public `removeFromCache()` for the rest. `removeFromCache` already takes care of `synchronizer.removeDocument`, `progressCache`, and `saveFns` cleanup.

Notable details:

- **Single-flight**: a tick is skipped if the previous one is still running, so a slow `removeFromCache` (e.g. a handle stuck in `REQUESTING`) can't stack ticks and race over `#handleCache`.
- **Per-tick cap (100 successful evictions)** bounds evictions per scan.
- `setInterval(...).unref()` so the timer never blocks process exit.
- `shutdown()` clears the interval.

A doc that's held externally but quiet for longer than `idleEvictionMs` and has no peers will be evicted and re-loaded from storage on next access.

## Why a heuristic rather than `WeakRef` + `FinalizationRegistry`

The thread on #358 discusses an `InternalDocHandle` + `DocHandle` split using `WeakRef` and `FinalizationRegistry`. This PR takes a simpler route on purpose:

- Stays inside the existing public API (`removeFromCache`).
- Deterministic to test, no GC timing dependence.
- Opt-in, so the risk surface for existing consumers is zero.
- Trivially removable if a more sophisticated approach lands later.

## Test plan

Five new tests in `Repo.test.ts > local only > idle eviction`:

- `does not evict any handles when idleEvictionMs/idleScanIntervalMs are unset`: confirms the default no-op path.
- `evicts an idle handle once idleEvictionMs has elapsed`: basic happy path.
- `treats a cached find as activity`: confirms cached handle access refreshes idleness.
- `does not evict a handle that has at least one active peer`: confirms the peers-skip gate.
- `does not enable activity tracking when eviction is disabled`: confirms no `heads-changed` listener is attached when off.

Local verification:

- `pnpm vitest run packages/automerge-repo/test/Repo.test.ts`
- `pnpm -F @automerge/automerge-repo build`
- `pnpm prettier --check packages/automerge-repo/src/Repo.ts packages/automerge-repo/test/Repo.test.ts`

### DIFF
--- a/packages/automerge-repo/src/Repo.ts
+++ b/packages/automerge-repo/src/Repo.ts
@@ -84,6 +84,12 @@ export class Repo extends EventEmitter<RepoEvents> {
   #saveFn: (payload: DocHandleEncodedChangePayload<any>) => void
 
   #handleCache: Record<DocumentId, DocHandle<any>> = {}
+  #lastActivity: Record<DocumentId, number> = {}
+  #evictionIntervalId: ReturnType<typeof setInterval> | null = null
+  #evictionInFlight = false
+  #touchOnHeadsChanged = ({ handle }: DocHandleEncodedChangePayload<any>) => {
+    this.#touch(handle.documentId)
+  }
 
   /** @hidden */
   synchronizer: CollectionSynchronizer
@@ -117,6 +123,8 @@ export class Repo extends EventEmitter<RepoEvents> {
     denylist = [],
     saveDebounceRate = 100,
     idFactory,
+    idleEvictionMs,
+    idleScanIntervalMs,
   }: RepoConfig = {}) {
     super()
     this.#remoteHeadsGossipingEnabled = enableRemoteHeadsGossiping
@@ -253,6 +261,7 @@ export class Repo extends EventEmitter<RepoEvents> {
     })
 
     this.synchronizer.on("sync-state", message => {
+      this.#touch(message.documentId)
       this.#saveSyncState(message)
 
       const handle = this.#handleCache[message.documentId]
@@ -322,6 +331,17 @@ export class Repo extends EventEmitter<RepoEvents> {
         }
       )
     }
+
+    if ((idleEvictionMs ?? 0) > 0 && (idleScanIntervalMs ?? 0) > 0) {
+      const interval = setInterval(() => {
+        void this.#evictIdleHandles(idleEvictionMs!)
+      }, idleScanIntervalMs)
+      // Don't keep the event loop alive for the eviction timer.
+      if (typeof interval === "object" && "unref" in interval) {
+        interval.unref()
+      }
+      this.#evictionIntervalId = interval
+    }
   }
 
   // The `document` event is fired by the DocCollection any time we create a new document or look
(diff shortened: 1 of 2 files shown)


caseprince:
This implementation is now stale and would need to be re-authored against the new `DocumentQuery/DocumentSource` models. The motivation (https://github.com/automerge/automerge-repo/issues/358) is still valid and unaddressed.

PR #617, opened 2026-05-07
`subductionjs` shutdown perf

expede:
Bench originally reported by DXOS. Handled some low hanging perf fruit that we've been meaning to address for a couple weeks:


expede:
@copilot please review

PR #618, opened 2026-05-07
Remove xstate, add DocumentQuery

alexjg:
The xstate state machine in DocHandle is not helping us very much. Since the move to an async `find()` which only returns a handle when the document is available the `DocHandle` is primarily an API for users to publish changes to, and to listen for changes. Mixing this functionality with the internal state tracking document state is quite confusing. The xstate state machine itself is quite simple, the problem is really all the different places in the Repo which interact with it.

This commit attempts to simplify the situation by introducing the idea of a `DocumentQuery`, which tracks the state of a request to find some document. Document queries are created whenever we call `findWithProgress` or when we receive a sync message for the document in question. Events which would have previously been routed to the DocHandle are now routed to the DocumentQuery - which has an attached `DocHandle` that is given out when the document is available.

I have also introduced the idea of a `DocumentSource`, which is something that knows how to try and find a document. In this commit there are two sources: the storage subsystem and the `CollectionSynchronizer`. This integration point is where we have been able to fit subduction in to the model.

This PR attempts to stay entirely compatible with the existing implementation so we don't have to bump a major yet. The only major departure there is that `find` no longer times out.


darcyparker:
@alexjg - This PR looks good and happens to solve an eviction race issue we're experiencing. I created https://github.com/automerge/automerge-repo/pull/624 with a test to demonstrate the eviction race (passes with this PR's changes).  https://github.com/automerge/automerge-repo/pull/626 is an alternative fix (different solution) to the eviction race to consider. But if this merges first, 626 can be closed.

PR #621, opened 2026-05-12
Allow injecting a logger; default to console

simleb:
## Summary

- Adds a `Logger` interface (`debug`/`info`/`warn`/`error`) and a `consoleLogger` default that wraps the global `console`.
- Adds an optional `logger` field on `RepoConfig`; the `Repo` plumbs it through to `DocHandle`, `CollectionSynchronizer`, and `DocSynchronizer`.
- Replaces the unconditional `console.log` / `console.warn` calls in `Repo`, `DocHandle`, and the synchronizers with calls on the injected logger. The remaining `console.warn` for the legacy-UUID deprecation in `AutomergeUrl.ts` is intentionally left as-is (it's a module-level utility with no instance context, and the deprecation is on its way out).
- The `Logger` interface shape matches the conventions used by winston, pino, and bunyan, so most consumers should be able to pass an existing logger instance directly.

## Why

Today `automerge-repo` writes error/deprecation output straight to `console.log` / `console.warn`. Consumers that have a structured logger (in my case Winston) end up with this output bypassing their log pipeline, breaking formatting, log levels, and downstream sinks. With this change, any consumer can pass a logger and route that output wherever they like; if they don't, behaviour is unchanged.

A few of the migrated call sites were also using `console.log` to log errors — those now go through `logger.error` instead.

## Not in scope (possible follow-ups)

- Routing the `debug`-package namespaced trace logs through the `Logger` (different ergonomics: `debug` is zero-cost when disabled and supports namespace-based filtering, whereas `Logger.debug` would always fire).
- Migrating the legacy-UUID `console.warn` in `AutomergeUrl.ts`.

## Test plan

- [x] `pnpm exec tsc --noEmit` in `packages/automerge-repo` passes.
- [x] Full `@automerge/automerge-repo` test suite still passes (256 passed, 1 pre-existing skip).
- [x] New `Logger.test.ts` covers: `consoleLogger` shape, injected logger receives `DocHandle.docSync` deprecation warning, and `console.warn` is *not* called when a custom logger is injected.


packages/automerge-repo/test/Logger.test.ts
```diff
@@ -0,0 +1,41 @@
...
+  warn: vi.fn(),
+  error: vi.fn(),
+})
+
+describe("Logger injection", () => {
+  afterEach(() => {
+    vi.restoreAllMocks()
+  })
+
+  it("defaults to the global console", () => {
```
msakrejda: No big deal, but maybe
```suggestion
  it("logs to the global console", () => {
```
? The defaulting itself happens in the callers: this logger is what is used by the default by them. We're not testing that here.

packages/automerge-repo/test/Logger.test.ts
```diff
@@ -0,0 +1,41 @@
...
+    const consoleWarn = vi.spyOn(console, "warn")
+
+    const repo = new Repo({ logger })
+    const handle = repo.create<{ x: number }>({ x: 1 })
+    handle.docSync()
+
+    expect(logger.warn).toHaveBeenCalledOnce()
+    expect(logger.warn.mock.calls[0][0]).toContain("docSync is deprecated")
+    expect(consoleWarn).not.toHaveBeenCalled()
+  })
```
msakrejda, line 65: Nice.

msakrejda (approved):
Makes sense and looks good to me. I'll give some time for others to review, but otherwise will merge soon.

msakrejda:
On second thought: stupid question, but can't we just route these direct `console.log` calls through `debug` instead?

simleb:
> On second thought: stupid question, but can't we just route these direct `console.log` calls through `debug` instead?

Yes, that's the alternative. I guess I'll return the question to you: are any/all of these logs valuable for the client or is it all internal/debug stuff? I'm also happy to convert only a subset if you tell me which are which.

simleb:
I guess I can attempt to answer my own question: I think I'd prefer my backend service (sync-server) to log these errors so I can investigate issues more easily. Even if I redeployed with the debug flag on, it wouldn't play well with my existing logger.

simleb:
I think another fair question is whether the default should be console logging or nothing. I went with console logging for preserving the status quo.

msakrejda:
> I guess I'll return the question to you: are any/all of these logs valuable for the client or is it all internal/debug stuff?

Yeah, fair question. I'm a relatively new maintainer so I don't know the reasoning behind all the logging decisions. But grepping the repository for both `debug(` and `console.*(`, we're not super consistent about when to use `debug` and where we're calling console methods directly (e.g., we've got [warnings](https://github.com/automerge/automerge-repo/blob/main/packages/automerge-repo/src/Repo.ts#L628-L630) and [errors](https://github.com/automerge/automerge-repo/blob/main/packages/automerge-repo/src/synchronizer/DocSynchronizer.ts#L230-L232) logged to `debug`). And `debug` doesn't really seem to have the concept of levels, which makes this trickier.

But I think this configurable logging (beyond what `debug` offers) would be useful in more than just these classes. And even in just these classes, treating these loggers separately leads to stuff like

```ts
    this.#logger = logger ?? consoleLogger

    const docId = handle.documentId.slice(0, 5)
    this.#log = debug(`automerge-repo:docsync:${docId}`)
```

in [DocSynchronizer](https://github.com/automerge/automerge-repo/pull/621/changes#diff-40d85e011bd64146db0940336f591d11a0470c4521cf30688f3fa2ee96b6f3fdR82-R85). I hadn't noticed that before, but having both `#logger` and `#log` variables is pretty confusing.

Maybe we need to wrap `debug` in our own custom logger and use the wrapper everywhere? And maybe that custom logger could remain a global (like `debug` itself) to avoid having to plumb it around where it's needed (since it's basically write-only, I don't think that will push us into the typical global pitfalls).

I'm not trying to move the goalposts for this PR, just trying to think through what the right logging approach should be. Then we can work backward to figure out how to address your use case. Thoughts?

simleb:
No worries about extending the scope, it's not like I'm coding by hand anyway… 🙂

I pushed a new version that unifies everything under a single system. It's global like you suggested. By default, it still uses `debug` for debug log (so you can still set `DEBUG` with filters and all) and console for info/warning/errors. It's easy to swap the default with a real logger. I added some documentation.

I force-pushed since it's basically restarting from scratch.

msakrejda:
Thanks for the updates and sorry for the delay here: I'll probably have time to look at this again later this week. If you can resolve the conflict (at a glance, it looks like it's trivial: I think we just need both imports) before then, that'd be great, but not a big deal.

packages/automerge-repo/src/network/NetworkSubsystem.ts
```diff
@@ -128,7 +128,9 @@ export class NetworkSubsystem extends EventEmitter<NetworkSubsystemEvents> {
   send(message: MessageContents) {
     const peer = this.#adaptersByPeer[message.targetId]
     if (!peer) {
-      this.#log(`Tried to send message but peer not found: ${message.targetId}`)
+      this.#log.warn(
+        `Tried to send message but peer not found: ${message.targetId}`
+      )
```
msakrejda, line 133: I tried to convince myself this would be better left as debug, but I agree. If it's too noisy, we can revisit.

packages/automerge-repo/test/Logger.test.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    expect(factory).toHaveBeenCalledWith("test:ns")
+    expect(loggers.get("test:ns")?.warn).toHaveBeenCalledWith("hello")
+  })
+
+  it("hands subsystem namespaces to the factory", () => {
+    const { factory } = installMockFactory()
+    const repo = new Repo()
+    // Triggers a warn on the Repo's logger (remote heads gossiping is off by default).
+    repo.subscribeToRemotes([])
```
msakrejda: I don't think it makes sense to rely on `Repo` in this test: if that behavior no longer emits a warning, this test would start failing. Can we test this by creating a dummy caller in the test instead?

packages/automerge-repo/test/Logger.test.ts
```diff
@@ -0,0 +1,83 @@
...
+    handle.docSync()
+
+    const dochandleLogger = loggers.get("automerge-repo:dochandle")
+    expect(dochandleLogger).toBeDefined()
+    expect(dochandleLogger!.warn).toHaveBeenCalledOnce()
+    expect(dochandleLogger!.warn.mock.calls[0][0]).toContain(
+      "docSync is deprecated"
+    )
+    expect(consoleWarn).not.toHaveBeenCalled()
+  })
```
msakrejda:
```suggestion
```
While I thought this test worked well for the targeted fix, now that we're fixing logging more broadly instead, this feels too specific. We should just drop it.

msakrejda (approved):
Thanks! This looks great. This promotes a lot of debug logging to warnings and errors, so it will mean noisier logging by default if people are actually hitting those issues, but that's probably good in the long term: it will push us to fix real issues and remove or deprioritize routine logging to avoid noise.

If you could address the minor comments, that'd be great.

msakrejda (approved):
Great, thanks for updating!

PR #623, opened 2026-05-13
test: hybrid fake/real timers in throttle.test.ts

darcyparker:
This improves tests for `asyncThrottle()` that was added in PR #606.

`asyncThrottle` tests verify two different kinds of property:

1. Logical timing behavior — coalescing, cancellation, return-value plumbing, error propagation, gap measurement, in-process serialization. These care about *what* runs *with what args* *in what order*; real wall-clock time is incidental.

2. Concurrency property — the headline guarantee that asyncThrottle prevents the overlap race plain throttle exhibits.

For (1): Tests now use fake timers
For (2): Tests are verified with both:

A) fake-timers (primary correctness), and 
B) again with real setTimeout and real microtask/macrotask interleaving and real scheduling.  

-  We don't have a known case where real timers catch a bug fake timers miss, but the runtime cost is small and the coverage is cheap insurance.
-  Real-timer tests run N iterations. The asyncThrottle assertion runs across every iteration (a single overlap anywhere = bug). The plain-throttle assertion requires overlap in at least one iteration (real scheduling can collapse the overlap window in a single run).
-  Test bodies are lifted into exercisePlainThrottle/exerciseAsyncThrottle helpers parameterized over `advance(ms)`, so the fake and real describes share identical scenarios and differ only in timer mode.

Tests still cover the same assertions and add explicit fake-timer coverage of the concurrency property.

Suite runtime rises (~800ms → ~1.7s) because the real-timer concurrency tests now run 3 iterations each. The fake-timer half runs in single-digit ms. The slowdown is the deliberate cost of defense-in-depth + flake reduction; tune ITERATIONS down if the trade-off shifts.

### DIFF
--- a/packages/automerge-repo/test/throttle.test.ts
+++ b/packages/automerge-repo/test/throttle.test.ts
@@ -1,8 +1,45 @@
-import { describe, it, expect } from "vitest"
+import { afterEach, beforeEach, describe, it, expect, vi } from "vitest"
 import { asyncThrottle, throttle } from "../src/helpers/throttle.js"
 import { pause } from "../src/helpers/pause.js"
 
+// These tests use two complementary styles deliberately:
+//
+// 1. **Fake timers** (the `asyncThrottle` describe and the fake half of
+//    the concurrency-property pair below) verify logical timing
+//    behavior — coalescing, cancellation, return-value plumbing, error
+//    propagation, gap measurement, in-process serialization, overlap
+//    detection. The assertions are about *what* runs *with what args*
+//    *in what order*; real wall-clock time is incidental, and faking it
+//    makes the tests fast and deterministic.
+//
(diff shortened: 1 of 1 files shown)


darcyparker:
@msakrejda - Can you have a look at these updates (to tests only)?  Thanks

darcyparker:
I closed/re-opened this PR to trigger tests to re-run.  There was a 'flaky' test that should pass if this re-runs.  I also created separate PR https://github.com/automerge/automerge-repo/pull/625 to fix that flaky test.

msakrejda (approved):
Nice, thanks. I agree the real timer tests are worth the additional runtime, considering it's still pretty quick and that this functionality is fundamental to what automerge-repo does. I'll give others some time to review but otherwise will merge soon.

PR #624, opened 2026-05-13
Add race test to PR 618, and improve related tests

darcyparker:
@alexjg - I reviewed PR #618.    This PR targets your branch `automerge:document-query`

- I'd like to add a test that demonstrates your PR #618 fixes a race we are experiencing 
- As well, there are 3 commits that strengthen some related tests to your PR.
    - use fake timers when appropriate
    - replace `setTimeout(0)`/`pause(0)` with microtask and macrotask flush patterns that are more idiomatic
    - and test `await flush()` instead of guessing storage-save timing
 - And use `foreverPromise` const in `DocHandle.whenReady()`. (No need to recreate new promises that never resolve)
 
@alexjg - I am adding 534af8b8e819519908e3346c5b64e4f844e603b6 and  43c0dceb0a9cde07a6fbcf73f8f411efef5fb2f9 to this as well. (last 2 commits) 
- Please see the commit messages explaining why and let me know if you have questions.  
- Splitting the current `Repo.#queries` up into `Repo.#queriesByHandle` and `#queryHandleByDocumentId` enables garbage collection.  
- Similarly I am weakening DocSYnchronizer references to help with GC
- Not for your PR, but an FYI of things I have drafted as follow ups to discuss/consider?
     - My branch https://github.com/automerge/automerge-repo/pull/612 mentions it is compatible with this PR... `darcy/weakCaches_on_PR618` branch is rebased on this PR's branch and demonstrates refactoring to use `WeakValueMap`.  This PR is setting things up to have better garbage collection / memory management.
    - Another branch `darcy/handleCacheRefactor-based-weakCaches_on_PR618`https://github.com/darcyparker/automerge-repo/tree/darcy/handleCacheRefactor-based-weakCaches_on_PR618 builds on `darcy/weakCaches_on_PR618`  (still a draft but finalizes the ideas I have on cleaning up memory management so that automerge-repo handles/caches can be GC'd; hopefully resolving memory challenges we've had)

PR #626, opened 2026-05-13
fix: drop sync messages for documents evicted mid-receive

darcyparker:
This resolves a race condition we're experiencing problems with randomly. PR #618 also resolves this race (but a different approach).  If 618 takes time to review/merge, maybe this PR can be reviewed/merged sooner as a stop gap?

CollectionSynchronizer.receiveMessage has an await on repo.find() that runs after the per-document cache lookup but before the synchronous fetchDocSynchronizer call. If removeFromCache (e.g. from a downstream LRU evictor) runs during this microtask window, the continuation captures an UNLOADED handle and installs a fresh DocSynchronizer with it. Subsequent sync messages get queued in #pendingSyncMessages and never processed — the peer is silently cut off from that document until restart.

Add a race guard after the find() await: if the handle has been unloaded or no longer matches the cached entry, drop the message and return. The peer's next attempt will create a fresh handle.

This is a stopgap targeting the legacy async receiveMessage path; the larger xstate-removal refactor in flight (#618) removes the await window structurally and obviates this guard. Lands cleanly on either.

### DIFF
--- a/packages/automerge-repo/test/CollectionSynchronizer.test.ts
+++ b/packages/automerge-repo/test/CollectionSynchronizer.test.ts
@@ -1,5 +1,6 @@
 import assert from "assert"
-import { beforeEach, describe, it } from "vitest"
+import { beforeEach, describe, it, vi } from "vitest"
+import { next as Automerge } from "@automerge/automerge"
 import { PeerId, Repo, SyncMessage } from "../src/index.js"
 import { CollectionSynchronizer } from "../src/synchronizer/CollectionSynchronizer.js"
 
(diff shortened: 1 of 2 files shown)


packages/automerge-repo/test/CollectionSynchronizer.test.ts
```diff
@@ -98,4 +99,63 @@ describe("CollectionSynchronizer", () => {
...
+        data: syncData,
+      }
+
+      // Patch repo.find to return the original handle even after eviction —
+      // this simulates the race outcome where receive's await find() had
+      // already grabbed the handle before removeFromCache unloaded it.
+      const originalFind = repo.find.bind(repo)
+      vi.spyOn(repo, "find").mockImplementation(async (id, options) => {
+        if (typeof id === "string" && id === documentId) {
+          return handle as never
```
msakrejda: Why the `as never`?

msakrejda (approved):
Looks good, thanks. I don't think we need to wait for #618. The test feels potentially brittle, but I think that's okay given the situation, especially since it's well commented. I'll give others time to review but otherwise will merge soon.

packages/automerge-repo/test/CollectionSynchronizer.test.ts
darcyparker: Good call out.  The spy did not have a correct type and was cast as `never` as a hack.  I improved this by casting to a more accurate type in 92ae17c0ee9e60e7a1ee7dc56a1e5b98eb5310f2

darcyparker:
https://github.com/automerge/automerge-repo/pull/618 merged, so this solution is no longer required.

The unit test is this PR is good because it illustrates a race that #618 happens to resolve. 
https://github.com/automerge/automerge-repo/pull/624 had this unit test for the race and some things I hoped to get into 618... but that PR was closed. 

I created `darcy/raceTest` and based of today's `main`.  And I'll resubmit. 

Update: https://github.com/automerge/automerge-repo/pull/636

PR #627, opened 2026-05-14
fix(websocket): drop sends to non-OPEN sockets instead of throwing

### LINKED ISSUE #297: WebSocket Ready flag should only be set after Open event

profiene:
Closes #297 and addresses a closely-related production crash mode.

## What

Both `WebSocketClientAdapter` and `WebSocketServerAdapter` previously called `socket.send()` without a readyState gate. When the socket was CONNECTING (issue #297), CLOSING, or CLOSED, the send threw — and the throw escaped through automerge-repo's xstate-driven event machinery as an uncaught exception with no async frame to catch it, crashing the host process.

Reproduced consistently in a hybrid sync-server spike (automerge-repo `Repo` + custom Postgres `StorageAdapter` inside a Node server) any time a client disconnected mid-sync — tab close, network blip, page nav.

## Fix

Both adapters now drop sends to non-OPEN sockets silently and log via the existing `debug` channel. The library's existing sync state and reconnect logic replays once the socket is OPEN again, so the only observable change is "no crash."

## Tests

- Updated `should throw an error if asked to send before ready` → `should silently drop sends before the socket is ready` (the throw behavior was the bug, not the spec).
- Added `should not throw when send() is called on a non-OPEN socket` for the client adapter.
- Added `should not throw when send() targets a peer whose socket is non-OPEN` for the server adapter.

Full suite (`pnpm test --run`): **690 passed, 1 skipped, 0 failed**. Prettier clean.

### DIFF
--- a/packages/automerge-repo-network-websocket/src/WebSocketClientAdapter.ts
+++ b/packages/automerge-repo-network-websocket/src/WebSocketClientAdapter.ts
@@ -179,8 +179,6 @@ export class WebSocketClientAdapter extends WebSocketNetworkAdapter {
     if (this.socket.readyState !== WebSocket.OPEN) {
       // CONNECTING / CLOSING / CLOSED. Drop silently — sync state and
       // reconnect logic will replay once the socket is OPEN again.
-      // Throwing here previously surfaced as an uncaught exception
-      // through automerge-repo's event machinery (see issue #297).
       this.#log(
         `Tried to send on a non-OPEN socket (readyState=${this.socket.readyState}); dropping.`
       )
--- a/packages/automerge-repo-network-websocket/src/WebSocketClientAdapter.ts
+++ b/packages/automerge-repo-network-websocket/src/WebSocketClientAdapter.ts
@@ -176,8 +176,16 @@ export class WebSocketClientAdapter extends WebSocketNetworkAdapter {
       this.#log("Tried to send on a disconnected socket.")
       return
     }
-    if (this.socket.readyState !== WebSocket.OPEN)
-      throw new Error(`Websocket not ready (${this.socket.readyState})`)
+    if (this.socket.readyState !== WebSocket.OPEN) {
+      // CONNECTING / CLOSING / CLOSED. Drop silently — sync state and
+      // reconnect logic will replay once the socket is OPEN again.
+      // Throwing here previously surfaced as an uncaught exception
+      // through automerge-repo's event machinery (see issue #297).
+      this.#log(
+        `Tried to send on a non-OPEN socket (readyState=${this.socket.readyState}); dropping.`
+      )
+      return
+    }
 
     const encoded = cbor.encode(message)
     this.socket.send(toArrayBuffer(encoded))
--- a/packages/automerge-repo-network-websocket/src/WebSocketServerAdapter.ts
+++ b/packages/automerge-repo-network-websocket/src/WebSocketServerAdapter.ts
@@ -116,6 +116,18 @@ export class WebSocketServerAdapter extends NetworkAdapter {
       return
     }
 
+    if (socket.readyState !== WebSocket.OPEN) {
+      // Peer's socket is CONNECTING / CLOSING / CLOSED. Drop silently —
+      // the underlying `ws.send()` would throw, and the throw escapes
+      // through automerge-repo's event machinery as an uncaught
+      // exception that crashes the process. Once the peer reconnects
+      // and rejoins, normal sync state resumes.
+      log(
+        `Tried to send to peer ${message.targetId} with non-OPEN socket (readyState=${socket.readyState}); dropping.`
+      )
+      return
+    }
+
     const encoded = encode(message)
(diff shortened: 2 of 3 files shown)


packages/automerge-repo-network-websocket/src/WebSocketClientAdapter.ts
```diff
@@ -176,8 +176,16 @@ export class WebSocketClientAdapter extends WebSocketNetworkAdapter {
...
       return
     }
-    if (this.socket.readyState !== WebSocket.OPEN)
-      throw new Error(`Websocket not ready (${this.socket.readyState})`)
+    if (this.socket.readyState !== WebSocket.OPEN) {
+      // CONNECTING / CLOSING / CLOSED. Drop silently — sync state and
+      // reconnect logic will replay once the socket is OPEN again.
+      // Throwing here previously surfaced as an uncaught exception
+      // through automerge-repo's event machinery (see issue #297).
+      this.#log(
```
msakrejda: This will need to be updated once #621 is merged (I plan to merge that first since it came first).

packages/automerge-repo-network-websocket/src/WebSocketClientAdapter.ts
```diff
@@ -176,8 +176,16 @@ export class WebSocketClientAdapter extends WebSocketNetworkAdapter {
       this.#log("Tried to send on a disconnected socket.")
       return
     }
-    if (this.socket.readyState !== WebSocket.OPEN)
-      throw new Error(`Websocket not ready (${this.socket.readyState})`)
+    if (this.socket.readyState !== WebSocket.OPEN) {
+      // CONNECTING / CLOSING / CLOSED. Drop silently — sync state and
+      // reconnect logic will replay once the socket is OPEN again.
+      // Throwing here previously surfaced as an uncaught exception
+      // through automerge-repo's event machinery (see issue #297).
```
msakrejda:
```suggestion
      // CONNECTING / CLOSING / CLOSED. Drop silently — sync state and
      // reconnect logic will replay once the socket is OPEN again.
```
I think git history and the PR description and discussion are enough for what
happened here previously and why it was changed. We should limit the comment
to what the code is doing now.

packages/automerge-repo-network-websocket/src/WebSocketServerAdapter.ts
```diff
@@ -116,6 +116,18 @@ export class WebSocketServerAdapter extends NetworkAdapter {
       return
     }
 
+    if (socket.readyState !== WebSocket.OPEN) {
+      // Peer's socket is CONNECTING / CLOSING / CLOSED. Drop silently —
+      // the underlying `ws.send()` would throw, and the throw escapes
+      // through automerge-repo's event machinery as an uncaught
+      // exception that crashes the process. Once the peer reconnects
+      // and rejoins, normal sync state resumes.
```
msakrejda: Similarly to above, I think we should just say
```suggestion
      // Peer's socket is CONNECTING / CLOSING / CLOSED. Drop silently —
      // once the peer reconnects and rejoins, normal sync state resumes.
```

packages/automerge-repo-network-websocket/test/Websocket.test.ts
```diff
@@ -174,7 +174,13 @@ describe("Websocket adapters", () => {
...
     })
 
-    it("should throw an error if asked to send before ready", async () => {
+    it("should silently drop sends before the socket is ready", async () => {
+      // Earlier versions of the adapter threw a "Websocket not ready"
+      // error here. That throw surfaced as an uncaught exception via
+      // automerge-repo's event machinery, crashing the host process
+      // (see issue #297). The current contract is to drop the message
+      // silently; sync state and reconnect logic will replay once the
+      // socket is OPEN.
```
msakrejda, line 183: I suggested dropping similar comments in the implementation, but given that this test is primarily guarding against a regression, I think this particular comment is useful and should stay.

msakrejda (approved):
Thanks! I left some minor comments, but looks good. I'll leave it open for a bit for others to review, but otherwise will merge soon.

PR #636, opened 2026-05-18
chore: add test for race that was fixed in PR618 and strengthen related tests 

darcyparker:
- Adds test to assert the `receiveMessage` / `removeFromCache` race is closed
    - This bug was resolved by PR #618, but there was no unit test to demonstrate it resolved it
    - Adding the test guards against this type of race occurring in future
- Minor: Refactors `DocHandle.ts` to use `foreverPromise` rather than re-create a promise that never resolves (which was slow memory leak)
- Minor: Removes unused `import` in `Repo.ts`
- Strengthen tests in `Websocket.test.ts` to use `fake` timers (and keeps real clock tests /w reason why).  (This reduces risk of flaky test and has side benefit of reducing time from ~500ms to 25ms)
- Replace `setTimeout(0)`/`pause(0)` with more idiomiatic JS microtask flush patterns. No functional changes

Note: This supersedes #624

### DIFF
--- a/packages/automerge-repo-network-websocket/test/Websocket.test.ts
+++ b/packages/automerge-repo-network-websocket/test/Websocket.test.ts
@@ -572,7 +572,7 @@ describe("Websocket adapters", () => {
         const handle = silentRepo.create()
         handle.update(() => A.clone(doc))
         const { documentId } = parseAutomergeUrl(handle.url)
-        await pause(150)
+        await silentRepo.flush([documentId])
         return {
           url: handle.url,
           doc,
--- a/packages/automerge-repo-network-websocket/test/Websocket.test.ts
+++ b/packages/automerge-repo-network-websocket/test/Websocket.test.ts
@@ -18,7 +18,7 @@ import * as CBOR from "cbor-x"
 import { once } from "events"
 import http from "http"
 import { getPortPromise as getAvailablePort } from "portfinder"
-import { describe, it } from "vitest"
+import { afterEach, describe, it, vi } from "vitest"
 import WebSocket from "ws"
 import { WebSocketClientAdapter } from "../src/WebSocketClientAdapter.js"
 import { WebSocketServerAdapter } from "../src/WebSocketServerAdapter.js"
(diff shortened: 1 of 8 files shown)


msakrejda (approved):

msakrejda:
@darcyparker the first test run after I merged to [main failed](https://github.com/automerge/automerge-repo/actions/runs/26430191878/job/77801681477). At a glance, I don't think that's related to your changes, and a restart resolved the issue, but if you could confirm when you have a chance, that'd be great.

darcyparker:
> @darcyparker the first test run after I merged to [main failed](https://github.com/automerge/automerge-repo/actions/runs/26430191878/job/77801681477). At a glance, I don't think that's related to your changes, and a restart resolved the issue, but if you could confirm when you have a chance, that'd be great.

Thanks for merging. I agree, I don't think its related to my changes. 

I did some analysis with AI...

> #Root cause
> The test makes a state-mutating call outside of act():
```typescript

render(<Wrapped />, { wrapper })

// OUTSIDE act — this is the flaky bit
handleA.change(doc => (doc.counter = 1))

await act(async () => {
  await Promise.resolve()
})
```

> Every other change-propagation test in `useDocuments.test.tsx` wraps its change() call in `act()` — see line 50, line 88-98, line 184-191. Only `useDocuments.test.tsx` line 402 does not.
> 
> Why this is flaky: `useDocuments` re-attaches its change listener every time `handleMap` changes by reference. In a healthy render cycle, `useDocHandles`'s effect calls `setHandleMap(nextHandleMap)`, which schedules a re-render and causes `useDocuments`'s effect to tear down and re-attach. There's a window in concurrent React where:

> - `render()` returns
> - The listener hasn't yet been attached (or is mid-cleanup)
> - `handleA.change(…)` fires the change event synchronously
> - Nobody is listening
> - The `await act(…)` block flushes the pending state but there's no `setDocMap` to flush

> Locally, the timing usually settles before render() returns. On CI under load, it occasionally doesn't.

> The test name "should handle document changes during loading" is also a bit of a fib — setup() creates a local handle that's synchronously ready, so there's no actual loading. The test is really just "does a change after mount propagate."

I'll create a PR to strengthen this test.

darcyparker:
https://github.com/automerge/automerge-repo/pull/650 @msakrejda FYI

PR #637, opened 2026-05-18
refactor: refactor caches to improve garbage collection and prevent memory leaks

darcyparker:
Builds on https://github.com/automerge/automerge-repo/pull/636 and https://github.com/automerge/automerge-repo/pull/618

The objective of this PR is to reduce a class of subtle memory leaks. Today, there are issues with tightly held references across coupled code in automerge-repo that prevent garbage collection.  The existing `Repo.removeFromCache()` has some edge cases where memory is not released because of some tightly held references.

This PR introduces a new mental model for the automerge-repo user: holding a strong reference keeps the document loaded; dropping the references lets the repo release its coordination state automatically.  `Repo.removeFromCache()` can still be used, but is unnecessary if the automerge-repo user drops reference.

The commit messages provide more details about the changes. The following is an abridged summary:

- New `WeakValueMap` encapusulates the existing `Map<string, WeakRef>` pattern already in place in `#refCache`  This allows the pattern to be reused with additional caches.  `refCache` and `viewCache` now use `WeakValueMap`.
    - The new `WeakValueMap` also affords cleanup of dead string keys
- Breaks up `Repo.#queries` into `Repo.#queriesByHandle` and `Repo.#querHandleByDocumentId`.  This enables Garbage Collection
- Weakens the `DocSynchronizer` references (which used to be strong refs) so that dropping the handle allows garbage collection
- Adds helpers for unit tests to do adaptive GC polling that demonstrates the weak refs are properly garbage collected
- `Repo` now has `FinalizationRegistry` to ensure cleanup of per-doc coordination state (that was coupled)
    - Adds unit tests to demonstrate consumer-driven cleanup per doc occurs
    - Adds tests for cross-cutting cases to guard against regressions

### DIFF
--- a/packages/automerge-repo/src/presence/Presence.ts
+++ b/packages/automerge-repo/src/presence/Presence.ts
@@ -184,6 +184,8 @@ export class Presence<
    * {@link https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilitychange_event | "visibilitychange"}
    * to stop sending and receiving updates when not active.
    *
+   * @remarks here?
+   * 
    * **You must call `stop()` to release a `Presence` instance.** Dropping the
    * reference is not enough: the heartbeat / pruning `setInterval` timers
    * scheduled in {@link start} are registered with the host timer queue,
--- a/packages/automerge-repo/src/presence/Presence.ts
+++ b/packages/automerge-repo/src/presence/Presence.ts
@@ -183,6 +183,19 @@ export class Presence<
    * or
    * {@link https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilitychange_event | "visibilitychange"}
    * to stop sending and receiving updates when not active.
+   *
+   * **You must call `stop()` to release a `Presence` instance.** Dropping the
+   * reference is not enough: the heartbeat / pruning `setInterval` timers
+   * scheduled in {@link start} are registered with the host timer queue,
+   * which is an external GC root. As long as those intervals are running,
+   * the timer callbacks (which capture `this`) keep this `Presence` alive,
+   * which in turn keeps its `DocHandle` alive via `this.#handle`.
+   *
+   * In a browser this is usually a non-issue because `Presence` is tied to
+   * a long-lived page; pair `stop()` with `pagehide` / `visibilitychange`
+   * for clean teardown. In a longer-running JS process that creates and
+   * tears down many `Presence` instances over time, forgetting `stop()`
+   * leaks per-instance state and pins the associated handle.
    */
   stop() {
     if (!this.#running) {
--- a/packages/automerge-repo/src/refs/ref.ts
+++ b/packages/automerge-repo/src/refs/ref.ts
@@ -72,6 +72,9 @@ export class RefImpl<
    * field-side reference to the original arrow function so it can be
    * GC'd.
    *
+   * 
+   * @remarks
+   * 
    * Optional. If you drop both the handle and the Ref together, the
    * natural internal cycle handles cleanup. Call `dispose()` only
    * when you want to release a Ref while keeping the handle alive —
--- a/packages/automerge-repo/src/refs/ref.ts
+++ b/packages/automerge-repo/src/refs/ref.ts
@@ -2,6 +2,7 @@ import * as Automerge from "@automerge/automerge/slim"
 import type { Doc, Prop, Heads } from "@automerge/automerge/slim"
 import type { DocHandle, DocHandleChangePayload } from "../DocHandle.js"
 import { encodeHeads, decodeHeads } from "../AutomergeUrl.js"
+import { noop } from "../helpers/noop.js"
 import type {
   Segment,
   PathSegment,
@@ -21,14 +22,10 @@ import type { CursorMarker } from "./types.js"
 import { stringifyRefUrl } from "./parser.js"
 import { MutableText } from "./mutable-text.js"
 
-/**
- * FinalizationRegistry for automatic cleanup of Ref instances.
- * This ensures subscriptions are cleaned up when Refs are garbage collected,
- * even if dispose() is never called.
- */
-const refCleanupRegistry = new FinalizationRegistry<() => void>(cleanup =>
-  cleanup()
-)
+// Refs and their handles form an internal cycle (handle → events →
+// listener → this → docHandle → handle) that's collected as a unit
+// when both are dropped — no FinalizationRegistry needed. For
+// releasing a Ref while keeping the handle alive, use `dispose()`.
 
 /**
  * Internal implementation of the Ref interface.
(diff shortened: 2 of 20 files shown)


darcyparker:
Closing and reopening because of flaky test (unrelated to this PR) https://github.com/automerge/automerge-repo/actions/runs/26042795669/job/76558741575?pr=637 that I'll fix in separate PR. https://github.com/automerge/automerge-repo/pull/638

darcyparker:
Side note: Tests are failing because of unrelated flaky test. https://github.com/automerge/automerge-repo/pull/650 fixes the flaky test.

darcyparker:
Side note: Another flaky test is fixed in https://github.com/automerge/automerge-repo/pull/651
- #650 and #651 were cherry-picked into this PR to avoid the flaky tests failing

packages/automerge-repo/src/presence/Presence.ts
```diff
@@ -183,6 +183,19 @@ export class Presence<
    * or
    * {@link https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilitychange_event | "visibilitychange"}
    * to stop sending and receiving updates when not active.
+   *
```
msakrejda, line 186: Should we maybe add
```suggestion
   *
   * @remarks here?
   * 
```

packages/automerge-repo/src/refs/ref.ts
```diff
@@ -21,14 +22,10 @@ import type { CursorMarker } from "./types.js"
...
- * This ensures subscriptions are cleaned up when Refs are garbage collected,
- * even if dispose() is never called.
- */
-const refCleanupRegistry = new FinalizationRegistry<() => void>(cleanup =>
-  cleanup()
-)
+// Refs and their handles form an internal cycle (handle → events →
+// listener → this → docHandle → handle) that's collected as a unit
+// when both are dropped — no FinalizationRegistry needed. For
+// releasing a Ref while keeping the handle alive, use `dispose()`.
```
msakrejda, line 28: Hmm. Previously refs would be cleaned up automatically through finalization, but now we would require manual cleanup. I know the caveats of relying on finalization, and this is a more principled way of handling it, but can we safely make this change? @alexjg @OrionReed any thoughts?

packages/automerge-repo/src/refs/ref.ts
```diff
@@ -67,13 +64,31 @@ export class RefImpl<
...
   }
 
-  #cleanup(): void {
+  /**
+   * Detach this Ref from its `DocHandle`: remove the internal
+   * `"change"` listener and any `onChange` callbacks, and release the
+   * field-side reference to the original arrow function so it can be
+   * GC'd.
+   *
+   * Optional. If you drop both the handle and the Ref together, the
```
msakrejda, line 78:
```suggestion
   * 
   * @remarks
   * 
   * Optional. If you drop both the handle and the Ref together, the
```

packages/automerge-repo/src/refs/types.ts
```diff
@@ -375,6 +375,22 @@ export interface Ref<TValue = unknown> {
     ) => void
   ): () => void
 
+  /**
+   * Detach this Ref from its `DocHandle`: remove the internal `change`
+   * listener and any `onChange` callbacks.
+   *
+   * Optional. If you drop both the handle and the Ref together, the
```
msakrejda, line 385:
```suggestion
   * 
   * @remarks
   *   
   * Optional. If you drop both the handle and the Ref together, the
```

msakrejda (commented):
In general, I think this makes a lot of sense. I'm not sure about removing the ref finalization is a good idea: couldn't that cause leaks for users churning through lots of refs for a single handle without disposing them manually? It might still be a good idea—I don't anticipate that's a common use case and I do like your design.

I'm also not sure how this is impacted by #605 and #640 (which I think is #605's successor).

I'll wait for others to weigh in.

packages/automerge-repo/src/refs/ref.ts
darcyparker: Good question. The short version: the old `FinalizationRegistry` wasn't actually cleaning anything up. `FinalizationRegistry` is a genuinely tricky tool to get right — MDN itself recommends [avoiding it where possible](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry#avoid_where_possible). The original use was a reasonable attempt at automatic cleanup; it was only when I dug into the specific leaks I was chasing that it became clear this setup couldn't do what it intended.

The details: `RefImpl` registered `() => this.#cleanup()` as the registry's *held value*, and that closure captured `this`. A registry keeps its held value alive for as long as the target is alive, so this left a permanent strong reference back to the Ref. That meant the Ref could never be garbage-collected and the cleanup callback could never fire. And because the Ref could never be collected, it also pinned `Ref.docHandle`, so every `DocHandle` that ever had a `.ref()` created from it stayed alive for the whole lifetime of the `Repo` — a real leak.

So we're not giving up working automatic cleanup — we're removing cleanup that wasn't actually firing (and was pinning memory in the process).

What replaces it is still automatic in the common case:

- **You drop a Ref and you're also done with its handle** → the Ref and handle form an internal cycle (`listener → this → docHandle → handle → events → listener`). Nothing outside points at it, so the garbage collector frees the whole thing together. No `dispose()` needed.
- **You keep the handle but want to release a Ref early** → this is the one case where you call `dispose()`. While you're holding the handle, its Refs are still reachable through `handle.events → listener → ref`, so they won't be collected on their own. `dispose()` unsubscribes that Ref's listener right away instead of waiting for the handle to go away.

So: if you're holding a Ref it stays (you're using it); if you drop a Ref and its handle it's collected for you; only "keep the handle, drop its Refs" needs `dispose()` — and the earlier registry didn't help that case either. Happy to document that clearly.

One thing worth carrying into the rework: a `FinalizationRegistry` only works when it holds its target *weakly* — which is exactly what the new subdoc-handles registry does (`WeakRef<DocHandle>` + a `FinalizationRegistry`). So removing the registry here and using one correctly there are really the same idea, just applied where it can actually fire.

darcyparker:
@msakrejda — thank you for the initial review!

I took a closer look at merging `main` into this branch, and the conflicts run deeper than they first appear. This branch and the **subdoc-handles** work that recently landed on `main` are effectively **two independent rewrites of the same core files** (`DocHandle`, `Repo`, `DocSynchronizer`). A textual merge would bolt this branch's GC model onto a structure that was rebuilt underneath it — producing something incoherent I'd have to untangle anyway. So this needs more care to rework than a conflict-resolution pass.

Two concrete observations from the merge attempt:

- **subdoc-handles already moves in the right direction** — its per-document registry caches handles via `WeakRef` + a `FinalizationRegistry` (more on that in [my note on the FinalizationRegistry thread](https://github.com/automerge/automerge-repo/pull/637#discussion_r3327347741)).
- **But the leak this PR targets isn't addressed**: `Repo.#queries` is still a strong `Record<DocumentId, DocumentQuery>`, and each query holds its handle strongly — so handles stay pinned for the repo's lifetime regardless of the weak registry. That's the sync-server memory leak this PR set out to fix.

Given that, I think the best move is to **close this PR and re-create it as a focused rework on top of current `main`**, rather than force the merge. But before I do, I'd really value @alexjg @OrionReed (and anyone close to the lifecycle code) weighing in on the **spirit** of this PR first — so we land on a shared plan and don't step on each other's toes.

A few options / things I'd like input on:

1. **Decouple the independently-useful pieces** that could land right away, ahead of the bigger change — e.g. the `WeakValueMap` helper and the memory-lifetime tests.
2. **The holistic GC rework** — GC-able `Repo` query storage + `FinalizationRegistry` cleanup, re-expressed against the new `Document`/registry model — really needs to be one coherent change. I'd like to agree on the approach + ownership/timing so it doesn't collide with in-flight core work, then land it in one well-scoped PR.
3. **The end goal I care most about**: adopt tests like the ones here that *demonstrate the leaks are actually resolved* on the new structure, and make sure the lifecycle best-practices this PR reaches for are understood and agreed.

This touches enough of the core lifecycle that I'd rather coordinate closely than risk colliding with in-flight work — **happy to hop on a call** to talk through the spirit and a plan. Mostly I want to avoid two of us reworking the same internals in parallel; if we can agree on the approach and rough ownership/sequencing first, I'll take a clean, well-scoped run at it.

None of this work is lost — it's the spec for the rework. Happy to write up a detailed plan (mapping the changes here onto the new structure) once I have your thoughts on direction.

darcyparker:
Take 2 of this work is now open, rebuilt against the subdoc-handles architecture:

- #736 groundwork (WeakValueMap, weak per-heads view cache, GC test infrastructure, standalone memory fixes)
- #737 the main change: all per-document coordination state weakly held, so a document stays loaded exactly while something observes it and is collected once nothing does
- #735 a synthetic sync-server memory benchmark, run byte-identical before and after, with a chart: heap flat at ~27 MB vs climbing past 49 MB, and zero documents held between sessions vs every document ever relayed
- follow-ups so consumers benefit too: #738 (storage-nodefs), #739 (react-hooks), #693 (svelte-store)

@msakrejda and @alexjg - On the `dispose()` concern from https://github.com/automerge/automerge-repo/pull/637#discussion_r3327347741: the new design needs no `dispose()` at all. Each document's objects already form a self-referential cluster that the Repo reaches only weakly, and "keep alive while observed" is a small listener-driven refcount, so dropping references is always enough, and a handle with listeners attached keeps working (the contract this PR would have broken). Hopefully the before/after charts in #735 motivates a look at the details.

PR #640, opened 2026-05-19
Rewrite subdoc handles on top of 'document-query' PR

OrionReed:
Continuation of #605 based on the recent changes to DocHandles in main.


pvh:
Ok! Lots and lots of updates this evening. Thanks to @OrionReed for strong core concept & design -- the Trie for patch distribution looks very promising in particular.

I've cleaned up a lot of loose ends: view caching, cursor resolving, consistency with various value types. I've tried to be thorough and keep my patch clean and not too LLM-bloaty.

A few things that could use a second or third opinion. 

Because we can have leaf values in docUrls, we need a sensible way of handling strings that lets us target them for splice & updateText. I've iterated Orion's MutableText proxy to just extend String. Both approaches are honestly pretty ugly, but I think I come down slightly on the side of extending string as being less weird. @alexjg, wdty? Also -- as I'm writing this, I'm thinking about other "leafy values" like Counter, RichText, ImmutableText and so on... I haven't tested or thought about those. Maybe they're fine? Worth checking for tests.

Next, history() on subdoc handles is unreasonably, unusably expensive. This will break things and makes subdoc handles unusable for a bunch of things we might want to do, like passing a subpath to a code document to an editor. I think we might need some help from @orionz.

I'm not super happy with the URL syntax. It's idiosyncratic but not (as far as I can tell) very thoughtfully designed. I wish it were a subset of some other more robust language, or closer to an existing syntax to reduce the mental burden of adopting it. But it's not and I don't have the energy to think more about it. I can say that at least a shallow interrogation of several other extant query languages (JSONPointer, JSONPath, xpath, CSS selectors, jq, &c) did not turn up an alternative that was obviously worth switching to.

In the original patch there is a bunch of likely LLM comment rewriting that is gratuitous and can probably be cut. I haven't done that. I don't know if it's necessary.

My recommendation is -- pending review -- to cut a prerelease on a preview tag. We should obviously merge subductionjs underneath that since we want it in patchwork.

All of this has me really eager to redesign a dochandle that doesn't rely on .doc() and rethink these apis. A property accessor should return a subdoc handle, I think! But this is a problem for... not tonight.

With that -- I leave this in your hands, @alexjg and @OrionReed. Paul is counting on us...

packages/automerge-repo/src/Document.ts
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+ * Per-document shared state - one per `documentId`, referenced by every
+ * `DocHandle` (root, sub, view) into that document. Owns the Automerge
+ * snapshot and the {@link HandleRegistry} (identity, pattern resolution,
+ * listeners, dispatch). Not part of the public API.
+ *
+ * @internal
+ */
+export class Document<T = unknown> {
```
alexjg, line 17: Do we really need the `Document` class? It seems to me like both the `HandleRegistry` and the `syncInfoLookup` could be stored directly on the `DocHandle` and then we wouldn't need `Document`. This would be nice because it's confusing t0 me to have a thing called `Document` which isn't the automerge document.

We do pass the `Document` to the `HandleRegistry` constructor but it appears to be only used to emit a log?

alexjg (approved):

packages/automerge-repo/src/Document.ts
pvh: Hmm! Good spot. I'll give it a think.

PR #649, opened 2026-05-25
SubductionSource: adopt fragments API + per-tick bundling + GC fixes

pvh:
## Summary

Three commits on top of `subductionjs` that rebuild `SubductionSource` around `automerge@3.3.0-fragments.1` and the new fragments API. Bench numbers below show the impact on a representative workload set.

### 1. `Adopt Automerge fragments API in SubductionSource`
- Outbound: replaces hand-rolled `getChangesMetaSince` / per-change `getChangeByHash` bookkeeping with `Automerge.getCommits(doc)` + `Automerge.getFragments(doc)`, shipped together in one `subduction.addBatch(...)`. Wire-level transition from "N LooseCommits" to "one Fragment bundle" now follows automerge core's compaction.
- Inbound: per-entry `pendingInbound` queue flushed once per microtask via a single concatenated `Automerge.loadIncremental`. Bursts of 1000+ live commits collapse into one handle update.
- Drops `recentlySavedHashes` and the old per-change `meta` indirection.
- Preserves the `BlobInterceptor` (E2EE, #646) at all four use sites.
- Bumps `@automerge/automerge` to `3.3.0-fragments.1`.

### 2. `SubductionSource: bundle only newly-formed fragments per save tick`
- `getCommits` / `getFragments` eagerly bundle bytes for every commit/fragment in the doc; calling them on every throttled save tick was O(N × #items). Measured: `getFragments(doc)` ~31s at 12k changes, ~52s at 15k, with WASM-heap overflows around 12k–15k.
- Switches to `getFragmentMetadata` + filter-by-`knownHashes` + bundle-only-new. Per-tick work becomes proportional to newly-formed items.

### 3. `SubductionSource: trigger compaction on inbound and hydrated data`
- Compaction was only scheduled from `#saveNewCommits`. Fixed two silent failures:
  - Read-only peers never enter `#saveNewCommits` and accumulate forever-superseded fragments.
  - Fresh-process boots start with empty `persistedCommitHashes`/`persistedFragmentHashes` and can't GC defunct prior-process records.
- `#flushInbound` and `#loadBlobsIntoHandle` now schedule compaction after `loadIncremental`. `attach()` seeds the persisted-hash sets from `listCommitIds`/`listFragmentIds` and kicks one pass.

## Bench impact (`subduction.22` → this branch, direct topology, 3 iter + 1 warmup)

| scenario | metric | before | after | Δ |
|---|---|---|---|---|
| `bigCommit × subduction` | net.bytesOut | 1.91 MiB | 5.7 KiB | **−99.7%** |
| `bigCommit × subduction` | storage.bytes | 7.64 MiB | 20.0 KiB | **−99.7%** |
| `wideCollab × subduction` | propagation | 7065 ms | 2787 ms | **−60.6%** |
| `wideCollab × subduction` | wallClock | 8937 ms | 4664 ms | **−47.8%** |
| `wideCollab × subduction` | net.bytesOut | 10.73 MiB | 5.06 MiB | **−52.8%** |
| `wideCollab × subduction` | storage.bytes | 6.16 MiB | 2.36 MiB | **−61.8%** |
| `partialSync × subduction` | incrementalSync | 53.6 s | 3.52 s | **−93.4%** |
| `partialSync × subduction` | initialSync | 7.03 s | 3.66 s | **−48.0%** |
| `partialSync × subduction` | net.bytesOut | 24.51 MiB | 2.56 MiB | **−89.6%** |
| `partialSync × subduction` | storage.bytes | 22.20 MiB | 2.63 MiB | **−88.1%** |

Run against registry `@automerge/automerge@3.3.0-fragments.1` + `@automerge/automerge-subduction@0.13.0` on both sides — so this isolates the contribution of the `automerge-repo` JS changes alone (no local Rust crate dependencies).

`deepHistory × subduction` also notable: on `subduction.22` it times out at 5 min; on this branch it propagates in ~2.6 s.

## Test plan

- [x] `pnpm --filter @automerge/automerge-repo build` — clean
- [x] `automerge-repo-bench` against the branch — bigCommit + partialSync + wideCollab all converge in both classic and subduction modes (results above)
- [ ] CI test suite (`pnpm test` config issue with vitest 4 on `subductionjs` HEAD is pre-existing; not introduced here)

🤖 Generated with [Cursor](https://cursor.com)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/automerge-repo/src/subduction/source.ts
+++ b/packages/automerge-repo/src/subduction/source.ts
@@ -985,21 +985,61 @@ export class SubductionSource implements DocumentSource {
     // higher-level fragments. Automerge core owns the compaction
     // policy (which commits get absorbed into which fragments); we
     // just mirror that view into subduction. Once a fragment forms,
-    // its component commits stop appearing in `getCommits` and the
-    // fragment itself appears in `getFragments` — so the wire-level
+    // its component commits stop appearing at level 0 and the
+    // fragment itself appears at level ≥ 1 — so the wire-level
     // forward switches from "N LooseCommits" to "one Fragment bundle"
     // automatically.
-    const docCommits = Automerge.getCommits(doc)
-    const docFragments = Automerge.getFragments(doc)
+    //
+    // We deliberately avoid `Automerge.getCommits` / `getFragments`
+    // here: those eagerly bundle the bytes for EVERY commit and
+    // fragment in the doc, and bundling is O(N) per item — so calling
+    // them on every throttled save gives O(N × #items) work on each
+    // tick, of which we then throw 99% away by filtering against
+    // `knownHashes`. At ~12k changes that's ~30 s of wasted bundle
+    // work per tick, and the resulting allocation churn inside
+    // automerge_wasm reliably overflows its 4 GiB linear-memory heap
+    // (panics in `slab/writer.rs` / `change/collector.rs`) somewhere
+    // around 12k–15k changes.
+    //
+    // Instead, we read just the metadata (cheap: a few ms even at
+    // 12k changes), filter against `knownHashes` first, and bundle
+    // only the small handful of newly-formed commits/fragments per
+    // tick. See proposals/bundle-fragments-single-walk.md for the
+    // matching automerge-core fix.
+    const commitMetas = Automerge.getFragmentMetadata(doc, 0)
+    const fragmentMetas = Automerge.getFragmentMetadata(doc, { start: 1 })
 
-    const newCommits = docCommits.filter(c => !entry.knownHashes.has(c.head))
-    const newFragments = docFragments.filter(
-      f => !entry.knownHashes.has(f.head)
+    const newCommitMetas = commitMetas.filter(
+      m => !entry.knownHashes.has(m.head)
+    )
+    const newFragmentMetas = fragmentMetas.filter(
+      m => !entry.knownHashes.has(m.head)
     )
 
-    if (newCommits.length === 0 && newFragments.length === 0) {
+    if (newCommitMetas.length === 0 && newFragmentMetas.length === 0) {
       return { prepFailures: 0 }
     }
+
+    const newCommitBytes =
+      newCommitMetas.length === 0
+        ? []
+        : Automerge.bundleFragmentMetadata(doc, newCommitMetas)
+    const newFragmentBytes =
+      newFragmentMetas.length === 0
+        ? []
+        : Automerge.bundleFragmentMetadata(doc, newFragmentMetas)
+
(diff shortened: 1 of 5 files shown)


packages/automerge-repo/src/subduction/source.ts
```diff
@@ -899,121 +1029,285 @@ export class SubductionSource implements DocumentSource {
...
+    // automerge_wasm reliably overflows its 4 GiB linear-memory heap
+    // (panics in `slab/writer.rs` / `change/collector.rs`) somewhere
+    // around 12k–15k changes.
+    //
+    // Instead, we read just the metadata (cheap: a few ms even at
+    // 12k changes), filter against `knownHashes` first, and bundle
+    // only the small handful of newly-formed commits/fragments per
+    // tick. See proposals/bundle-fragments-single-walk.md for the
+    // matching automerge-core fix.
+    const commitMetas = Automerge.getFragmentMetadata(doc, 0)
```
pvh, line 1059: @orionz this is the part with the slightly weird API

packages/automerge-repo/src/subduction/source.ts
```diff
@@ -84,7 +67,60 @@ interface SedimentreeEntry {
...
+   *    persisted on our behalf (typically inbound from peers).
+   *
+   * Drained by `#compactAbsorbed`: after every successful save we
+   * diff this set against `Automerge.getCommits(doc)` and `remove`
+   * any hash that's no longer reported by automerge (i.e. has been
+   * absorbed into a fragment). The disk footprint then shrinks to
+   * track the level-0 layer, not the full history of commits we
+   * ever wrote.
+   */
+  persistedCommitHashes: Set<string>
```
alexjg, line 105: I think we can get away with a single `persistedHashes` set here rather than one for each of commits and fragments.

packages/automerge-repo/src/subduction/source.ts
```diff
@@ -353,8 +391,16 @@ export class SubductionSource implements DocumentSource {
       mgr.onChange(() => this.#recompute())
     }
 
-    this.#storage.on("commit-saved", this.#handleDataFound.bind(this))
-    this.#storage.on("fragment-saved", this.#handleDataFound.bind(this))
+    this.#storage.on("commit-saved", (sid, commitId, blob) => {
+      const entry = this.#entries.get(sid.toString())
+      if (entry) entry.persistedCommitHashes.add(commitId.toHexString())
```
alexjg, line 396: If we had a single `persistedHashes` set then this logic could live in `#handleDataFound`

alexjg (changes requested):
Can we get the robot to add some integration tests?

PR #650, opened 2026-05-26
test: stabilize useDocuments "changes during loading" test

darcyparker:
Addresses flaky test: https://github.com/automerge/automerge-repo/pull/636#issuecomment-4544741530

`handle.change()` ran outside `act()` and the assertion relied on a single microtask flush. Under load the change event could fire while `useDocuments` was mid-re-attach of its change listener, leaving the `docMap` stale. Wrap the change in `act()` like every other change-propagation test in this file, and `waitFor` the assertion to tolerate extra microtasks.

Verified with 50/50 local runs.


msakrejda (approved):
Looks great, thanks.

PR #651, opened 2026-05-26
fix: flaky repo test

darcyparker:
https://github.com/automerge/automerge-repo/actions/runs/26454774295/job/77885482020?pr=637

I believe this test was flaky because it used `createConnectedPair()` with `{latency = 0}` default, and `pause(0)` is a macro task (not micro task) that caused event-loop starvation under rount-trip-heavy tests.   For `{latency = 0}`, it is better to have `micro` task for 0 latency case.


msakrejda (approved):
Looks great! Thanks.

PR #653, opened 2026-05-29
chore: modernize the build & dev tooling for 2026

darcyparker:
A **tooling-only modernization** — no changes to library runtime behavior or public API semantics (a few *additive* type-only exports for docs aside). It brings the monorepo's toolchain, dependencies, and CI to sensible 2026 standards, on a foundation that makes the *next* round of upgrades cheap.

Individual commits carry the detailed rationale, measurements, and gotchas; this is the map. The branch is merged up to date with `main` (currently v2.6.0-alpha.1, including the subdoc-handles work), so the diff and metrics below are against current `main`.

## Why, and in what order

The work is sequenced so each step unlocks the next:

1. **pnpm 10 → 11 first** — the foundation: the **`catalog:` protocol** (one source of truth for shared dependency versions) and a **supply-chain defense** (`minimumReleaseAge`), which everything afterward builds on.
2. **TypeScript 5.4 → 6**, positioning for a trivial bump to **TypeScript 7** (~June 2026).
3. **Vite 7 → 8** and the rest of the dependency surface brought current.
4. **prettier → oxfmt** and **eslint → oxlint** — a much faster, unified Rust-based lint/format toolchain.
5. Cleared **all `pnpm audit` advisories** and modernized the **create-\*** scaffolders.

### Bundle size — automerge-repo was being bundled into every adapter

`react-hooks` and `solid-primitives` weren't externalizing `@automerge/automerge-repo`, so each shipped a **full inlined copy** of it (react-hooks also inlined automerge's wasm). Beyond the bloat, this is a **dual-instance hazard**: a consumer app pulling in two copies of automerge-repo gets `DocHandle`/`instanceof` mismatches. Externalizing all `@automerge/*` (so the consumer supplies one shared copy) shrinks the published bundles dramatically:

| `dist/index.js` | main | this branch | reduction |
|---|--:|--:|--:|
| `@automerge/automerge-repo-react-hooks` | 244.7 kB | **48.2 kB** | **−80%** |
| `@automerge/automerge-repo-solid-primitives` | 48.7 kB | **6.3 kB** | **−87%** |

### Tarball footprint — stop shipping source noise, start shipping debuggable maps

Published packages were shipping `src/`, `test/`, `fuzz/`, and config files (no `files` allow-list). Scoped to `dist` + `src` and added **source maps**, so consumers debug into real `.ts`:

| package | main | this branch |
|---|--:|--:|
| `@automerge/automerge-repo` | 293 files / 271 kB | 341 files / 240 kB |

> Note, even with the addition of source maps, it is smaller because tests and fuzz are not included (and the extra files are the per-file `.js.map`s).

## Dependencies removed

Where a modern native API exists, the dependency was dropped outright — and a couple of dead ones were found along the way:

| Removed | Replaced by |
|---|---|
| `rimraf` | native `fs.rm` |
| `isomorphic-ws` | native `WebSocket` (Node 22+ global) |
| `npm-watch` (unmaintained) | native `tsc --watch` / `vite build --watch` |
| `ts-node` | `tsx` |
| `classnames` | `clsx` |
| `npm-run-all` (abandoned 2017) | `npm-run-all2` |
| `jsdom` | `happy-dom` (workspace-wide) |
| `c8`, `@types/uuid` | dropped (redundant) |
| **`cabbages`** | dead — a leftover from the original author's config template in solid-primitives, never imported |
| **`tailwindcss`** (react-hooks external) | dead — a never-imported bundler external |

### Package manager — pnpm 11
- `catalog:` for shared dev **and** published runtime deps; the release flow resolves `catalog:`/`workspace:` to real versions at publish time.
- Supply-chain guard: `minimumReleaseAge` (3-day quarantine on fresh versions), an `allowBuilds` allow-list, pinned `overrides` for known-vulnerable transitive deps (`semver`, `tmp`).
- pnpm is pinned via the standard `packageManager` field (see the reviewer note below on why not `devEngines`).

### Language & build
- **TypeScript 6** workspace-wide (ready for 7), with a shared **base/lib/app tsconfig** hierarchy replacing 17 duplicated configs.
- **Vite 8** + vitest 4.1, unified via the catalog.
- **React 19** for dev/examples; `@automerge/react` now declares React as a **peer** dependency.

### Lint & format — Rust-based, and stricter
- **oxfmt** replaces prettier 2, **oxlint** replaces eslint 9 + typescript-eslint.
- `lint` now also **type-checks** (`lint:check-types`), so type regressions fail lint.
- `isolatedModules` enabled everywhere — catches the runtime-breaking `export { Type }` re-export bug class at compile time.
- *Note:* oxfmt required a few minor formatting changes — these would have been required by prettier 3 too, so it's not oxfmt-specific churn.

### Security
- **`pnpm audit` is clean** — no known vulnerabilities. Every dependency is current.

## Continuous integration

The workflows were on long-stale action majors, several running on Node runtimes GitHub is retiring: **as of June 2, 2026 GitHub forces JavaScript actions onto the Node 24 runtime**, with Node 20 removed from runners on Sept 16, 2026. The old pins (`checkout` v2/v3, `setup-node` v3, `gh-release` v1, `upload-artifact` v4) ran on the older Node 16/20 runtimes; the current majors all target Node 24 — clearing the deprecation ahead of the deadline, and unlocking modern CI features along the way:

| Action | main | now |
|---|---|---|
| `actions/checkout` | v2 / v3 | **v6** |
| `actions/setup-node` | v3 | **v6** |
| `pnpm/action-setup` | v4 | **v6** |
| `actions/upload-artifact` | v4 | **v7** |
| `softprops/action-gh-release` | v1 | **v3** |

Newly enabled — none of these existed on `main`:

- **Concurrency cancellation** (`cancel-in-progress`) on lint / test / Playwright — a new push supersedes in-flight runs instead of stacking them.
- **Least-privilege `permissions`** — `contents: read` on the PR workflows, scoped writes on release; the broad default `GITHUB_TOKEN` is no longer handed to every job.
- **`persist-credentials: false`** on PR checkouts — the token isn't left on the runner for jobs that don't push.
- **Native pnpm caching** via `setup-node@v6`'s `cache: pnpm`.
- **Node test matrix `[22, 24, 25]`** (was a single `20.x`).
- **Build before lint**, so the type-aware oxlint can resolve workspace `.d.ts`.

## Examples & docs
- **create-vite-app** / **create-repo-node-app** modernized to TypeScript templates with current patterns.
- Examples now **build in CI** — this caught a real worker-resolution break under Vite 8's stricter resolver.
- Release modernized: `lerna version` + `pnpm -r publish` (so `catalog:`/`workspace:` resolve), validated end-to-end via dry-run.
- typedoc warnings **68 → 40**, and made runnable locally via `build:docs`.

## Subtle bugs uncovered

Moving to `tsx`/esbuild (file-by-file transpile, no type-aware elision) surfaced a latent **`export { Type }` re-export** that `tsc` had been silently erasing — it would throw at runtime under any non-type-aware transpiler/bundler. Fixed, and `isolatedModules` now prevents recurrence. A flaky/incorrect fuzz test was also fixed (deterministic sync wait + a sharePolicy peer-id mismatch).

## Verification

- `pnpm build` · `pnpm lint` (incl. type-check) · `CI=true pnpm test` (**817 passing**, 2 skipped — includes the subdoc-handles tests from the merged-in main) · `pnpm repocheck` · `pnpm audit` — all green.
- Full release flow dry-run (`lerna version` + `pnpm -r publish --dry-run`) — green.

## Notes for reviewers

- **pnpm pinning & Dependabot**: pnpm is pinned via the standard `packageManager` field (`pnpm@11.x`). An earlier revision used pnpm 11's `devEngines.packageManager` to accept any 11.x, but that makes `pnpm-lock.yaml` multi-document, which Dependabot's pnpm-v11 support can't yet parse (dependabot/dependabot-core#14794) — and this repo relies on Dependabot security updates. The `packageManager` pin keeps the lockfile single-document and Dependabot-parseable. (Thanks @msakrejda for catching that.)

---

*Detailed rationale, measurements, and edge cases are in the individual commit messages.*


msakrejda:
This is great! It will take a while to review, but some preliminary feedback:

> This repo doesn't use Dependabot, so no impact today; Renovate handles pnpm catalogs fine.

It does use Dependabot: https://github.com/automerge/automerge-repo/pull/639 -- or were you talking about something else?

> catalog: for shared dev and published runtime deps; the release flow resolves catalog:/workspace: to real versions at publish time.

You're referencing [this behavior](https://pnpm.io/catalogs#publishing), right? I think that makes sense.

It might be good to split this into several PRs if that's not too much work. E.g., the different phases described in "Why, and in what order" could make sense. Right now it's just a lot.

darcyparker:
> It does use Dependabot: #639 -- or were you talking about something else?

Sorry, There is no committed `.github/dependeabot.yml`; which led me to think you weren't using it. But clearly you have Dependabot security updates / alerts enabled in repo settings.  Sorry about that. 

`devEngines.packageManager` is not going to work (until issues are resolved by dependabot team), so I reverted back to `packageManager` and tested.  Lockfile is now canonical single-doc. 

I also just saw a branch conflict and merged current `main` in and fixed conflict. (they were related to the typedoc issues that I fixed and the `subdoc-handles` refactor.)

> You're referencing [this behavior](https://pnpm.io/catalogs#publishing), right? I think that makes sense.

Yes — the publish-time substitution from https://pnpm.io/catalogs#publishing (`pnpm publish/pnpm pack` rewrites `catalog:/workspace` into to concrete versions in the published manifest). I confirmed it with `pnpm pack` on each package.

> It might be good to split this into several PRs if that's not too much work. E.g., the different phases described in "Why, and in what order" could make sense. Right now it's just a lot.

On splitting — that's fair, and I think it's the right instinct; the work does fall into the phases I listed and could be partitioned into groups. Two things make a clean split take some care, though:

1. The commits are only approximately linear. The `minimumReleaseAge` supply-chain gate (a 3-day quarantine on freshly published versions) meant some dependency updates couldn't be installed until they'd aged out — so a few follow-up commits landed days after where they should be from a 'topic' pov.  And then after I waited 3 days, the new toolchain surfaced issues (a type-only re-export that only broke under esbuild/tsx, a couple of flaky/buggy tests, and the devEngines/Dependabot interaction you just caught, and the merge conflicts I just resolved from the `subdoc-handles` refactor that merged to `main` a moment ago.) that were fixed in later commits rather  than folded back into the originals. So partitioning each sub-PR to be green on its own is doable, but it's **not** a mechanical `rebase --split`.

2. The CI piece has a real deadline. As of June 2, 2026 GitHub forces JavaScript actions onto the Node 24 runtime, which breaks the older action majors this repo currently pins (`checkout` v2/v3, `setup-node` v3, …). So that part is time-sensitive in a way the rest isn't — though you may weigh that urgency differently, and I'd defer to you.

Candidly, my near-term availability is limited — I won't be able to do a careful split over the weekend or next week. So, options depending on your preference:

  - Carve out just the CI/Actions upgrade as a small standalone PR now — it's self-contained, quick to review, and it's the part with the June 2 deadline. The rest can then be split at a more deliberate pace.
  - Keep this as one PR for now, and I'll break it into the phase-based PRs when I have a proper block of time.

It's a bit like a feature branch that accumulated a chunk of related work — I'm happy to go whichever way is least friction for you as the reviewer. What's your preference?

msakrejda:
> Carve out just the CI/Actions upgrade as a small standalone PR now — it's self-contained, quick to review, and it's the part with the June 2 deadline. The rest can then be split at a more deliberate pace.

If you can do that, it would be fantastic. Then we could keep CI going and I could help reorganize the other pieces with more time.

packages/automerge-repo-svelte-store/tsconfig.json
```diff
@@ -1,14 +1,13 @@
 {
+  "extends": "../../tsconfig.base.json",
   "compilerOptions": {
-    "target": "ES2020",
     "module": "ES2020",
-    "moduleResolution": "bundler",
+    "moduleResolution": "Bundler",
```
neftaly: i think this is meant to be lowercase? 
https://www.typescriptlang.org/tsconfig/#moduleResolution

packages/create-vite-app/template/package.json
```diff
@@ -1,30 +1,29 @@
 {
   "name": "projectName",
   "private": true,
-  "version": "1.2.1",
+  "version": "0.1.0",
```
neftaly: i think this should be 0.0.0
https://github.com/vitejs/vite/blob/main/packages/create-vite/template-react/package.json

packages/create-vite-app/template/package.json
```diff
@@ -1,30 +1,29 @@
...
   "type": "module",
+  "engines": {
+    "node": ">=20.19"
```
neftaly: these should probably be all the same (20 or 22)

packages/create-vite-app/template/tsconfig.json
```diff
@@ -1,13 +1,13 @@
...
+    "target": "ESNext",
     "useDefineForClassFields": true,
-    "lib": ["ES2020", "DOM", "DOM.Iterable"],
+    "lib": ["ESNext", "DOM", "DOM.Iterable"],
     "module": "ESNext",
     "skipLibCheck": true,
 
     /* Bundler mode */
-    "moduleResolution": "bundler",
+    "moduleResolution": "Bundler",
```
neftaly: `bundler`

.editorconfig
```diff
@@ -1,3 +1,12 @@
-[*.{ts,js}]
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
```
neftaly, line 7: should md and js config for trailing whitespace be different?

tsconfig.app.json
```diff
@@ -0,0 +1,13 @@
+{
+  "extends": "./tsconfig.base.json",
+  "compilerOptions": {
+    "module": "ESNext",
+    "moduleResolution": "Bundler",
```
neftaly: `bundler`

packages/automerge-repo-react-hooks/src/useStateRef.ts
neftaly, line 1: i think this could be a separate pr

packages/automerge-repo-svelte-store/src/lib/index.ts
```diff
@@ -121,8 +121,10 @@ export function createAutomergeStore(repo: Repo) {
       subscribe,
       set,
       update: updater => {
+        // doc() returns `Doc<T> | undefined`; the store value is
+        // `Doc<T> | null`, so bridge undefined -> null (matching onChange above).
         const currentDoc = handle.doc()
-        const newValue = updater(currentDoc)
+        const newValue = updater(currentDoc ?? null)
```
neftaly, line 127: is there a reason to match the onchange?

neftaly:
I think the tests can also be broken out. But 100% i get why you would want to land all this at roughly the same time, otherwise it will get stale and never land

packages/automerge-repo-svelte-store/src/lib/index.ts
darcyparker: Because of typescript error otherwise. Typescript 6 surfaces this error, typescript 5 did not.

packages/automerge-repo-react-hooks/src/useStateRef.ts
darcyparker: The vite 8 upgrade requires it to happen in same PR.

From the commit message:

> - Vite 8 is Rolldown-based: it still honors build.rollupOptions (react-hooks/solid-primitives   externals + .d.ts verified). Unlike Rollup, Rolldown preserves a bundled CJS dep's require() of an external, so the CommonJS react-usestateref left a bare require-of-react in react-hooks'  bundle (broke ESM consumers + the example e2e). Replaced it with a small inlined ESM  useStateRef hook and dropped the dep (also from the react-use-presence example)

packages/automerge-repo-svelte-store/tsconfig.json
darcyparker: Thanks - I pushed commit to lower case.

packages/create-vite-app/template/package.json
darcyparker: Thanks - Done. I set both `create-*` templates to 0.0.0, matching the create-vite convention.

packages/create-vite-app/template/package.json
darcyparker: The split was intentional: `create-repo-node-app` runs node `index.ts` and needs Node's native TS stripping (default-on at 22.18), while create-vite-app only needed Vite 8's 20.19 floor — so I'd resisted bumping it. 

But you're right that one floor is cleaner; I've aligned create-vite-app up to >=22.18 so both templates match.

packages/create-vite-app/template/tsconfig.json
darcyparker: Done

tsconfig.app.json
darcyparker: Done

.editorconfig
darcyparker: Intentional — Markdown uses two trailing spaces as a hard line break, so the `[*.md]` block sets `trim_trailing_whitespace = false` while everything else trims. It's the standard editorconfig pattern for
markdown.

packages/automerge-repo-svelte-store/src/lib/index.ts
darcyparker: Note, I thought about whether the type should allow `undefined`. But automerge changes work better with explicit `null`.  So I think it was an oversight to not have `?? null` before.

darcyparker:
> I think the tests can also be broken out. But 100% i get why you would want to land all this at roughly the same time, otherwise it will get stale and never land

Thanks for your review @neftaly.  I added some commits to address some of your feedback and responded to others.

WRT to tests:   Happy to split where it helps — but looking at the actual test footprint here, there isn't really a standalone "tests" change to lift out. Only two `*.test.ts` files changed, and each rides with a specific change
  rather than being test work of its own:

  - `Websocket.test.ts` — required by dropping `isomorphic-ws`: `new WebSocket.Server()` → `new WebSocketServer()` (the named `ws` import), plus the keep-alive test rewritten to drive the server with a raw ws client, since the now-native client auto-pongs and can't be silenced from script. Belongs with the native-WebSocket change.
  - `DocHandle.test.ts` — pure `oxfmt` reflow (redundant parens dropped, one call re-wrapped); no logic change. The kind of churn any formatter migration produces — this was Prettier 2 → oxfmt, and a Prettier 3/4 bump would reflow similar lines. Belongs with the formatting change.

The only other test-adjacent edits are a one-line `vitest.config.ts` env swap (`jsdom` → `happy-dom`) and the `fuzz`/`fuzz.ts` fix (a deterministic sync wait + a sharePolicy peer-id bug) — again each tied to its own  change.

So in a split, these naturally fold into their feature PRs rather than a separate "tests" PR. 

Was there a particular test change you had in mind, though — the fuzz fix, the happy-dom swap, or something else?
Happy to consider carving out whatever you were picturing.

darcyparker:
FYI: I am breaking this up into separate PRs (with some help from AI)... The first is CI + node baseline.  The others I have a plan for, but may or may not get to this week @msakrejda.

The PRs will need to be merged sequentially...

darcyparker:
Update: the split is complete. #653 is now broken into **8 stacked PRs**, each branched off `main` and intended to merge in order:

1. #654 — CI: modernize GitHub Actions + Node 22 baseline
2. #655 — adopt pnpm 11 + supply-chain controls
3. #656 — migrate to oxfmt + oxlint
4. #657 — TypeScript 6 + type-aware oxlint
5. #658 — upgrade to Vite 8 + vitest 4.1
6. #659 — dependency currency + native-API swaps
7. #660 — pnpm catalog + publishing hygiene + `pnpm -r publish` release
8. #661 — examples, scaffolders, docs & dev-tooling cleanup

#654 is ready for review; #655–#661 are drafts that will be flipped to ready and rebased onto `main` as each predecessor merges (until then each draft's diff still includes its ancestors' commits — review the topmost commit).

**Feedback from this PR has all been either addressed in the split PRs or answered inline above**, e.g.:
- lowercase `moduleResolution` + `bundler` in the tsconfigs → #657
- svelte-store `undefined` → `null` (surfaced by TS 6) → answered inline; lands in #657
- inlined `useStateRef` (coupled to the Vite 8 / Rolldown bundle fix, so it can't be a separate PR) → answered inline; lands in #658
- `create-*` template fixes (`version: 0.0.0`, engines, `bundler`) → #661
- `.editorconfig` Markdown trailing-whitespace kept distinct (two trailing spaces = a hard line break) → answered inline; config lands in #661
- carving CI out as a standalone PR first → #654

**Please direct any further feedback to the relevant PR in the stack** rather than here, so it lands next to the change it concerns.

Closing this in favor of the stack above. Thanks @msakrejda and @neftaly — the review directly shaped how it was split. 🙏

neftaly:
For future reference i think a pr of this size is best landed in like 3-4 stages?

darcyparker:
> For future reference i think a pr of this size is best landed in like 3-4 stages?

I wanted to club these into one PR because they were all related (like multiple PRs merging into a feature branch).  I could see splitting it up into 8 separate modular pieces.  A couple of these could be clubbed together to hit 3-4 stages... but its a lot of ceremony when they should merge together relatively quickly.

PR #654, opened 2026-06-01
ci: modernize GitHub Actions + Node 22 baseline (part 1 of #653)

darcyparker:
Part 1 of breaking up #653 into smaller, stacked PRs (per the review discussion there). This is the time-sensitive piece, and it's self-contained — it doesn't depend on any of the pnpm 11 / TypeScript 6 / Vite 8 work that follows.

## Why this one first

As of **June 2, 2026**, GitHub forces JavaScript actions onto the Node 24 runtime (Node 20 is removed from runners on Sept 16, 2026). The workflows currently pin action majors that run on the retiring Node 16/20 runtimes (`checkout` v2/v3, `setup-node` v3, `gh-release` v1, …), so this is the part of #653 with a real deadline.

## GitHub Actions → current majors (Node 24 runtime)

| Action | before | after |
|---|---|---|
| `actions/checkout` | v2 / v3 | **v6** |
| `actions/setup-node` | v3 | **v6** |
| `pnpm/action-setup` | v4 | **v6** |
| `actions/upload-artifact` | v4 | **v7** |
| `softprops/action-gh-release` | v1 | **v3** |

## Workflow hygiene

- **Concurrency cancellation** (`cancel-in-progress`) on lint / tests / Playwright — a new push supersedes in-flight runs.
- **Least-privilege `permissions: contents: read`** on the PR workflows (release keeps `contents: write`).
- **`persist-credentials: false`** on PR-only checkouts.
- Reorder to **checkout → pnpm → setup-node** with **`cache: pnpm`**.

## Node baseline

- `engines.node` `>=20.0 <22.0.0` → `>=22.13`.
- `.nvmrc` `lts/iron` → `lts/jod` (current Node 22 LTS).
- Drop the stale `.node-version` (was `v20.19.5`) so nvm/fnm/asdf all read `.nvmrc`.
- Test matrix `20.x` → **`[22, 24, 25]`**.

Stays on pnpm 10 and the existing lerna-publish release flow — pnpm 11 / catalog / the publish-flow change come in later PRs in the stack.

## A pre-existing lint fix rides along

`main`'s `pnpm lint` is currently **red** independently of this work: four files from the subdoc-handles work were committed with a newer prettier's formatting (multi-line imports), but the repo pins **prettier 2.8.8**, whose `prettier --check` rejects them. The last commit re-runs the pinned prettier over those files (`AutomergeUrl.ts`, `subdoc-handles/parser.ts`, `subdoc-handles/types.ts`, `DocHandle.test.ts`) to get a green lint here.

Flagging it — you may prefer to bump prettier instead; the prettier→oxfmt migration later in the stack also resolves it.

## Playwright: bump `@playwright/test` to 1.60

Moving the runner to Node 24 surfaced that `@playwright/test` 1.57's `playwright install --with-deps` **hangs on the current GitHub-hosted Ubuntu runner** — Chromium downloads to 100%, then the step stalls until the 60-minute job timeout. 1.60 installs the browsers and runs the react-use-presence e2e in under two minutes on the same runner, so this bumps `@playwright/test` to `^1.60.0`.

## Verification

`pnpm build`, `pnpm lint` (prettier + eslint), and `pnpm test` (**817 passing**) all green on Node 25 / pnpm 10.

---

Part of #653. Subsequent PRs in the stack: pnpm 11 foundation → format/lint (oxfmt + oxlint) → TypeScript 6 → Vite 8 → dependency currency → published-package hygiene → release flow → scaffolders & examples.


darcyparker:
@msakrejda - This is part 1... the time sensitive one for June 2nd.

.github/workflows/lint.yaml
```diff
@@ -15,23 +15,31 @@ on:
...
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/setup-node@v3
+      - uses: actions/checkout@v6
         with:
-          node-version: "20.x"
-          registry-url: "https://registry.npmjs.org"
-      - uses: actions/checkout@v3
-        with:
-          fetch-depth: 0
```
msakrejda, line 30: Why drop `fetch-depth`?

msakrejda (approved):
Nice, thanks for breaking this out. One minor question, but otherwise this looks great. I'll merge soon.

.github/workflows/lint.yaml
darcyparker: Sorry. Reason was dropped in commit message.

From #653 - 
```
- **Drop redundant `ref: ${{ github.ref }}`** from every checkout (5
  occurrences across the 4 files): it's exactly what
  `actions/checkout` already defaults to. Removing keeps the
  pull_request merge-commit simulation behavior intact (default ref
  on PR events is `refs/pull/<num>/merge`). Also dropped
  `fetch-depth: 0` from lint/tests/playwright - they don't need
  history, only the working tree. Release keeps `fetch-depth: 2`
  (salsify, typedoc) and `fetch-depth: 0` (lerna version) as before.
```

PR #656, opened 2026-06-01
chore: migrate to oxfmt + oxlint (part 3 of #653)

darcyparker:
> **Draft — stacked on #654 and #655, waiting for them to merge.**
> Until the predecessors land, the commit list / diff here also includes their commits — review just the `chore: migrate formatting and linting to oxfmt + oxlint` commit, or wait.

Part 3 of breaking up #653. Migrates the formatter and linter from prettier + eslint to the oxc toolchain.

## What this does

- **prettier 2 → oxfmt** — `.oxfmtrc.json` carries over the existing prettier settings (semi:false, singleQuote:false, trailingComma:es5, arrowParens:avoid, printWidth:80); `format` → `oxfmt`. Most of the diff is the one-time reformat where oxfmt and prettier 2 differ (e.g. lowercase `<!doctype>`, `await import()` wrap style).
- **eslint 9 (+ airbnb, @typescript-eslint) → oxlint** — `.oxlintrc.json`. The custom `automerge-slimport` eslint rule that enforced `/slim` imports is replaced by oxlint's built-in `no-restricted-imports` (same intent; `automerge-react` and `automerge-vanillajs` are exempted via overrides). Removed `eslint.config.mjs`, `.prettierignore`, and the 12 eslint/prettier devDependencies. `lint` → `oxfmt --check . && oxlint`.
- Migrated two inline disable directives to oxlint's rule names (`…/enforce-automerge-slim-import` → `no-restricted-imports`, `@typescript-eslint/ban-types` → `no-wrapper-object-types`).

## Type-aware linting comes with TypeScript 6

oxlint runs **non-type-aware** here. The type-aware engine (`oxlint-tsgolint`) requires each tsconfig to set `rootDir` explicitly, which the base/lib/app tsconfig refactor adds in the TypeScript 6 PR — so type-aware linting (`typeAware`, `no-floating-promises`, build-before-lint, `lint:check-types`) is enabled there.

## Verification

`pnpm build`, `pnpm lint`, and `pnpm test` (**817 passing**) all green on Node 25 / pnpm 11.5.

---

Part of #653 · stacked on #654, #655.


msakrejda:
@darcyparker #654 and #655 are merged now, so this is ready, right?

.oxlintrc.json
```diff
@@ -0,0 +1,75 @@
...
+    "**/.svelte-kit/**",
+    "examples/**",
+    "**/test/**",
+    "**/*.d.ts",
+    "**/*.config.ts",
+    "**/fuzz/**",
+    "**/testSetup.ts",
+    "playwright.base.ts",
+    "packages/create-vite-app/**",
+    "packages/create-repo-node-app/**"
```
psmolaga: @darcyparker what is the reason we need to ignore these two packages?

psmolaga (approved):

.oxlintrc.json
darcyparker: Good catch — you're right that the whole `create-repo-node-app` package shouldn't be ignored here.

Both scaffolders ship code the monorepo's oxlint can't meaningfully lint, but in different amounts:

- `create-vite-app/**` has a `template/` directory — a standalone project scaffold with its own lint/tsconfig, whose imports resolve against the *generated* app's dependencies rather than the monorepo's (and it's excluded from the pnpm workspace), so the root oxlint can't lint it.
- `create-repo-node-app`, though, is still just a plain CLI in this PR, so ignoring the whole package was too broad. I've narrowed it to `packages/create-repo-node-app/postbuild.js` (the only generated file), so `src/index.ts` is now linted. It gains a `template/` directory in #661 (part 8 of #653) — that's the PR that re-broadens this ignore to the whole package, since the template is out-of-tree scaffold code like create-vite-app's.

Pushed (rebased onto current `main`).

darcyparker:
> @darcyparker #654 and #655 are merged now, so this is ready, right?

Yes - ready. I just did a small update based on @psmolaga review above (https://github.com/automerge/automerge-repo/pull/656#discussion_r3341111463). The other drafts are also ready to go...

I'll watch as you merge and take them out of draft.  But if you want to continue reviewing/merging at a faster cadence than I can keep up, please feel free to take them out of draft.

.oxlintrc.json
```diff
@@ -0,0 +1,75 @@
...
+    "no-empty-function": "off",
+    "no-param-reassign": "off",
+    "typescript/no-non-null-assertion": "off",
+    "typescript/no-explicit-any": "off",
+    "react-hooks/exhaustive-deps": "off"
+  },
+  "settings": {
+    "react": { "version": "18" }
+  },
+  "overrides": [
```
neftaly, line 64: Just checking these overrides will get fixed later?

darcyparker: These overrides are intentional and stay - not a temporary workaround. `@automerge/react` and `@automerge/vanillajs` are the "quick-import" / batteries-included packages that deliberately re-export the *full* `@automerge/automerge` build (not `/slim`), so the slim-import rule (`no-restricted-imports`) shouldn't apply to them. The old eslint setup exempted them the same way (via the custom `automerge-slimport` allowlist); this just carries that intent over to oxlint.

(For the rest of the config: this PR runs oxlint non-type-aware; type-aware linting is enabled in the TS 6 PR, #657 - but it leaves this exemption unchanged.)

neftaly: thanks!

neftaly (approved):

msakrejda (approved):

PR #657, opened 2026-06-01
chore: TypeScript 6 + type-aware oxlint (part 4 of #653)

darcyparker:
> Part 4 of #653. 

Upgrades TypeScript 5 → 6, consolidates the tsconfigs, and enables the type-aware oxlint layer the upgrade unlocks.

## TypeScript 6
- `typescript` ^5.x → **^6.0.3** in the packages that declare it.
- New root `tsconfig.base.json` / `tsconfig.lib.json` / `tsconfig.app.json`; every package tsconfig now extends one instead of duplicating settings (`isolatedModules` on base; `declaration`+`declarationMap`+`sourceMap` on lib; bundler/noEmit on app). `moduleResolution` uses the lowercase forms from the TS docs (`nodenext`, `bundler`).
- `isolatedModules` surfaced a type-only re-export → `export type` in `Repo.ts`.
- A few latent issues TS 6 / type-aware lint surfaced, fixed in place: `T | unknown` collapses to `unknown` (`path-ops.ts`, `DocHandle.ts`); svelte-store's `update()` bridges `doc()`'s `Doc<T> | undefined` to the store's `Doc<T> | null`.

### Looking ahead: TypeScript 7

As #653 noted, moving to TypeScript 6 positions us for a **trivial bump to TypeScript 7 (~June 2026)** — the Go-based native compiler. 7.0 targets the same language and config as 6.x (6.x mainly clears the deprecations 7 drops), so it should be a fast, low-risk follow-up, after which the whole workspace inherits the native compiler's ~10× type-checking/build performance win.

## Type-aware oxlint
- Added `oxlint-tsgolint`; `.oxlintrc.json` sets `typeAware: true` + `no-floating-promises`. (tsgolint needs the explicit tsconfig `rootDir` the base/lib/app refactor provides — which is why this rides with TS 6.)
- `lint.yaml` builds before linting; added `lint:check-types` (`tsc --noEmit` across packages) chained into `lint`. Example type-checking arrives with the examples-cleanup PR (which gives them tsconfigs and declares their worker deps).

## Verification

`pnpm build`, `pnpm lint`, and `pnpm test` (**817 passing**) all green on Node 25 / pnpm 11.5.

---

Part of #653 · stacked on #654, #655, #656.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -40,8 +40,9 @@ jobs:
           node-version: 24
           registry-url: "https://registry.npmjs.org"
           cache: pnpm
-      - name: Install
+      - name: Install and Build
         run: |
           pnpm install
+          pnpm build
       - name: Lint
         run: pnpm lint
--- a/examples/react-counter/tsconfig.json
+++ b/examples/react-counter/tsconfig.json
@@ -1,20 +1,4 @@
 {
-  "compilerOptions": {
-    "allowJs": false,
-    "allowSyntheticDefaultImports": true,
-    "esModuleInterop": false,
-    "forceConsistentCasingInFileNames": true,
-    "isolatedModules": true,
-    "jsx": "react-jsx",
-    "lib": ["DOM", "DOM.Iterable", "ESNext"],
-    "module": "ESNext",
-    "moduleResolution": "Bundler",
-    "noEmit": true,
-    "resolveJsonModule": true,
-    "skipLibCheck": true,
-    "strict": true,
-    "target": "ESNext",
-    "useDefineForClassFields": true
-  },
+  "extends": "../../tsconfig.app.json",
   "include": ["src"]
 }
--- a/package.json
+++ b/package.json
@@ -18,12 +18,14 @@
     "preinstall": "npx only-allow pnpm",
     "repocheck": "manypkg check",
     "format": "oxfmt",
-    "lint": "oxfmt --check . && oxlint",
+    "lint": "oxfmt --check . && oxlint && pnpm run lint:check-types",
     "test": "vitest",
     "test:log": "cross-env DEBUG='automerge-repo:*' vitest",
     "test:coverage": "vitest --coverage",
     "test:ui": "vitest --coverage --ui",
-    "watch": "pnpm run --parallel --stream -r watch "
+    "watch": "pnpm run --parallel --stream -r watch ",
+    "lint:check-types": "pnpm run lint:check-types:packages",
+    "lint:check-types:packages": "pnpm --filter './packages/*' exec tsc --noEmit"
   },
   "engines": {
     "node": ">=22.13",
@@ -58,10 +60,11 @@
     "npm-watch": "^0.11.0",
     "oxfmt": "^0.52.0",
     "oxlint": "^1.67.0",
+    "oxlint-tsgolint": "^0.23.0",
     "portfinder": "^1.0.32",
     "ts-node": "^10.9.2",
     "typedoc": "^0.28.7",
-    "typescript": "^5.4.2",
+    "typescript": "^6.0.3",
     "vite": "^7.3.0",
     "vite-plugin-wasm": "^3.5.0",
     "vitest": "^4.0.15"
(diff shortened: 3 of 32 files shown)


.github/workflows/lint.yaml
```diff
@@ -40,8 +40,9 @@ jobs:
           node-version: 24
           registry-url: "https://registry.npmjs.org"
           cache: pnpm
-      - name: Install
+      - name: Install and Build
         run: |
           pnpm install
+          pnpm build
```
psmolaga, line 46: IMHO, I like to keep these two as separate steps in CI because you can clearly see how much time the install takes and how much the build takes.

darcyparker: Good idea. To keep this PR focused on the TS 6 change, I've pulled it into a follow-up — #663 — which also applies the same split to `tests.yaml`, `playwright.yaml`, and `release.yaml`, since they already carry the combined `Install and Build` step (from #654).

The `lint.yaml` step here (the build-before-lint this PR adds) will fold into #663 once this one lands.

examples/react-counter/tsconfig.json
```diff
@@ -1,20 +1,4 @@
...
-    "module": "ESNext",
-    "moduleResolution": "Bundler",
-    "noEmit": true,
-    "resolveJsonModule": true,
-    "skipLibCheck": true,
-    "strict": true,
-    "target": "ESNext",
-    "useDefineForClassFields": true
-  },
+  "extends": "../../tsconfig.app.json",
```
msakrejda, line 2: This means that the examples are not self-contained (though I'm not sure if they were before), right? I think that's okay: using these standalone as-is seems unlikely anyway.

package.json
```diff
@@ -17,13 +17,15 @@
...
+    "format": "oxfmt",
+    "lint": "oxfmt --check . && oxlint && pnpm run lint:check-types",
     "test": "vitest",
     "test:log": "cross-env DEBUG='automerge-repo:*' vitest",
     "test:coverage": "vitest --coverage",
     "test:ui": "vitest --coverage --ui",
-    "watch": "pnpm run --parallel --stream -r watch "
+    "watch": "pnpm run --parallel --stream -r watch ",
+    "lint:check-types": "pnpm run lint:check-types:packages",
+    "lint:check-types:packages": "pnpm --filter './packages/*' exec tsc --noEmit"
```
msakrejda, line 28: Why both of these?

msakrejda (approved):
It looks like there are some conflicts since #656 was merged, and I had a question on the type linting scripts added, but otherwise this looks great.

msakrejda:
Ah, you're on it! Looking again.

examples/react-counter/tsconfig.json
darcyparker: Right: before this they each carried a full standalone `compilerOptions` block; now they `extend` the repo's shared `tsconfig.app.json`, which makes setting these options more DRY.  That's the point of the `base/lib/app` `tsconfig` split in this PR: packages and examples share one set of compiler settings instead of duplicating them. Agreed that standalone use is unlikely, so leaning on the shared base is fine here. (No packaging angle either - the examples aren't published, and `tsc` just resolves the extends chain at build time.)

package.json
darcyparker: Both of these are the TypeScript type-check (`tsc --noEmit`) - separate from oxlint, which doesn't fully type-check, so `lint` runs both. I broke `lint:check-types:packages` out as a named step rather than inlining the longer `pnpm --filter './packages/*' exec tsc --noEmit` into the wrapper, mainly for readability - and it lets the examples check drop in cleanly when #661 adds `lint:check-types:examples`. Happy to collapse them into a single `lint:check-types` if you'd rather have it as one.

darcyparker:
> Ah, you're on it! Looking again.

Yep.  Done with rebase and replying to your recent questions. Let me know if you have more.  If you merge, I'll update the next one and take it out of draft.

package.json
msakrejda: Okay, that's fine. It just looked a little odd, but we can revisit once I merge the full series.

msakrejda (approved):

PR #660, opened 2026-06-01
chore(deps): pnpm catalog + publishing hygiene + pnpm-publish release (part 7 of #653)

darcyparker:
Adopts the pnpm `catalog:` protocol, finishes publishing hygiene on the released packages, and switches the release to `pnpm -r publish` so `catalog:` resolves at publish time.

## Catalog adoption
- A `catalog:` block in `pnpm-workspace.yaml` holds the shared dependency versions; every workspace package (including the root) points at it, ending the per-package version drift.
- **Single `@automerge/automerge` instance.** Cataloging `@automerge/automerge` at one version (`^3.2.6`) collapses it to a single copy. This matters for wasm: the full entry (`@automerge/automerge`, used by the test setup) and the `/slim` entry (used by `automerge-repo`) must share one wasm backend. Two coexisting copies leave the `/slim` copy uninitialized → `RangeError: Automerge.use() not called`. (Same de-duplication as #653.)
- Published runtime deps are cataloged too: `pnpm -r publish` resolves `catalog:` to a concrete range in the published `package.json`, so consumers never see `catalog:`.

## Publishing hygiene
- **Tarball footprint: stop shipping source noise, start shipping debuggable maps.** `files: ["dist", "src"]` scopes each published package to its build output + sources, so npm stops packing the test/fuzz/config noise it shipped before. For `automerge-repo` the tarball drops **304 kB → 220 kB (−28%)**, 355 → 313 files, **38 → 0** test/fuzz entries, while keeping all 124 per-file `.js.map` source maps (emitted since #657) and shipping `src/` so they resolve. Net *smaller even with* the source maps added; the extra files are just the per-file `.js.map`s. This too feeds #653's "smaller published packages" headline.
- `exports` maps + per-package `engines.node` on the published packages.
- Declare every previously-phantom dependency (deps that resolved only via hoisting).
- Externalize `@automerge/*` in the `react-hooks` and `solid-primitives` bundles. Consumers no longer get a second bundled copy of automerge, and the published bundles shrink sharply: `react-hooks` ~173 kB → ~49 kB (−72%), `solid-primitives` ~38 kB → ~6 kB (−83%). This is part of what drives #653's "smaller published packages" headline.
- Remove two dead bundler externals from those same `external` arrays: `tailwindcss` (in `react-hooks`) and `cabbages` (in `solid-primitives`). Neither is ever imported; both are stale leftovers from the original Vite config templates.
- Drop the stale `solid-primitives/.npmignore` (`files:` supersedes it).

## Release flow
- `release.yaml` publishes via `lerna version` + `pnpm -r publish` (was `lerna publish`), so `catalog:` is resolved at publish time; `lerna` only versions and tags.

## Verification
`pnpm build`, `pnpm lint`, `pnpm test` (**819 passing**), and `pnpm repocheck` all green; `pnpm pack` confirms `catalog:` resolves to real ranges (`@automerge/automerge: ^3.2.6`, `cbor-x: ^1.6.4`, `uuid: ^14.0.0`) in the tarball.

---

Part of #653.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -46,7 +46,7 @@ jobs:
           pnpm install
           pnpm build
       - name: Build docs
-        run: npx typedoc
+        run: pnpm exec typedoc
       - name: Deploy 🚀
         uses: JamesIves/github-pages-deploy-action@v4
         with:
@@ -93,12 +93,16 @@ jobs:
           fi
           git config user.name github-actions
           git config user.email github-actions@github.com
-          EXTRA_ARGS=""
+          # lerna handles versioning, inter-package ranges, git tag and push;
+          # pnpm does the publish so that workspace: and catalog: specifiers
+          # resolve to real versions (lerna publish leaves catalog: unresolved).
+          pnpm exec lerna version ${VERSION} --yes --force-publish
+          PUBLISH_ARGS=""
           if [[ $VERSION == *"alpha."* ]] || [[ $VERSION == *"beta."* ]] || [[ $VERSION == *"rc."* ]]; then
             echo "Is pre-release version"
-            EXTRA_ARGS="$EXTRA_ARGS --dist-tag next"
+            PUBLISH_ARGS="--tag next"
           fi
-          npx lerna publish ${VERSION} --yes --force-publish $EXTRA_ARGS
+          pnpm -r publish --no-git-checks $PUBLISH_ARGS
         env:
           VERSION: ${{ needs.check_if_version_upgraded.outputs.version }}
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
--- a/examples/react-counter/package.json
+++ b/examples/react-counter/package.json
@@ -10,7 +10,7 @@
   },
   "dependencies": {
     "@automerge/react": "workspace:*",
-    "react": "^19.2.6",
-    "react-dom": "^19.2.6"
+    "react": "catalog:",
+    "react-dom": "catalog:"
   }
 }
--- a/examples/react-todo/package.json
+++ b/examples/react-todo/package.json
@@ -13,9 +13,9 @@
     "autoprefixer": "^10.4.13",
     "clsx": "^2.1.1",
     "postcss": "^8.4.21",
-    "react": "^19.2.6",
-    "react-dom": "^19.2.6",
-    "react-error-boundary": "^6.1.2"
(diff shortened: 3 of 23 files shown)


darcyparker:
@msakrejda This one is ready now.

darcyparker:
Did minor rebase ontop of `main`. No changes to PR.

.github/workflows/release.yaml
```diff
@@ -93,12 +93,16 @@ jobs:
...
+          # resolve to real versions (lerna publish leaves catalog: unresolved).
+          pnpm exec lerna version ${VERSION} --yes --force-publish
+          PUBLISH_ARGS=""
           if [[ $VERSION == *"alpha."* ]] || [[ $VERSION == *"beta."* ]] || [[ $VERSION == *"rc."* ]]; then
             echo "Is pre-release version"
-            EXTRA_ARGS="$EXTRA_ARGS --dist-tag next"
+            PUBLISH_ARGS="--tag next"
           fi
-          npx lerna publish ${VERSION} --yes --force-publish $EXTRA_ARGS
+          pnpm -r publish --no-git-checks $PUBLISH_ARGS
```
msakrejda, line 105: I'm definitely not a packaging expert, but based on the [lerna docs](https://lerna.js.org/docs/api-reference/commands) and the [pnpm doc](https://pnpm.io/cli/publish), this seems legit to me.

msakrejda (approved):
This looks good. Once we've merged the full series, I'll publish another alpha and we can verify that everything is still working as expected.

.github/workflows/release.yaml
darcyparker: @msakrejda Sounds good about publishing another alpha. Verifying the full series with an alpha once it's all in makes sense to me too. I'll keep #661 and #663 moving toward ready; if anything looks off when you publish, ping me.

msakrejda: Since #663 is just CI changes, I'll review and merge it later. I'm now publishing [v2.6.0-alpha.2](https://github.com/automerge/automerge-repo/pull/667); I'll let you know if anything looks weird.

darcyparker: @msakrejda - I saw an issue after alpha release.  The `lerna.json` was written by lerna w/o trailing new line. This breaks oxfmt check.  I am creating a quick PR to address this. (ignore the lerna.json for format checks) #668

PR #659, opened 2026-06-01
chore(deps): dependency currency + native-API swaps (part 6 of #653)

darcyparker:
Brings dependencies current, swaps several deps for platform-native APIs, and clears audit/workspace hygiene.

## Native-API swaps (drop a dependency for a native equivalent)

- **isomorphic-ws → native `WebSocket`** in the client; the server keeps `ws` (`WebSocketServer`), now a direct dependency. The keep-alive test drives a raw `ws` client (the native client auto-pongs).
- **rimraf → `node:fs` `rm`** in storage-nodefs.
- **classnames → clsx** in the react-todo example.
- **npm-watch (unmaintained) → native** `tsc --watch` / `vite build --watch` + `vitest --watch`; dropped the per-package npm-watch config blocks.
- **jsdom → happy-dom** (solid-primitives inherits the root happy-dom env).

## Dependency bumps

- **React 19** across dev/examples; the `useRef` in `useDocHandle` gains the now-required argument. `@automerge/react` makes `react` a peer + devDependency (not a direct dependency, so npm consumers don't dual-install); peer stays `^18 || ^19`.
- **uuid 9 → 14** (drop `@types/uuid`; v14 ships its own types; `Uuid.v4`'s first arg `null → undefined`), **bs58check → 4** (ESM), **express → 5** (sync-server), plus current `@manypkg/cli`, `cross-env`, `@testing-library/*`, `@types/*`, and the shared runtime deps.

## Audit / workspace hygiene

- Pin `tmp ^0.2.6` (transitive via lerna → nx; older lines have a path-traversal flaw).
- Fix the `repository` fields (`master → main`) flagged by manypkg.

## Verification

`pnpm build`, `pnpm lint`, `pnpm test` (**819 passing**), and `pnpm repocheck` all green on Node 25 / pnpm 11.5.

---

Part of #653.

PR #661, opened 2026-06-01
chore: examples, scaffolders, docs & dev-tooling cleanup (part 8 of #653)

darcyparker:
The last part of #653: builds & type-checks the examples in CI, modernizes the `create-*` scaffolders, wires up docs, fixes the fuzz harness, and clears the leftover dead dev-tooling.

## Scaffolders
- `create-repo-node-app` moves to template-based scaffolding (a `template/` dir, like `create-vite-app`) instead of inline strings; adds a `test.js` that scaffolds + builds a throwaway project.
- Both templates adopt oxlint over eslint. `create-vite-app` writes `.gitignore` programmatically (npm strips `.gitignore` from published packages, so a template file can't carry one) and drops its dead template `.gitignore` + `tsconfig.node.json`.

## Examples
- Every example gets a `build` script and declares its build deps (`catalog:`), so the root `pnpm -r build` builds them in CI.
- Convert `react-use-presence`'s `vite.config.js` → `vite.config.ts` (+ tsconfig + `main.tsx`), and re-add `lint:check-types:examples` to the root (deferred from #657 until the examples had tsconfigs).

## Docs
- Re-export the public type names (`export type`) so typedoc resolves the `{@link}` cross-references; drop stale JSDoc `@param` tags from the awareness hooks; exclude the scaffolder CLIs from typedoc; add a root `build:docs` (runs clean, 0 errors).

## Fuzz
- Make the harness deterministic (wait on heads-match instead of a fixed `pause(0)`) and fix the share-policy exclusion, which compared the wrong peer id, so Charlie was never actually excluded.

## Dead dev-tooling cleanup (kept as separate commits)
- Drop `c8` (coverage runs through `@vitest/coverage-v8`; the `/* c8 ignore */` source pragmas are honored by the v8 provider, so they stay), `globals` (eslint-only, removed in #656), `dev-null-cli` (no consumer), `ts-node`, and the stale top-level `prettier` config block.
  - **Why `ts-node` lands here:** it was the fuzz runner (`ts-node --esm … fuzz/fuzz.ts`) through #659. #660 migrated fuzz to `tsx` and dropped `ts-node` from `automerge-repo`, which orphaned the root copy, so this is the first PR where removing it is safe.
- Replace `npm-run-all` (unmaintained) with the drop-in `npm-run-all2`.

## Verification
`pnpm build`, `pnpm lint` (oxfmt + oxlint + check-types for packages **and** examples), `pnpm test` (**819 passing**), and `pnpm repocheck` all green on Node 25 / pnpm 11.5.

---

Part of #653.


darcyparker:
@msakrejda This one is ready now.

packages/automerge-repo/fuzz/fuzz.ts
```diff
@@ -106,12 +107,29 @@ for (let i = 0; i < 100000; i++) {
...
+
+  const a = (await aliceRepo.find<TestDoc>(doc.url)).doc()
+  const b = (await bobRepo.find<TestDoc>(doc.url)).doc()
   assert.deepStrictEqual(a, b, "A and B should be equal")
-  assert.deepStrictEqual(b, c, "B and C should be equal")
+  let c: Automerge.Doc<TestDoc> | undefined
+  if (doc.documentId !== notForCharlie) {
+    c = (await charlieRepo.find<TestDoc>(doc.url)).doc()
+    assert.deepStrictEqual(b, c, "B and C should be equal")
+  }
```
msakrejda, line 140: Should we assert that Charlie *didn't* get the doc in the `else` branch?

packages/create-repo-node-app/test.js
```diff
@@ -0,0 +1,62 @@
...
+await $$`./node_modules/@automerge/create-repo-node-app/dist/index.js test-app`
+
+const cwd = path.join(tempDir, "test-app")
+const $$$ = $({ cwd })
+
+console.log("installing test app...")
+// pnpm 11 gates optional native build scripts (e.g. cbor-extract via cbor-x)
+// pending approval and exits non-zero; dependencies still install and the demo
+// runs without that optional build, so don't treat the gate as a failure.
+await $$$`pnpm install`.catch(() => {})
```
msakrejda: Do we want to catch and ignore _everything_ here?

packages/create-vite-app/test.js
```diff
@@ -1,64 +1,51 @@
...
-})
-
-const success = output.includes("VITE") && output.includes("ready")
-
-// cleanup
-await $`git clean *.tgz -f`
+// pnpm 11 gates optional native build scripts (e.g. cbor-extract via cbor-x)
+// pending approval and exits non-zero; dependencies still install and the dev
+// server runs without that optional build, so don't treat the gate as a failure.
+await $$$`pnpm install`.catch(() => {})
```
msakrejda: Same as above: do we want to ignore _all_ errors?

msakrejda (commented):
A couple of minor questions, but otherwise this looks great.

packages/automerge-repo/fuzz/fuzz.ts
darcyparker: Good point. Added it: the else branch now asserts Charlie's `find` resolves to unavailable (with a bounded abort so a silent peer can't hang the run), so a broken share policy would be caught instead of silently skipped.

packages/create-repo-node-app/test.js
darcyparker: Fair point. Narrowed it: it now tolerates only the build-script-approval gate (by checking `node_modules` was actually populated) and re-throws any genuine install failure.

packages/create-vite-app/test.js
darcyparker: Same fix as the create-repo-node-app one: it now tolerates only the approval gate (via the `node_modules` check) and re-throws a real install failure rather than swallowing everything.

packages/automerge-repo/fuzz/fuzz.ts
msakrejda: Okay, great. I think there's still a chance that if we reject because of a silent peer, this will be a false negative (i.e., the test will still pass), but it's probably not worth worrying about.

msakrejda (approved):
Thanks, looks great now.

packages/automerge-repo/fuzz/fuzz.ts
darcyparker: Agreed, that's a real gap. With the bounded abort, a silent peer (timeout) also satisfies `assert.rejects`, so it was asserting "Charlie didn't get it in time" rather than "the share policy returned unavailable." Since it's a one-liner I went ahead and tightened it in #669: it now matches the rejection on `/unavailable/`, so a silent-peer timeout fails instead of passing. Thanks for flagging it.

PR #663, opened 2026-06-02
ci: split the Install and Build step into separate steps

darcyparker:
Splits the combined `Install and Build` CI step into separate `Install` and `Build` steps, so the per-step timing (install vs build) is visible in each run's summary.

Per review feedback from @psmolaga on #657.

Splits the combined step in all four workflows that carry it:
- `lint.yaml` (the build-before-lint step added in #657)
- `tests.yaml`
- `playwright.yaml`
- `release.yaml`'s docs job (the publish job already used separate Install/Build steps)

---

Part of #653.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -40,9 +40,9 @@ jobs:
           node-version: 24
           registry-url: "https://registry.npmjs.org"
           cache: pnpm
-      - name: Install and Build
-        run: |
-          pnpm install
-          pnpm build
+      - name: Install
+        run: pnpm install
+      - name: Build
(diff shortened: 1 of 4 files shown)


darcyparker:
@msakrejda This one is ready now. It is the last of the #653 series.

darcyparker:
@msakrejda - I see https://github.com/automerge/automerge-repo/releases/tag/v2.6.0-alpha.2 So looks like the test alpha release went well.  This PR is minor improvement in my opinion. It should be easy to review/merge but no urgency to do another alpha release.

msakrejda (approved):

PR #669, opened 2026-06-05
test(fuzz): assert the excluded doc is unavailable to charlie

darcyparker:
Follow-up to the review on #661: tightens the fuzz harness's excluded-doc assertion.

Charlie is excluded from `notForCharlie` by the share policy, so its `find` should reject as **unavailable**. The original assertion accepted any rejection, so a silent peer (which would instead abort via the bounded timeout) would still pass it, the false negative @msakrejda flagged on #661. This matches the rejection on `/unavailable/` so a silent-peer timeout fails the assertion rather than passing it.

The fuzz change is fuzz-only and type-checks (the fuzz harness is excluded from oxlint).


msakrejda (approved):
Great, thanks for the follow-up patch.

PR #671, opened 2026-06-05
fix(StorageSource): settle the storage source when a load fails

darcyparker:
## Problem

`StorageSource.attach` loads the document with a fire-and-forget `loadDoc().then(...)` that has no `.catch`. If the storage read rejects (corrupt store, disk/IO error, or a failed fetch in a remote-backed adapter), the `"storage"` source is never marked ready or unavailable, so it stays `pending`. As a result:

- the `DocumentQuery` is stuck `loading`, so `whenReady()` / `Repo.find()`   never settle, and
- the dropped rejection surfaces as an `unhandledRejection`.

## Fix

Add a `.catch` to the `loadDoc()` chain in `StorageSource.attach`. On a failed read (or a throw while applying the loaded data), mark the `"storage"` source unavailable so the query can settle, and log the error instead of dropping it. Other sources (e.g. sync) can still deliver the document, so a storage read error no longer blocks a find that a peer could satisfy.

## Commits

Split into two commits:

1. `test(StorageSource)` adds a failing test that reproduces the hang (the query stays `loading` and the rejection goes unhandled).
2. `fix(StorageSource)` adds the `.catch` that makes it pass.

## Related

- #676 - the save-side sibling in `StorageSource` (this PR is the load side). They overlap on the `#log` field, a trivial rebase for whichever merges second.
- #672 / #673 / #674 - the same unhandled-rejection / uncaught-error class in `StorageSubsystem`, `Repo`, and the WebSocket server adapter.

### DIFF
--- a/packages/automerge-repo/test/StorageSource.test.ts
+++ b/packages/automerge-repo/test/StorageSource.test.ts
@@ -1,6 +1,6 @@
 import { next as A } from "@automerge/automerge"
 import assert from "assert"
-import { describe, it } from "vitest"
+import { describe, it, vi } from "vitest"
 import { generateAutomergeUrl, parseAutomergeUrl } from "../src/AutomergeUrl.js"
 import { DocumentQuery } from "../src/DocumentQuery.js"
 import { StorageSource } from "../src/StorageSource.js"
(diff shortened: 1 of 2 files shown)


darcyparker:
cc @alexjg @pvh: flagging this since you last touched `StorageSource.ts`.

packages/automerge-repo/test/StorageSource.test.ts
```diff
@@ -73,4 +73,46 @@ describe("StorageSource", () => {
...
+      // (rejected) microtask chain deterministically (no real-time wait).
+      await vi.runAllTimersAsync()
+
+      // A failed storage load must settle the "storage" source as unavailable
+      // so the query can resolve. With storage the only source and no handle
+      // data, the query becomes "unavailable" and whenReady() rejects.
+      //
+      // The bug: the failed loadDoc() rejection is dropped (no .catch), so
+      // "storage" is left in "pending" forever, so the query stays "loading",
+      // whenReady() never settles, and an unhandledRejection is emitted.
```
msakrejda:
```suggestion
```
I don't think comments like this are useful: it's talking about a bug that's now (as of the second commit here) fixed and code that that has changed (there is a `.catch` now). Having this detail in the commit message is fine, but in a comment I think it's more confusing than helpful.

msakrejda (approved):
Thanks for the fix! I think part of the test comment should be dropped (I can just commit my suggestion unless you feel strongly otherwise), but otherwise this looks good.

packages/automerge-repo/test/StorageSource.test.ts
darcyparker: Good point: I'll address in each of the PRs open.

PR #672, opened 2026-06-06
fix(StorageSubsystem): reset the compacting flag in a finally

darcyparker:
## Problem

`StorageSubsystem.#saveTotal` (compaction) sets the `#compacting` flag, writes the snapshot and removes the old chunks, then clears the flag, with no try/finally. If a write rejects (disk/IO error, IndexedDB quota or an aborted transaction, or a failed fetch in a remote-backed adapter), the flag stays set. `#shouldCompact` returns `false` whenever the flag is set, so compaction is **permanently disabled** for that subsystem: every later save falls back to an incremental, and incremental chunks grow without bound (larger storage, slower loads). No data is lost, but the document is never re-consolidated, and the condition persists until the process restarts.

## Fix

Wrap the body of `#saveTotal` in `try/finally` so the flag is always cleared, even when a write throws. Compaction then resumes on the next save once storage recovers; a transient storage hiccup should not disable compaction forever.

## Commits

1. `test(StorageSubsystem)` reproduces the bug: a flaky adapter fails the second  snapshot write, then asserts the next save compacts again and that a reload still sees the latest state (no data loss). It fails on `main`.
2. `fix(StorageSubsystem)` adds the `try/finally`.

## Known limitation (pre-existing, minor)

The flag is a single boolean on the shared subsystem, and `#saveIncremental`'s no-prior-save fallback can enter `#saveTotal` even while it is set, so two documents can briefly compact at once and whichever finishes first clears the
flag. This is best-effort serialization, not a lock. It does not corrupt data (each document writes its own keys; same-document saves are serialized by the throttle in `StorageSource`), so the only effect is occasional concurrent compaction of different documents. A per-document guard would tighten it but is not worth the added complexity for that minor effect.

### DIFF
--- a/packages/automerge-repo/test/StorageSubsystem.test.ts
+++ b/packages/automerge-repo/test/StorageSubsystem.test.ts
@@ -536,3 +536,69 @@ describe("StorageSubsystem", () => {
     })
   })
 })
+
+describe("StorageSubsystem compaction recovery", () => {
+  it("keeps compacting after a failed snapshot write (does not get stuck not-compacting)", async () => {
+    // Simulate a storage write that fails mid-compaction. Real, usually
+    // transient causes:
+    //   - IndexedDB: quota exceeded, or a transaction aborted under memory
+    //     pressure (often succeeds on retry).
+    //   - Remote/HTTP-backed adapter: a 503, a timeout, or an expired auth
+    //     token (refreshed before the next save).
+    //   - Local filesystem: disk full, or a momentary lock from another tab.
+    // This adapter fails the second snapshot write and succeeds otherwise.
+    class FlakySnapshotAdapter extends DummyStorageAdapter {
+      snapshotSaves = 0
+      override async save(key: string[], binary: Uint8Array) {
+        if (key[1] === "snapshot") {
+          this.snapshotSaves++
+          if (this.snapshotSaves === 2) {
+            throw new Error("snapshot write failed")
+          }
+        }
+        return super.save(key, binary)
+      }
+    }
+
+    const adapter = new FlakySnapshotAdapter()
+    const storage = new StorageSubsystem(adapter)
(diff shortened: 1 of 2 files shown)


darcyparker:
cc @alexjg, flagging this small StorageSubsystem compaction fix for your review.

packages/automerge-repo/test/StorageSubsystem.test.ts
```diff
@@ -536,3 +536,73 @@ describe("StorageSubsystem", () => {
...
+    doc = A.change(doc, d => {
+      d.n = 2
+    })
+    await expect(storage.saveDoc(documentId, doc)).rejects.toThrow(
+      "snapshot write failed"
+    )
+
+    // Save 3: should compact again. The bug leaves the internal compacting flag
+    // stuck true, which permanently disables compaction, so this falls back to
+    // an incremental write and snapshotSaves stays at 2.
```
msakrejda: Similarly to #671, I don't think this (or the comment below) should be talking about "the bug", since it will be fixed when this is merged.

msakrejda (commented):
I think this fix is also fairly straightforward and I can just merge it after we resolve the comments, unless you specifically want @alexjg's eyes on it.

packages/automerge-repo/test/StorageSubsystem.test.ts
darcyparker: Thanks. Good point - I will address in all of my PRs like this that are open.

darcyparker:
> I think this fix is also fairly straightforward and I can just merge it after we resolve the comments, unless you specifically want @alexjg's eyes on it.

No need.  #671, #672, #673, #674, #676 are each the same type of issue. Thanks for reviewing!  I hope these will improve robustness issues on our sync server.

msakrejda (approved):

PR #674, opened 2026-06-06
fix(WebSocketServerAdapter): catch errors thrown while handling a message

darcyparker:
## Problem

`WebSocketServerAdapter.receiveMessage` decodes untrusted bytes from a peer and runs in the socket `"message"` event-loop callback. `decode()` is guarded (it closes the connection on a corrupt frame), but the handling after it was not: a decoded-but-malformed message throws there. Concretely, a peer can crash the server with a single byte: a frame containing CBOR null (`0xf6`) decodes to `null`, and `const { type, senderId } = message` throws. The `"message"` listener is registered on `connection`, before any join handshake, so this needs no prior message and no authentication.

That throw is not trapped, so it reaches the event loop and, in Node, terminates the process by default. Same class of bug as #673 (inbound `"message"` handler); see that PR for the EventEmitter / `uncaughtException` mechanism and a runnable
reproduction. It is also not hypothetical here: #297 was a host-process crash from a throw escaping this adapter's event-driven send path, fixed by not throwing there. This PR addresses the inbound (receive) side.

## Fix (2 commits)

1. `test`: deliver CBOR `null` to `receiveMessage` and assert it does not throw (and closes the connection). Red until the fix.
2. `fix`: wrap the post-decode body in `try/catch` (the `decode()` guard already handles corrupt frames) and close the connection on failure.

## Related

- #673 - `Repo` inbound `"message"` handler (mechanism + reproduction)
- #676 - the same class on the `StorageSource` save path (a storage write rejection floating from a "heads-changed" listener)

### DIFF
--- a/packages/automerge-repo-network-websocket/src/WebSocketServerAdapter.ts
+++ b/packages/automerge-repo-network-websocket/src/WebSocketServerAdapter.ts
@@ -140,53 +140,68 @@ export class WebSocketServerAdapter extends NetworkAdapter {
       return
     }
 
-    const { type, senderId } = message
+    // A decoded message is still untrusted: a malformed one (e.g. CBOR null,
+    // which makes the destructure below throw) can throw while being handled.
+    // This runs in the socket "message" event-loop callback, so an uncaught
+    // throw would terminate a Node process by default. Drop the message and
+    // close the misbehaving connection.
+    // See https://nodejs.org/api/process.html#event-uncaughtexception
+    try {
+      const { type, senderId } = message
 
-    const myPeerId = this.peerId
-    assert(myPeerId)
+      const myPeerId = this.peerId
+      assert(myPeerId)
 
-    const documentId = "documentId" in message ? "@" + message.documentId : ""
-    const { byteLength } = messageBytes
-    log(`[${senderId}->${myPeerId}${documentId}] ${type} | ${byteLength} bytes`)
+      const documentId = "documentId" in message ? "@" + message.documentId : ""
+      const { byteLength } = messageBytes
+      log(
+        `[${senderId}->${myPeerId}${documentId}] ${type} | ${byteLength} bytes`
+      )
 
-    if (isJoinMessage(message)) {
-      const { peerMetadata, supportedProtocolVersions } = message
-      const existingSocket = this.sockets[senderId]
-      if (existingSocket) {
-        if (existingSocket.readyState === WebSocket.OPEN) {
-          existingSocket.close()
+      if (isJoinMessage(message)) {
+        const { peerMetadata, supportedProtocolVersions } = message
+        const existingSocket = this.sockets[senderId]
+        if (existingSocket) {
+          if (existingSocket.readyState === WebSocket.OPEN) {
+            existingSocket.close()
+          }
+          this.emit("peer-disconnected", { peerId: senderId })
         }
-        this.emit("peer-disconnected", { peerId: senderId })
-      }
 
-      // Let the repo know that we have a new connection.
-      this.emit("peer-candidate", { peerId: senderId, peerMetadata })
-      this.sockets[senderId] = socket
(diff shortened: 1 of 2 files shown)


darcyparker:
cc @alexjg. Sibling of #673 (same uncaughtException crash class), on the WebSocket server adapter inbound decode path.

darcyparker:
## Open question: `catch` recovery behavior (close vs log-and-drop)

The `try` is the crash fix; the `catch` body is a policy choice I would appreciate a second opinion on.

What actually reaches this `catch` is almost always a broken or malicious peer, not valid traffic: the realistic trigger is a frame that decodes to `null`/`undefined` (CBOR `0xf6`/`0xf7`), because `const { type, senderId } = message` only throws on `null`/`undefined`. A legitimate client (even a cross-version one) sends a message object, which does not throw there; it yields `undefined` fields and falls through `isJoinMessage` to the existing `return`. So the real question is how to treat a peer that sent garbage.

**A. `socket.close()` (what this PR does)**
- Consistent with the `decode()` failure path right above, which already closes on a corrupt frame.
- Cleanup is automatic: `connect()` registers `socket.on("close", () => this.#removeSocket(socket))`, which emits `peer-disconnected` and clears `this.sockets`.
- Sheds the misbehaving peer.
- Against: tears down the whole connection, aborting in-flight sync for every document on it (the peer must reconnect and re-sync); and if a buggy client deterministically resends, `close -> reconnect -> resend -> close` is a churn loop (a malicious peer reconnects regardless).

**B. log and drop, keep the connection (what #673 does for the Repo handler)**
- Smaller blast radius: one bad frame disturbs nothing else on the connection.
- Consistent with the sibling fix #673 ("isolate one bad message, keep serving").
- No reconnect churn.
- Against: a persistently-broken peer stays connected (we keep decoding and dropping its frames).

A third option, replying with `{ type: "error" }` before closing, is not reliable here: the throw can happen at the `senderId` destructure, and `send` needs a target, so we often cannot tell the peer why.

I went with A for local consistency with the decode path, but I am happy to switch to B if you prefer matching #673. Either is a large improvement over the current crash.

packages/automerge-repo-network-websocket/src/WebSocketServerAdapter.ts
```diff
@@ -140,53 +140,68 @@ export class WebSocketServerAdapter extends NetworkAdapter {
...
-        return
+        if (this.sockets[senderId] !== socket) {
+          return
+        }
+        this.emit("message", message)
       }
-      this.emit("message", message)
+    } catch (e) {
+      log("error handling inbound message, closing connection", e)
+      socket.close()
```
msakrejda, line 204: If one of the `.send`s in the `isJoinMessage` branch throws, the socket could be left abandoned in `this.sockets`, right? I think that's okay, though, since a subsequent join message would just replace it.

msakrejda (approved):
I think closing the socket is the right call, per your reasoning. I'll leave this open for a bit for other feedback, but otherwise I plan to merge early next week.

packages/automerge-repo-network-websocket/src/WebSocketServerAdapter.ts
darcyparker: Good question. The socket is registered in `this.sockets[senderId]` (line 173) before those sends, so a throwing `send` lands in the new `catch`, which calls `socket.close()`. That fires the `socket.on("close")` handler from the connection setup, which calls `#removeSocket` and deletes it from `this.sockets`, so it's cleaned up rather than abandoned (the entry lingers only for the brief async window until "close" fires, during which `send()` to that peer just drops on the non-OPEN readyState check). And as you say, a re-join would replace it anyway. It's also strictly better than before this PR, where a throwing send escaped the handler entirely.

msakrejda (approved):
Thanks, this looks good to me. I'll merge soon.

PR #676, opened 2026-06-06
fix(StorageSource): catch errors from the throttled save

darcyparker:
## Problem

`StorageSource` persists on every `"heads-changed"` event via a throttled save:

    handle.on("heads-changed", saveFn)   // saveFn = asyncThrottle(() => storage.saveDoc(...))

`asyncThrottle` returns a promise that rejects when `saveDoc` rejects, and that promise is ignored (the listener's return value is dropped; the bare call at attach is not awaited). So a storage write failure (disk/IO error, quota, an aborted IndexedDB transaction, a failed remote write) floats as an unhandled rejection, which in Node exits the process by default ([unhandled-rejections][1]).

This is the save side of #389 ("Storage errors cause node to exit") and the same class as the intermittent unhandled-rejection failures in #275. The load side is handled in #671; this is its sibling.

## Fix (2 commits)

1. `test`: point `StorageSource` at an adapter whose `save()` rejects, make a  change, and assert the failure is caught and logged. Red until the fix (it fails the assertion and Vitest reports the floating unhandled rejection).
2. `fix`: wrap `saveDoc` in `try/catch` inside the throttled function and log. The change stays in memory, so a later save or reload can re-persist it.

## Related

- #671 - load-side sibling in `StorageSource` (lightly overlaps: both add the `#log` field, a trivial rebase for whichever merges second)
- #672 - `StorageSubsystem` compaction error handling
- #673 / #674 - same crash class in the `Repo` inbound-message and WebSocket-server handlers
- Issues: #389 (storage errors exit the process), #275 (intermittent unhandled-rejection CI failures), #264 (durability)

[1]: https://nodejs.org/api/process.html#event-unhandledrejection

### DIFF
--- a/packages/automerge-repo/src/StorageSource.ts
+++ b/packages/automerge-repo/src/StorageSource.ts
@@ -89,8 +89,24 @@ export class StorageSource implements DocumentSource {
     let fn = this.#saveFns[documentId]
     if (!fn) {
       fn = this.#saveFns[documentId] = asyncThrottle(
-        ({ doc, handle }: DocHandleEncodedChangePayload<any>): Promise<void> =>
-          this.#storage.saveDoc(handle.documentId, doc),
+        async ({
+          doc,
+          handle,
+        }: DocHandleEncodedChangePayload<any>): Promise<void> => {
+          try {
+            await this.#storage.saveDoc(handle.documentId, doc)
+          } catch (err) {
+            // This save runs fire-and-forget from a "heads-changed" listener,
+            // so a rejection would surface as an unhandled rejection and, in
+            // Node, exit the process by default. Catch and log it; the change
+            // stays in memory and a later save or reload can re-persist it.
+            // See https://nodejs.org/api/process.html#event-unhandledrejection
+            this.#log.error(
+              `Error saving document ${handle.documentId} to storage`,
+              err
+            )
+          }
+        },
         this.#saveDebounceRate
       )
     }
--- a/packages/automerge-repo/test/StorageSource.test.ts
+++ b/packages/automerge-repo/test/StorageSource.test.ts
@@ -111,4 +111,40 @@ describe("StorageSource", () => {
       vi.useRealTimers()
     }
   })
+
+  it("logs a failed save instead of leaking an unhandled rejection", async () => {
+    // A save runs fire-and-forget from the "heads-changed" listener. If the
+    // storage write rejects, the rejection must be caught and logged, not left
+    // to surface as an unhandled rejection.
+    const errSpy = vi.spyOn(console, "error").mockImplementation(() => {})
+    try {
+      const documentId = parseAutomergeUrl(generateAutomergeUrl()).documentId
+
+      // An adapter whose save() always rejects (disk/IO error, quota, ...).
+      const adapter = new DummyStorageAdapter()
+      adapter.save = async () => {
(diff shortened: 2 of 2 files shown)


darcyparker:
cc @alexjg. Save-side sibling of #671 (load); same unhandled-rejection class as #673/#674. Addresses the save path called out in #389.

msakrejda (approved):

packages/automerge-repo/src/StorageSource.ts
```diff
@@ -89,8 +89,24 @@ export class StorageSource implements DocumentSource {
...
+          } catch (err) {
+            // This save runs fire-and-forget from a "heads-changed" listener,
+            // so a rejection would surface as an unhandled rejection and, in
+            // Node, exit the process by default. Catch and log it; the change
+            // stays in memory and a later save or reload can re-persist it.
+            // See https://nodejs.org/api/process.html#event-unhandledrejection
+            this.#log.error(
+              `Error saving document ${handle.documentId} to storage`,
+              err
+            )
```
msakrejda, line 107: I wonder if we  want to have some cap on this behavior: with this change, the storage adapter could totally fail to store any changes, but this would only be visible in the logs. Maybe at some point that should be escalated somehow?

msakrejda (commented):
I'm a little concerned that repeated failures could lead to a situation where a sync server seems to work (unless you're watching logs), but is not actually storing anything. Thoughts?

packages/automerge-repo/src/StorageSource.ts
darcyparker: Fair concern, and I don't want to wave it away. The storage adapter should be robust, but some failures are rare and recoverable, and what this change buys is time to heal: by not letting a failed write take the process down, and keeping the change in memory, a later save or a reload can re-persist it.

On where recovery and escalation should sit, my thinking:

1. Network adapters provide redundancy: connected peers still have the change and can re-sync it.
2. The adapter gets another shot on the next save, since the in-memory change is re-included.
3. The error is logged, so an observability/alerting layer can surface persistent failures to an operator.

For exponential backoff, and for deciding when a failure is unrecoverable enough to escalate, I'd put that in the storage adapter rather than `StorageSource`: the adapter knows its backend (IndexedDB vs filesystem vs an HTTP-backed store), so it's the right place for that policy. A generic cap in `StorageSource` would bake one policy into a layer that can't know the backend's characteristics.

So my inclination is to keep `StorageSource`'s role narrow: keep the process alive and hold onto the data, which maximizes the window for peers to sync and the adapter to recover, and treat backoff/escalation as an adapter responsibility (and alerting as an observability one). Happy to write that expectation down in a follow-up so it's explicit for adapter authors. What do you think?

msakrejda: Okay, I think that makes sense. We may want to document the storage adapter contract more carefully with respect to errors, as you suggest, but we can do that as a follow-up. I'll merge soon unless anyone else has thoughts.

darcyparker: Opened the follow-up: #679. It writes down the contract we discussed: `StorageSource` keeps the process up and retains the change in memory, storage adapters own retry/backoff/escalation, reliability comes from peer redundancy, and observability goes through the pluggable logger (so no first-class storage-error signal). Thanks!

PR #679, opened 2026-06-11
docs: storage error handling and the storage adapter contract

darcyparker:
## What

Adds a dev-docs page, `packages/automerge-repo/dev-docs/storage-error-handling.md`, documenting what happens when storage I/O fails and where the responsibility for reliability lies. This is the follow-up suggested in the review discussion on #676. Doc-only; no code changes.

## Contents

- **What `StorageSource` guarantees:** a failed save does not crash the process, and the change is retained in memory so a later save or reload re-persists it. It converts a fatal crash into a recoverable condition and does nothing more.
- **The storage adapter contract:** adapters own robustness; retry and backoff are one consideration (not a mandate, and not always the best fit); escalation is the adapter's responsibility, since only the adapter knows its backend.
- **Reliability strategies:** network redundancy via peers is usually the strongest lever, since a document synced to the repo instances you connect to is already durable in more than one place; adapter retry/backoff and other strategies (multi-backend write-through, a durable queue, reconciliation) are options to weigh.
- **Observability and alerting:** failures surface via the pluggable logger (`setLoggerFactory`, shaped for console/pino/winston); shipping logs (for example via OpenTelemetry) to an alerting layer is a deployment concern owned by the application embedding automerge-repo.
- **Why there is no first-class storage error event:** the logging path already exists and is configurable, escalation belongs in the adapter, and redundancy comes from the network; a custom `LoggerFactory` is the hook for programmatic handling.

### DIFF
--- a/packages/automerge-repo/dev-docs/storage-error-handling.md
+++ b/packages/automerge-repo/dev-docs/storage-error-handling.md
@@ -1,0 +1,116 @@
+# Storage error handling
+
+Guidance on what happens when storage I/O fails in `automerge-repo`, and where
+the responsibility for reliability lies. The behavior described here is
+implemented in [`src/StorageSource.ts`](../src/StorageSource.ts); the adapter
+contract applies to implementations of
+[`src/storage/StorageAdapterInterface.ts`](../src/storage/StorageAdapterInterface.ts).
+
+## What `StorageSource` guarantees
+
+When a throttled save rejects (a disk/IO error, quota exceeded, an aborted
+IndexedDB transaction, a failed remote write, and so on), `StorageSource`
+catches the error and logs it rather than letting it escape. Two properties
+follow:
+
+- **The process stays up.** The save runs fire-and-forget from a
+  `"heads-changed"` listener, so an unhandled rejection would, in Node, exit the
+  process by default. Catching it keeps the repo alive.
+- **The change is retained in memory.** Nothing is dropped. A later save or a
+  reload re-persists the current document, because `lastSavedHeads` only
+  advances on a successful write, so the next successful save re-includes
+  whatever a failed one did not persist.
+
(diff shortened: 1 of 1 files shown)


msakrejda:
Cool, thanks! Broadly, I think these docs are good. I'm not sure if this is the right place for them. It might be, but we're not really doing that for other stuff. Other options:

 - a note in https://automerge.org/docs/reference/repositories/storage/ (maybe not the right place either, since it doesn't talk about implementation much)
 - comments on StorageAdapterInterface (though maybe this is too high-level for something like that)

I'll sleep on it.

darcyparker:
> Cool, thanks! Broadly, I think these docs are good. I'm not sure if this is the right place for them. It might be, but we're not really doing that for other stuff. Other options:
> 
> * a note in https://automerge.org/docs/reference/repositories/storage/ (maybe not the right place either, since it doesn't talk about implementation much)
> * comments on StorageAdapterInterface (though maybe this is too high-level for something like that)
> 
> I'll sleep on it.

NP - Use it as a reference. Feel free to close this PR.

msakrejda:
So I ran into related issues that prompted me to write #725; I think it'd be nice to have these docs land. I'm still not sure what the best place for it: I think ideally, it'd be good to have docs in source comments, and separately something in the public docs. After #725 lands I can try pulling apart this into those pieces unless you want to take a stab at it.

darcyparker:
Agreed, let's land these. I'll take a stab at the split after #725 merges (per your suggestion): source comments plus a docs page for the load-path semantics (`DocumentUnavailableError` vs `DocumentLoadFailedError`), and we can settle placement (this `dev-docs` file, the storage reference, or elsewhere) then. Fuller notes are on #725 (and #726, a small follow-up there making `DocumentLoadFailedError` extend `AggregateError` so structured loggers expand the sub-errors).

PR #689, opened 2026-06-26
fix(throttle): settle superseded asyncThrottle calls instead of orphaning them

darcyparker:
## What

`asyncThrottle` coalesces calls that arrive within the throttle window so that only the latest one runs. Each call used to create its own `Promise` wired to its own `setTimeout`. When a later call arrived it cleared the earlier call's timeout, but nothing ever resolved or rejected the earlier call's promise, so awaiting a superseded call hung forever. (A concern for GC)

```ts
const throttled = asyncThrottle(fn, 100)
const a = throttled(1) // superseded: its timeout is cleared by the next call
const b = throttled(2) // runs
await a // never settles
```

## Fix

Share a single deferred (`Promise.withResolvers`) across every call coalesced into the next run. The run still uses the latest call's args, and the existing `await currentPromise` serialization (each call waits for the previous execution to settle) is unchanged. The difference is that when the run settles, every caller coalesced into it settles with that same result, instead of only the last caller's promise being wired to the timeout.

```diff
-    return new Promise<TReturn>((resolve, reject) => {
-      timeout = setTimeout(async () => {
-        try {
-          currentPromise = fn(...args)
-          resolve(await currentPromise)
-        } catch (error) {
-          reject(error)
-        } finally {
-          lastCall = Date.now()
-          currentPromise = undefined
-          timeout = undefined
-        }
-      }, wait)
-    })
+    pending ??= Promise.withResolvers<TReturn>()
+    const deferred = pending
+    ...
+    timeout = setTimeout(async () => {
+      pending = undefined
+      timeout = undefined
+      try {
+        currentPromise = fn(...args)
+        deferred.resolve(await currentPromise)
+      } catch (error) {
+        deferred.reject(error)
+      } finally {
+        lastCall = Date.now()
+        currentPromise = undefined
+      }
+    }, wait)
+    return deferred.promise
```

## Tests

Adds two regression tests covering the coalesced callers:

- a superseded caller's promise resolves with the winning run's result
- a rejection propagates to every superseded caller

`vitest run test/throttle.test.ts`: 12 passed.

Follows #675, which clamped the same throttle's `setTimeout` delay.

### DIFF
--- a/packages/automerge-repo/src/helpers/throttle.ts
+++ b/packages/automerge-repo/src/helpers/throttle.ts
@@ -50,6 +50,8 @@ export const throttle = <F extends (...args: Parameters<F>) => ReturnType<F>>(
  * - Previous calls complete before new ones start (so there is no race with previous calls)
  * - There's always a minimum delay between executions
  * - The latest call always runs (canceling previous pending calls)
+ * - Superseded calls still settle — every coalesced caller resolves (or rejects)
+ *   with the winning run's result rather than being left with an orphaned promise
  * - Each call waits for the previous execution to complete
  *
  * This creates a batching behavior that prevents flooding while ensuring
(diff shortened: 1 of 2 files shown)


packages/automerge-repo/src/helpers/throttle.ts
```diff
@@ -50,6 +50,8 @@ export const throttle = <F extends (...args: Parameters<F>) => ReturnType<F>>(
  * - Previous calls complete before new ones start (so there is no race with previous calls)
  * - There's always a minimum delay between executions
  * - The latest call always runs (canceling previous pending calls)
+ * - Superseded calls still settle — every coalesced caller resolves (or rejects)
+ *   with the winning run's result rather than being left with an orphaned promise
```
msakrejda, line 54: Yeah, this is probably more useful than having superseded calls always reject.

msakrejda (approved):
Nice, good catch.

packages/automerge-repo/src/helpers/throttle.ts
darcyparker: Yes - in most cases, the caller should never reject...  (it settles the way the caller settled with)

PR #690, opened 2026-06-27
fix(websocket): clear timers and listeners on disconnect

darcyparker:
## What

`WebSocketServerAdapter` left its keepalive `setInterval` running after `disconnect()` (it was only cleared if the underlying `WebSocketServer` emitted `close`), and registered `connection`/`close` listeners on the shared server that were never removed, so a second `connect()` stacked duplicates and orphaned the earlier interval. On a long-running sync server these accumulate.

`WebSocketClientAdapter` never tracked the reconnect `setTimeout` armed in `onClose()` or the forceReady fallback `setTimeout`, so a reconnect scheduled just before `disconnect()` would still fire and revive a torn-down adapter. `disconnect()` was also not idempotent: the first call clears `this.socket`, so a second call tripped `assert(this.socket)` and threw. That surfaces under React StrictMode, which double-invokes the effect cleanup that runs `repo.shutdown()`. https://discord.com/channels/1200006940210757672/1497472347869806624

## Fix

- Server: track the keepalive interval and the server listeners on the instance and tear them down in `disconnect()`; a repeated `connect()` no longer stacks them.
- Client: track and cancel both timers in `disconnect()`, reset `#retryIntervalId` so a later `connect()` can re-arm retries, and use a `#disconnected` flag both to stop a pending reconnect from reviving the socket and to make `disconnect()` idempotent (a second call returns early instead of throwing).

## Tests

Adds regression tests: the server adapter removes its listeners and clears the keepalive interval on `disconnect()`, and the client adapter is safe to `disconnect()` more than once. Existing suite passes.

### DIFF
--- a/packages/automerge-repo-network-websocket/src/WebSocketClientAdapter.ts
+++ b/packages/automerge-repo-network-websocket/src/WebSocketClientAdapter.ts
@@ -45,6 +45,14 @@ export class WebSocketClientAdapter extends WebSocketNetworkAdapter {
   }
 
   #retryIntervalId?: TimeoutId
+  // The one-shot reconnect timer armed in onClose() and the forceReady fallback
+  // timer armed in connect(), held on the instance so disconnect() can cancel
+  // them.
+  #reconnectTimeoutId?: TimeoutId
+  #forceReadyTimeoutId?: TimeoutId
+  // Set by disconnect(); guards the reconnect timer so a pending reconnect that
+  // fires after disconnect() does not revive the socket.
+  #disconnected = false
   #log = debug("automerge-repo:websocket:browser")
 
   remotePeerId?: PeerId // this adapter only connects to one remote client at a time
@@ -58,6 +66,7 @@ export class WebSocketClientAdapter extends WebSocketNetworkAdapter {
   }
 
   connect(peerId: PeerId, peerMetadata?: PeerMetadata) {
+    this.#disconnected = false
     if (!this.socket || !this.peerId) {
       // first time connecting
       this.#log("connecting")
@@ -89,8 +98,10 @@ export class WebSocketClientAdapter extends WebSocketNetworkAdapter {
 
     // Mark this adapter as ready if we haven't received an ack in 1 second.
     // We might hear back from the other end at some point but we shouldn't
-    // hold up marking things as unavailable for any longer
-    setTimeout(() => this.#forceReady(), 1000)
+    // hold up marking things as unavailable for any longer. Clear any pending
+    // fallback so reconnects don't accumulate timers.
+    clearTimeout(this.#forceReadyTimeoutId)
+    this.#forceReadyTimeoutId = setTimeout(() => this.#forceReady(), 1000)
     this.join()
   }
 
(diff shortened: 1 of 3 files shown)


packages/automerge-repo-network-websocket/src/WebSocketClientAdapter.ts
```diff
@@ -45,6 +45,14 @@ export class WebSocketClientAdapter extends WebSocketNetworkAdapter {
   }
 
   #retryIntervalId?: TimeoutId
+  // The one-shot reconnect timer armed in onClose() and the forceReady fallback
+  // timer armed in connect(), held on the instance so disconnect() can cancel
+  // them.
```
msakrejda:
```suggestion
```

packages/automerge-repo-network-websocket/src/WebSocketClientAdapter.ts
```diff
@@ -45,6 +45,14 @@ export class WebSocketClientAdapter extends WebSocketNetworkAdapter {
...
+  #reconnectTimeoutId?: TimeoutId
+  #forceReadyTimeoutId?: TimeoutId
+  // Set by disconnect(); guards the reconnect timer so a pending reconnect that
+  // fires after disconnect() does not revive the socket.
```
msakrejda:
```suggestion
```

packages/automerge-repo-network-websocket/src/WebSocketClientAdapter.ts
```diff
@@ -138,8 +151,13 @@ export class WebSocketClientAdapter extends WebSocketNetworkAdapter {
   }
 
   disconnect() {
+    // disconnect() can be called more than once (e.g. React StrictMode double-
+    // invokes the effect cleanup that runs repo.shutdown()); make it idempotent
+    // rather than tripping the assert below on the second call.
```
msakrejda:
```suggestion
```

packages/automerge-repo-network-websocket/src/WebSocketClientAdapter.ts
```diff
@@ -148,7 +166,15 @@ export class WebSocketClientAdapter extends WebSocketNetworkAdapter {
       socket.removeEventListener("error", this.onError)
       socket.close()
     }
+    // Cancel every pending timer and clear its id, so a later connect() can
+    // re-arm the retry interval (the connect() guard re-arms only when the id
+    // is unset).
```
msakrejda:
```suggestion
```

packages/automerge-repo-network-websocket/src/WebSocketServerAdapter.ts
```diff
@@ -93,13 +106,32 @@ export class WebSocketServerAdapter extends NetworkAdapter {
...
+
     const clients = this.server.clients as Set<WebSocketWithIsAlive>
     clients.forEach(socket => {
       this.#terminate(socket)
       this.#removeSocket(socket)
     })
   }
 
+  /** Clear the keepalive interval and detach the listeners we registered on
+   * the shared WebSocketServer in connect(). Idempotent. */
```
msakrejda:
```suggestion
```

msakrejda (approved):
Looks good, thanks! We should drop some of the comments (they're fine, but fairly obvious, and the project tends terser), but otherwise this makes sense.

PR #691, opened 2026-06-27
fix(SyncStateTracker): catch errors from the throttled sync-state save

darcyparker:
## What

The per-storage-id throttled `saveSyncState` handler ran fire-and-forget (`void handler(payload)`) with no error handling, so a rejected storage write surfaced as an unhandled rejection and, in Node, exits the process by default.

## Fix

Wrap the throttled save in try/catch and log, mirroring `StorageSource`'s heads-changed save. Sync state is re-derived from the next sync exchange, so a dropped save is recoverable.

## Tests

Adds a regression test asserting a rejecting `saveSyncState` is caught and logged rather than floated.


msakrejda (approved):
Looks great.

PR #692, opened 2026-06-27
perf(storage): bound storage fan-out and make flush drain before teardown

darcyparker:
## What

`Repo.flush()` and the nodefs adapter's `loadRange()`/`walkdir()` each used `Promise.all(items.map(...))` over a bounded resource, launching every storage write / file read at once. On a sync server holding thousands of documents this spikes the storage adapter's connections, and for nodefs can exhaust the process's file-descriptor limit (EMFILE) when a document has many chunks.

## Fix

- Add a small, dependency-free `semaphore()` concurrency limiter and wrap the fan-outs in it. Concurrency is configurable: `RepoConfig.flushConcurrency` (default 20) and `NodeFSStorageAdapterOptions.maxConcurrency` (default 100), each with TSDoc guidance for picking a value per storage-adapter kind (filesystem / HTTP1.1 / HTTP2 / database).
- `flush()` now drains before it settles: it awaits every save via `allSettled` (collecting failures into an `AggregateError`) instead of fail-fast `Promise.all`, and `shutdown()` disconnects in a `finally`.

## Note: public-contract change to `flush()`

This intentionally changes `flush()`'s rejection behavior: it now rejects with an `AggregateError` after every save has settled, rather than rejecting on the first failure while other saves are still writing. The reason is correctness on teardown: awaiting `flush()` before `shutdown()` should mean nothing is still writing when teardown proceeds, which a fail-fast `Promise.all` does not guarantee.

## Tests

`semaphore` unit tests (including a synchronous-start contract guard), and `flush` drain / `AggregateError` plus shutdown-disconnect-on-failure regressions.

### DIFF
--- a/packages/automerge-repo/src/Repo.ts
+++ b/packages/automerge-repo/src/Repo.ts
@@ -702,8 +702,7 @@ export class Repo extends EventEmitter<RepoEvents> {
    *   `flush()` settles, even if some fail; this is what makes `flush()` safe to
    *   `await` before teardown in {@link Repo.shutdown} (nothing is still writing
    *   when the caller proceeds). Failures are collected and rethrown as an
-   *   `AggregateError`. This differs from a fail-fast `Promise.all`, which would
-   *   reject on the first failure while other saves were still in flight.
+   *   `AggregateError`.
    */
   async flush(documents?: DocumentId[]): Promise<void> {
     if (!this.storageSubsystem) {
@@ -827,7 +826,7 @@ export interface RepoConfig {
    *
    * @remarks
    * Tie this to the constraining resource of your
-   * {@link StorageAdapterInterface}, not a round number:
+   * {@link StorageAdapterInterface}:
    * - **filesystem** (e.g. the nodefs adapter): stay well under the process's
    *   file-descriptor ceiling; the default 20 is comfortable.
    * - **HTTP/1.1-backed**: a browser caps connections per origin at ~6, so a
--- a/packages/automerge-repo/src/Repo.ts
+++ b/packages/automerge-repo/src/Repo.ts
@@ -43,6 +43,15 @@ import { truePromiseFactory } from "./helpers/truePromiseFactory.js"
 import { isPlainObject } from "./helpers/isPlainObject.js"
 import { hasAtLeastOneKey } from "./helpers/has-at-least-one-key.js"
 import { noop } from "./helpers/noop.js"
+import { semaphore } from "./helpers/semaphore.js"
+
+/**
+ * Default for {@link RepoConfig.flushConcurrency}: the number of documents
+ * {@link Repo.flush} writes to storage concurrently. A sync server may hold
+ * thousands of documents; flushing them all at once would spike the storage
+ * adapter's connections / file descriptors / memory, so the fan-out is bounded.
+ */
+const DEFAULT_FLUSH_CONCURRENCY = 20
 
 export type { DocumentProgress } from "./DocumentQuery.js"
 export { DocumentQuery } from "./DocumentQuery.js"
@@ -87,6 +96,7 @@ export class Repo extends EventEmitter<RepoEvents> {
   #remoteHeadsSubscriptions = new RemoteHeadsSubscriptions()
   #remoteHeadsGossipingEnabled = false
   #idFactory: ((initialHeads: Heads) => Promise<Uint8Array>) | null
+  #flushConcurrency: number
 
   constructor({
     storage,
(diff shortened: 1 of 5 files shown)


packages/automerge-repo/src/helpers/semaphore.ts
```diff
@@ -0,0 +1,83 @@
...
+ * once. Returns a gate, `limit`, that you wrap each call in; at most
+ * `concurrency` run concurrently and the rest queue (FIFO), starting as slots
+ * free up.
+ *
+ * @remarks
+ * This is the fix for the most common async pitfall: `Promise.all(items.map(fn))`
+ * launches *every* task at once (promises are eager), overwhelming a bounded
+ * resource such as a connection pool, an HTTP host's connection cap, a
+ * file-descriptor limit, an API rate limit, or just memory. Gate each call so
+ * at most `concurrency` run concurrently:
```
msakrejda:
```suggestion
 * Gate each call so at most `concurrency` run concurrently:
```

packages/automerge-repo/src/helpers/semaphore.ts
```diff
@@ -0,0 +1,83 @@
...
+ * const results = await Promise.all(items.map(item => limit(() => work(item))))
+ * ```
+ *
+ * `Promise.all` still resolves results in input order, and its rejection
+ * semantics are unchanged; the gate only throttles *when* each function is
+ * invoked. A slot is released in `finally`, so a rejecting task does not stall
+ * the queue. Note this is the "run a function under the gate" shape
+ * (`limit(() => fn())`), not a classic `acquire()` / `release()` semaphore.
+ *
+ * **Choosing `concurrency`**: tie it to the constraining resource, not a vibe.
```
msakrejda:
```suggestion
 * **Choosing `concurrency`**: tie it to the constraining resource.
```

packages/automerge-repo/src/Repo.ts
```diff
@@ -665,21 +677,52 @@ export class Repo extends EventEmitter<RepoEvents> {
...
+   * safely (a sync server may hold thousands of documents):
+   *
+   * - **Bounded fan-out.** At most {@link RepoConfig.flushConcurrency} document
+   *   saves run at once, instead of launching every save simultaneously.
+   * - **Drain before settle.** Every save is awaited (`allSettled`) before
+   *   `flush()` settles, even if some fail; this is what makes `flush()` safe to
+   *   `await` before teardown in {@link Repo.shutdown} (nothing is still writing
+   *   when the caller proceeds). Failures are collected and rethrown as an
+   *   `AggregateError`. This differs from a fail-fast `Promise.all`, which would
+   *   reject on the first failure while other saves were still in flight.
```
msakrejda:
```suggestion
   *   `AggregateError`.
```

packages/automerge-repo/src/Repo.ts
```diff
@@ -757,6 +806,24 @@ export interface RepoConfig {
    */
   saveDebounceRate?: number
 
+  /**
+   * Maximum number of documents {@link Repo.flush} (and {@link Repo.shutdown})
+   * write to storage concurrently. Defaults to 20.
+   *
+   * @remarks
+   * Tie this to the constraining resource of your
+   * {@link StorageAdapterInterface}, not a round number:
```
msakrejda:
```suggestion
   * Tie this to the constraining resource of your
   * {@link StorageAdapterInterface}:
```

msakrejda (approved):
Looks good. I think the API contract change makes sense, but I'll leave this open for a bit in case someone else has concerns.

darcyparker:
@msakrejda - FYI; There was a conflict with base branch. I fixed it... and force pushed with lease. (no changes to my work and kept your trimmed comments)

pawelsmolaga (approved):

msakrejda:
Okay, sorry about the delay here--I just didn't have any time for automerge for a while. It looks like there's a conflict with main now--could you rebase @darcyparker? Also, #710 has been merged into the subduction branch and mentions this PR. If I'm reading correctly, that's not relevant until the subduction branch lands, but I wanted to double-check.

darcyparker:
> Okay, sorry about the delay here--I just didn't have any time for automerge for a while. It looks like there's a conflict with main now--could you rebase @darcyparker? Also, #710 has been merged into the subduction branch and mentions this PR. If I'm reading correctly, that's not relevant until the subduction branch lands, but I wanted to double-check.

WRT to subduction branch - if it lands first or this lands first, it should be okay. The common change in both of them is the `semaphore`. It's the same code, so there shouldn't be a conflict.

I am looking at the latest merge conflict. Should be fixed soon.

darcyparker:
@msakrejda ready again

packages/automerge-repo/test/Repo.test.ts
```diff
@@ -1031,6 +1031,52 @@ describe("Repo", () => {
         await flushPromise
       }
     })
+
+    it("drains all saves and rejects with an AggregateError when saves fail", async () => {
```
msakrejda, line 1035: Not a big deal, but we're not really testing "drains all saves" here, right?

msakrejda (approved):
Looks good. Minor test title comment, but I'll merge anyway: it's pretty trivial. Thanks for the improvement.

PR #693, opened 2026-06-27
fix(svelte-store): remove the handle listener when the store has no subscribers

darcyparker:
## What

`createDocumentStore` attached a `change` listener to the handle at store-creation time and never removed it, so every store created and then dropped leaked a handle listener.

## Fix

Move the subscription into `writable`'s start/stop notifier so the listener is attached on the first subscriber and detached when the last unsubscribes.

## Tests

Adds a regression test asserting the store contributes exactly one handle listener while subscribed and none after the last unsubscribe.

---

**Update:** added a test covering catch-up after a subscriber-free gap (the start notifier re-reads the handle, so a returning subscriber sees changes made while nobody was subscribed). This fix also matters for #737 (consumer-driven document GC): under that model an attached listener roots its document, so the missing teardown here would permanently pin every document opened through a store, and subscriber-free gaps become routine once unobserved documents can be evicted and re-materialized.

### DIFF
--- a/packages/automerge-repo-svelte-store/src/lib/index.ts
+++ b/packages/automerge-repo-svelte-store/src/lib/index.ts
@@ -105,16 +105,23 @@ export function createAutomergeStore(repo: Repo) {
   function createDocumentStore<T>(
     handle: DocHandle<T>
   ): AutomergeDocumentStore<T> {
-    // Create a writable store with the current document
-    const { subscribe, set } = writable<Doc<T> | null>(handle.doc() ?? null)
-
-    // Set up change listener
-    const onChange = ({ doc }: { doc: Doc<T> | undefined }) => {
-      set(doc ?? null)
-    }
-
-    // Subscribe to changes
-    handle.on("change", onChange)
+    // Create a writable store with the current document. The start/stop
+    // notifier ties the handle's "change" subscription to the store's own
+    // subscriber lifecycle: attach the listener when the first subscriber
+    // arrives, detach it when the last unsubscribes.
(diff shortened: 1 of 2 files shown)


packages/automerge-repo-svelte-store/src/lib/index.ts
```diff
@@ -105,16 +105,23 @@ export function createAutomergeStore(repo: Repo) {
...
+    const { subscribe, set } = writable<Doc<T> | null>(
+      handle.doc() ?? null,
+      () => {
+        const onChange = ({ doc }: { doc: Doc<T> | undefined }) => {
+          set(doc ?? null)
+        }
+        handle.on("change", onChange)
+        // Re-sync in case the document changed between store creation and the
+        // first subscription.
+        set(handle.doc() ?? null)
```
pawelsmolaga, line 121: @darcyparker you can reuse `onChange` a few lines above

```suggestion
        onChange(handle.doc())
```

pawelsmolaga (approved):

packages/automerge-repo-svelte-store/src/lib/index.ts
darcyparker: Good call. There was a small syntax issue. I am making it `onChange({ doc: handle.doc() })`

darcyparker: Actually, on further reflection `set(handle.doc() ?? null)` is simpler. I think it was a good instinct, but on closer look, because of destructuring, it is better to just call `set()` directly.

PR #694, opened 2026-06-27
fix(broadcastchannel): tear down the channel on disconnect and drop malformed messages

darcyparker:
## What

`connect()` registered an anonymous `message` listener that could never be detached, and `disconnect()` never removed it or closed the channel, so the adapter and the channel leaked for the channel's lifetime and a repeated `connect()` stacked duplicate listeners. Separately, a data message with no data made the listener throw, and a synchronous throw inside an event listener escapes dispatch and can crash a Node process.

## Fix

- Hoist the listener to a named field, remove it and `close()` the channel on `disconnect()`, reopen a fresh channel on reconnect, and guard `disconnect()` to stay idempotent.
- Log the malformed message via the repo's `makeLogger` and drop it instead of throwing.

## Tests

Adds a regression test for the disconnect teardown. Existing suite passes.

### DIFF
--- a/packages/automerge-repo-network-broadcastchannel/src/index.ts
+++ b/packages/automerge-repo-network-broadcastchannel/src/index.ts
@@ -15,11 +15,14 @@
  */
 
 import {
+  makeLogger,
   NetworkAdapter,
   type Message,
   type PeerId,
   type PeerMetadata,
 } from "@automerge/automerge-repo/slim"
+
+const log = makeLogger("automerge-repo:broadcastchannel")
 
 export type BroadcastChannelNetworkAdapterOptions = {
   /** BroadcastChannel name to use */
@@ -30,6 +33,9 @@ export type BroadcastChannelNetworkAdapterOptions = {
 
 export class BroadcastChannelNetworkAdapter extends NetworkAdapter {
   #broadcastChannel: BroadcastChannel
+  // Held on the instance so disconnect() can remove it and connect() can avoid
+  // stacking duplicates.
+  #messageListener?: (e: { data: BroadcastChannelMessage }) => void
   #disconnected = false
   #ready = false
   // reassigned in constructor, but keeps TS from complaining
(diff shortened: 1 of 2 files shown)


packages/automerge-repo-network-broadcastchannel/src/index.ts
```diff
@@ -162,6 +180,11 @@ export class BroadcastChannelNetworkAdapter extends NetworkAdapter {
   }
 
   disconnect() {
+    // Idempotent: a second disconnect() must not post on / close the channel
+    // again (closing it makes a further postMessage throw).
```
msakrejda:
```suggestion
```

packages/automerge-repo-network-broadcastchannel/test/index.test.ts
```diff
@@ -64,4 +64,34 @@ describe("BroadcastChannel", () => {
...
+
+      const removesBefore = removeSpy.mock.calls.length
+      const closesBefore = closeSpy.mock.calls.length
+
+      a.disconnect()
+      expect(removeSpy.mock.calls.length).toBe(removesBefore + 1)
+      expect(closeSpy.mock.calls.length).toBe(closesBefore + 1)
+
+      // A second disconnect is a no-op: it must not throw or touch the channel
+      // again (the channel is already closed).
```
msakrejda:
```suggestion
```

msakrejda (approved):
Looks good, thanks!

packages/automerge-repo-network-broadcastchannel/src/index.ts
```diff
@@ -68,66 +74,78 @@ export class BroadcastChannelNetworkAdapter extends NetworkAdapter {
   connect(peerId: PeerId, peerMetadata?: PeerMetadata) {
     this.peerId = peerId
     this.peerMetadata = peerMetadata
+
+    // disconnect() closes the channel, so reopen one when reconnecting after a
+    // disconnect; otherwise detach the current listener before re-adding so we
+    // never stack duplicates.
```
msakrejda:
```suggestion
```

PR #695, opened 2026-06-27
fix(messagechannel): log and drop a misrouted message instead of throwing

darcyparker:
## What

A message whose `targetId` is not ours made the `MessagePort` `message` listener throw; a synchronous throw inside an event listener escapes dispatch and can terminate a Node process.

## Fix

Log it and drop it instead. Logging is routed through the repo's `makeLogger` (the same mechanism `Repo` uses) rather than the `debug` package directly, dropping the now-unused `debug` dependency.


msakrejda (approved):
Looks good, thanks!

PR #696, opened 2026-06-27
fix(react-hooks): guard useDocHandle against stale results and a stuck rejected cache entry

darcyparker:
## What

`useDocHandle`'s non-suspense effect set state from the resolved `find` with no guard against unmount or an `id` change, and the shared `wrapperCache` was never evicted. Two consequences:

- A late result from a superseded `id` could still call `setHandle`. The url-match guard reconciles it on the next render, so the visible effect was a brief flicker / extra render rather than wrong data, but the stale `setState` is avoidable.
- A `find` that rejects (unavailable doc, or aborted on an `id` change) stayed cached forever, so on the non-suspense path the doc could not recover.

## Fix

- Add an `active` flag, cleared in the effect's cleanup, so a result that arrives after unmount or an `id` change does not `setHandle` for a stale id (this removes the flicker above).
- Evict the rejected wrapper on the non-suspense path so a later render re-fetches instead of replaying the stored rejection.

The suspense-path stuck-on-unavailable case, and the cache's unbounded and not-repo-keyed retention, need a cache redesign (a naive eviction would infinite-loop a suspended render); that is left to follow-up.

## Tests

Adds a regression test asserting the rejected wrapper is evicted. Existing suite passes.

### DIFF
--- a/packages/automerge-repo-react-hooks/src/useDocHandle.ts
+++ b/packages/automerge-repo-react-hooks/src/useDocHandle.ts
@@ -75,14 +75,25 @@ export function useDocHandle<T>(
     if (currentHandle || suspense || !wrapper) {
       return
     }
+    // Stays true until this effect run is torn down (unmount or id change); a
+    // result that arrives after that must not setState for a stale id.
+    let active = true
     wrapper.promise
       .then(handle => {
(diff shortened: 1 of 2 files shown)


msakrejda (approved):
Makes sense, thanks.

PR #697, opened 2026-06-27
fix(react-hooks): ignore a superseded url set in useDocHandles

darcyparker:
## What

`useDocHandles`'s effect resolved pending finds and called `setHandleMap` with no cleanup, so if the url set changed before the finds settled, an older set's `Promise.allSettled` could settle last and overwrite the newer set's map, briefly showing the wrong set of handles (it self-corrects on the next change).

## Fix

Add an `active` flag, cleared in the effect's cleanup, so a superseded run cannot install a stale map.

## Tests

Adds a regression test that switches the url set mid-load and asserts the superseded set does not overwrite the current one.


msakrejda (approved):
Looks good; thanks for the fix.

PR #698, opened 2026-06-27
fix(repo): handle a network-adapter readiness failure at the source

darcyparker:
## What

`networkReady` was `networkSubsystem.whenReady().then(() => {})` with no rejection handler. `whenReady()` is a `Promise.all` over the adapters' own `whenReady()`, so it rejects if any adapter fails to become ready. That derived promise is only consumed by `DocSynchronizer`s, which attach their own `.then().catch()`. If the network failed before any document was open, nothing had attached yet, so the rejection surfaced as an unhandled promise rejection (which in Node can terminate the process).

## Fix

Handle it once at the source: resolve `networkReady` to `void` both on success and on adapter failure (logged via the repo logger), so it never rejects. No consumer acts on the rejection, and resolving on failure matches what the gate means ("peers have had their chance to connect"), so a document on a failed network settles to `unavailable` instead of hanging.

## Tests

Adds a regression test: a repo with a failing network adapter and no documents logs the failure instead of leaking an unhandled rejection. Existing suite passes.

### DIFF
--- a/packages/automerge-repo/src/Repo.ts
+++ b/packages/automerge-repo/src/Repo.ts
@@ -42,6 +42,7 @@ import { Document } from "./Document.js"
 import { truePromiseFactory } from "./helpers/truePromiseFactory.js"
 import { isPlainObject } from "./helpers/isPlainObject.js"
 import { hasAtLeastOneKey } from "./helpers/has-at-least-one-key.js"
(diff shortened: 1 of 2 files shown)


msakrejda (approved):
Thanks for the fix!

PR #699, opened 2026-06-27
perf(sync): bound the loadSyncState fan-out when adding peers

darcyparker:
## What

Adding a peer loads persisted sync state once per document. `CollectionSynchronizer` adds every peer to every document (`addPeer` loops all docs, `attach` loops all peers), so on a many-document sync server one peer connecting issues one `loadSyncState` storage read per document at once, with no bound.

This is the **same family as #692** (which bounds `Repo.flush`'s storage writes): an unbounded burst of storage I/O scaled by collection size. Where #692 bounds the write fan-out, this bounds the per-connect read fan-out.

## Fix

`CollectionSynchronizer` owns one `semaphore` limiter that gates every `loadSyncState` read, so concurrent reads are capped across all documents together (default `DEFAULT_SYNC_STATE_LOAD_CONCURRENCY = 20`). Configurable via `RepoConfig.syncStateLoadConcurrency`, which flows through to `AutomergeSyncConfig.syncStateLoadConcurrency` and defaults when unset.

## Tests

Adds a regression test asserting concurrent `loadSyncState` reads stay within the cap (deterministic deferreds plus a synchronous assertion via the limiter's synchronous-start, no fixed sleeps).

## Stacking

Stacked on #692 for the `semaphore` helper. **#692 should merge first**: the first commit shown here is #692's, and it drops out once #692 lands and this branch rebases onto `main`.

Sibling of #700, which bounds the share-policy callback fan-out driven by the same `addPeer` / `attach` loops (a separate limiter for a separate resource: auth/share-policy callbacks vs storage reads). The two touch the same files, so whichever merges second needs a trivial rebase.

### DIFF
--- a/packages/automerge-repo/src/Repo.ts
+++ b/packages/automerge-repo/src/Repo.ts
@@ -109,6 +109,7 @@ export class Repo extends EventEmitter<RepoEvents> {
     denylist = [],
     saveDebounceRate = 100,
     flushConcurrency = DEFAULT_FLUSH_CONCURRENCY,
+    syncStateLoadConcurrency,
     idFactory,
   }: RepoConfig = {}) {
     super()
@@ -194,6 +195,7 @@ export class Repo extends EventEmitter<RepoEvents> {
           .then(noop, err =>
             this.#log.error("network adapters failed to become ready", err)
           ),
+        syncStateLoadConcurrency,
       },
       denylist
     )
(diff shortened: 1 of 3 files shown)


pawelsmolaga (approved):

msakrejda:
@darcyparker looks like this (and #700) also have the Repo.ts conflict. If you could rebase, that'd be great.

darcyparker:
> @darcyparker looks like this (and #700) also have the Repo.ts conflict. If you could rebase, that'd be great.

Done - they were easy rebases. It was because of the squash commit of 692.  
I think the other PRs will have no conflicts when the 699 and 700 merge because each are independent and also just 1 commit. But I can watch/rebase them as needed.

msakrejda (approved):
Looks good, thanks!

PR #700, opened 2026-06-27
perf(sync): bound the share-policy resolution fan-out

darcyparker:
## What

Resolving a peer's share policy runs a user `announce` / `access` callback, which may hit a DB or auth service. `CollectionSynchronizer` fans these out unbounded in two places:

- **`addPeer`** resolves the policy per peer, and the collection adds every peer to every document (`addPeer` loops all docs, `attach` loops all peers), so one peer connecting fires one resolution per document.
- **`reevaluateDocumentShare`** (on `shareConfigChanged`) re-resolves every peer on every document, i.e. `(docs x peers)` at once.

On a many-peer / many-document sync server, both fire that many callbacks simultaneously.

## Fix

One collection-owned `semaphore` routes **both** paths, so concurrent share-policy resolutions are capped across all peers and documents together (default `SHARE_POLICY_CONCURRENCY = 10`). Configurable via `RepoConfig.sharePolicyConcurrency`, which flows through to `AutomergeSyncConfig.sharePolicyConcurrency` and defaults when unset. `DocSynchronizer.addPeer` takes an optional `limit` (an unbounded passthrough default for standalone use); `reevaluateSharePolicy` keeps its per-call limiter default.

## Tests

A regression test asserts both the `addPeer` and `shareConfigChanged` fan-outs stay within the cap (deterministic deferreds plus a synchronous assertion via the limiter's synchronous-start, no fixed sleeps).

## Stacking

Stacked on #692 for the `semaphore` helper and its `Limit` type. **#692 should merge first**: the first commit shown here is #692's, and it drops out once #692 lands and this branch rebases onto `main`.

Sibling of #699, which bounds the `loadSyncState` storage-read fan-out driven by the same `addPeer` / `attach` loops (a separate limiter for a separate resource: storage reads vs auth/share-policy callbacks). The two touch the same files, so whichever merges second needs a trivial rebase.

### DIFF
--- a/packages/automerge-repo/src/Repo.ts
+++ b/packages/automerge-repo/src/Repo.ts
@@ -110,6 +110,7 @@ export class Repo extends EventEmitter<RepoEvents> {
     saveDebounceRate = 100,
     flushConcurrency = DEFAULT_FLUSH_CONCURRENCY,
     syncStateLoadConcurrency,
+    sharePolicyConcurrency,
     idFactory,
   }: RepoConfig = {}) {
     super()
@@ -196,6 +197,7 @@ export class Repo extends EventEmitter<RepoEvents> {
             this.#log.error("network adapters failed to become ready", err)
           ),
         syncStateLoadConcurrency,
+        sharePolicyConcurrency,
       },
       denylist
     )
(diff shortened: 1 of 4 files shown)


psmolaga (approved):

msakrejda:
@darcyparker looks like this one does need a separate rebase as well.

darcyparker:
@msakrejda - rebase done.

msakrejda (approved):
Looks great, thanks!

PR #701, opened 2026-06-29
fix(indexeddb): settle reads on transaction-level failures

darcyparker:
## What

The IndexedDB adapter's read paths (`load`, `loadRange`) rejected with `request.error`, which is `null` when the failure is transaction-level (quota exhaustion, an abort), so the real reason was lost. They also had no handler for an aborted transaction, so that failure left the returned promise unsettled: a hang, and a pending promise keeps its continuation (and the captured transaction/request) alive.

## Fix

One helper handles failure for every operation. It rejects with `transaction.error ?? request.error` (the transaction error is the one populated on quota / abort) and listens for the request error, the transaction error, and a transaction abort, so reads and writes always settle with a meaningful reason. The write paths already used `transaction.error`; they gain the same abort coverage.

The transaction and request are per-operation objects, collected once the operation settles, so the `.on*` property style is kept: there is nothing to remove, and a handler that fires more than once (a request error also bubbles to the transaction) is harmless because the promise is already settled.

## Tests

A regression test (no `fake-indexeddb` dependency) injects a controllable fake `dbPromise` and drives a transaction-level error and an abort through `load` and `save`. Sabotage-verified: all three fail against the previous behavior.

## Out of scope

Closing the long-lived `IDBDatabase` connection on teardown is a separate, connection-lifetime concern (it needs a dispose hook on `StorageAdapterInterface`). That is addressed in #702.

### DIFF
--- a/packages/automerge-repo-storage-indexeddb/src/index.ts
+++ b/packages/automerge-repo-storage-indexeddb/src/index.ts
@@ -10,6 +10,48 @@ import {
   type StorageKey,
 } from "@automerge/automerge-repo/slim"
 
+/**
+ * Reject on any failure of `transaction` or its `request`, so the caller's
+ * promise always settles.
+ *
+ * A failure can surface through more than one event, so all are wired:
+ *   - `request.onerror` fires for a request-level failure (e.g. a constraint
+ *     violation).
+ *   - `transaction.onerror` / `transaction.onabort` fire for a transaction-level
+ *     failure (quota exhaustion, an I/O error, an explicit abort). They also
+ *     fire when an unhandled request error bubbles up and aborts the
+ *     transaction.
+ *
+ * The reason prefers `transaction.error` because, per the IndexedDB spec, it is
+ * either a reference to the same error the failing request raised (so nothing
+ * is lost) or the transaction-level reason such as `QuotaExceededError`, for
+ * which `request.error` is null. `request.error` is the fallback for the moment
+ * a request fails before the abort has set `transaction.error`. If both are
+ * null (an explicit `abort()`), a generic error is used.
+ *
+ * Exported only so it can be unit-tested directly; `@internal` keeps it out of
+ * the package's published types.
+ * @internal
+ */
+export function rejectOnFailure(
+  transaction: IDBTransaction,
+  request: IDBRequest,
+  reject: (reason: unknown) => void
+): void {
+  const fail = () =>
+    reject(
+      transaction.error ??
+        request.error ??
+        new Error("IndexedDB transaction failed")
+    )
+  // A request error bubbles to the transaction, so `fail` may run more than
+  // once; that is harmless because `reject` is a no-op once the promise has
+  // settled.
+  request.onerror = fail
+  transaction.onerror = fail
+  transaction.onabort = fail
+}
+
 export class IndexedDBStorageAdapter implements StorageAdapterInterface {
   private dbPromise: Promise<IDBDatabase>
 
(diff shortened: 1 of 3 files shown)


packages/automerge-repo-storage-indexeddb/src/index.ts
```diff
@@ -10,6 +10,31 @@ import {
   type StorageKey,
 } from "@automerge/automerge-repo/slim"
 
+/**
+ * Reject on any failure of `transaction` or its `request`, so the returned
+ * promise always settles. Prefers `transaction.error`, which (unlike
+ * `request.error`) carries the reason when the transaction aborts, e.g. on
+ * quota exhaustion; `request.error` is null in that case.
```
msakrejda: If `request.error` is null in that case anyway, should we invert this (to `request.error ?? transaction.error`)? I'm not that familiar with IDB APIs, so what you have might make more sense, but the comment doesn't make it clear to me why that's preferable.

packages/automerge-repo-storage-indexeddb/test/index.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    // overwrites dbPromise with a controllable fake right after constructing.
+    if (typeof globalThis.indexedDB === "undefined") {
+      globalThis.indexedDB = { open: () => ({}) } as unknown as IDBFactory
+    }
+  })
+
+  it("rejects load with the transaction error when the request error is null", async () => {
+    const adapter = new IndexedDBStorageAdapter()
+    const { db, transaction, request } = fakeDb()
+    ;(adapter as any).dbPromise = Promise.resolve(db)
```
msakrejda: It's not great that we're reaching into private state, but I'm not sure if we have a lot of great options if we want to test this at all.

msakrejda (commented):
I had a question about the error coalescing order, and I wanted to see if there's maybe a better way to handle stubbing out the DB (I'll follow up, unless you have ideas). The fix looks good; I just don't love the test reaching into private state (I concede this may be our best option).

packages/automerge-repo-storage-indexeddb/src/index.ts
darcyparker: Good question. Per the IndexedDB spec, `transaction.error` is either a reference to the same DOMException the failing request raised, or the transaction-level reason (e.g. `QuotaExceededError`) for which `request.error` is null. So for these single-request transactions the two orders are actually equivalent: when a request causes the abort they hold the same object, and on a transaction-level failure `request.error` is null so `??` skips it either way. I kept `transaction.error` first since it is the superset (it also carries abort reasons) and rewrote the comment to explain that rather than inverting. Happy to flip it if you would still prefer.

packages/automerge-repo-storage-indexeddb/test/index.test.ts
darcyparker: Agreed. Rather than reach into `dbPromise`, I now export `rejectOnFailure` and test it directly against plain stand-in transaction/request objects (the request-error, transaction-error-when-request-null, abort, and both-null paths). To avoid widening the public API I marked it `@internal` and enabled `stripInternal` for the package, so it is gone from the published `index.d.ts` but still importable by the test. It stays in `index.ts` since it is the only user. If you would rather keep it fully private and test through the adapter instead, I can go back to that.

darcyparker:
@msakrejda - please trigger re-run failed jobs. tests were cancelled... I looked, and it doesn't seem like there are long-running or flaky tests.  And I am pretty sure my change didn't cause this. I think it is just a mysterious case where a re-run will just work.

darcyparker:
> @msakrejda - please trigger re-run failed jobs. tests were cancelled... I looked, and it doesn't seem like there are long-running or flaky tests. And I am pretty sure my change didn't cause this. I think it is just a mysterious case where a re-run will just work.

-- Edit: I closed/re-opened PR. This triggered tests to re-run. They are good now.

packages/automerge-repo-storage-indexeddb/test/index.test.ts
msakrejda: That's a reasonable compromise.

PR #708, opened 2026-06-30
Merge main into subductionjs and resolve conflicts

darcyparker:
Hi @expede, I put this together to merge the latest `main` into `subductionjs` and resolve the conflicts. Mostly I wanted to be helpful and to make sure the conflicts with my recent `main` changes landed in the spirit I intended, rather than getting resolved later without that context.

## What this does

Merges `main` into `subductionjs`. Only three files conflicted; everything else auto-merged.

**Conflict resolutions (kept both sides' intent):**

- `storage/StorageAdapterInterface.ts`: keep both subduction's `saveBatch()` and main's optional `close?()`. Independent additions; the conflict was positional.
- `Repo.ts` imports: drop the now-unused `AbortError` (`find()` picked up main's `signal.throwIfAborted()`, which auto-merged) and keep subduction's imports.
- `test/Repo.test.ts` imports: use main's `isAbortErrorLike` (the abort tests auto-merged to the new assertions) and keep subduction's `toSedimentreeId`.
- `Repo.ts` `shutdown()`: reconciled, see below.

**Two follow-on fixes so the merge is sound:**

1. **Storage teardown unified on `close?()`.** The merge left two teardown contracts: subduction's `Repo.#storage.dispose()` and main's `StorageSubsystem.close()` (which forwards to `adapter.close?()`). I unified on the interface method: `IndexedDBWorkerStorageAdapter` (and its inner worker-backed impl) now implement `close()` instead of `dispose()`, and `Repo.shutdown()` releases storage solely through `storageSubsystem.close()`.
2. **NetworkAdapter readiness rejection.** `NetworkAdapter`'s constructor wires `void this.whenReady().then(() => set("ready"))` with no rejection handler, so an adapter whose `whenReady()` rejects produced an unhandled rejection (a crash in Node). Bringing `main` in added the readiness regression test, which exposed it. Handled it with `noop` (the same shape as the Repo-side handling on `main`); the adapter stays `"connecting"` and the failure is surfaced separately. I also added a direct base-class regression test (`test/NetworkAdapter.test.ts`): a subclass whose `whenReady()` rejects stays `"connecting"` with no unhandled rejection, complementing the repo-level test from `main` that surfaced the leak.

**Housekeeping:**

- Formatted a handful of files that predated the `oxfmt --check` lint step this merge brings in (three subduction sources/tests and two react-remote-heads examples) so `pnpm lint` is green. Pure formatting, no code change, in a separate `style:` commit.

## Validation

[...]

PR #709, opened 2026-06-30
test(subduction): wait on signals instead of fixed sleeps (test-robustness pass)

darcyparker:
> Rebased onto `subductionjs` now that #708 has merged; the diff is just the test
> commits.

## What and why

A small, low-risk robustness pass on `test/subduction/*`. The tests waited for async
work with fixed `pause()` sleeps. A fixed sleep is a guess: too short and it flakes
under CI load, too long and every run pays for it, and it never keys off the actual
completion signal. This replaces them with waits that resolve exactly when the work is
done.

It's deliberately simple, and each commit is scoped to one file. **The per-file commit
messages have the detail**: every one states what became genuinely event-driven versus
what didn't and why. At a glance, the changes are three kinds:

- **Event-driven (the robustness win):** wait on the signal the code already emits,
  the handle's `heads-changed`, `subduction-connection`, `subduction-remote-heads`, the
  `repo.subduction` init promise, `repo.flush()`. A few small composable test helpers
  (`awaitEvent` / `awaitDoc` / `awaitProgress` / `awaitSyncedHandle` /
  `awaitSubductionConnected`) back these; `waitFor` also gains an optional bounded
  timeout that reports the failing assertion.
- **DRY consolidation (cosmetic, labelled as such):** some suites only moved a
  duplicated local `waitForCondition` to the shared `waitFor`. No robustness change;
  the commit titles say "(DRY)".
- **Intentional bounded waits (kept, and commented):** where no signal is exposed. A
  raw `Subduction` server has no "stored" event (so `getBlobs` is polled directly,
  which is also the right independent check for the policy tests), disconnect detection
  is lazy, and "must-not-happen" negatives need a window. These are deliberate, not
  oversights.

No production code changes; tests only. Full subduction suite is green.

## Please review carefully: suggestions from a non-expert

I don't know the subduction internals deeply, so treat this as suggestions. The calls
most likely to be wrong are the "no signal exists, so this stays a bounded wait" ones.
If there's a settle/stored/disconnect signal I missed, those waits should become
event-driven too. Corrections very welcome.

### DIFF
--- a/packages/automerge-repo/test/helpers/awaitDoc.ts
+++ b/packages/automerge-repo/test/helpers/awaitDoc.ts
@@ -1,0 +1,20 @@
+import type { DocHandle } from "../../src/DocHandle.js"
+import { awaitEvent } from "./awaitEvent.js"
+
+/**
+ * Resolve once the handle's document satisfies `predicate`, driven by the
+ * handle's `heads-changed` event rather than a fixed sleep. Returns immediately
+ * if the predicate already holds. Pass `timeout` to bound the wait and reject
+ * with a clear error if it never holds; omit it to fall back on the enclosing
+ * test's own timeout.
+ */
+export async function awaitDoc<T>(
+  handle: DocHandle<T>,
+  predicate: (handle: DocHandle<T>) => boolean,
+  { timeout }: { timeout?: number } = {}
+): Promise<void> {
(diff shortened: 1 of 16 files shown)


expede (approved):
Helpful & got rid of some flaky tests. Thanks @darcyparker !

PR #710, opened 2026-07-01
perf(subduction): bound the compaction delete fan-out

darcyparker:
> Rebased onto `subductionjs` now that #708 has merged; the diff is just the two
> commits below.

## What and why

Stale-blob cleanup during compaction pushed one `deleteCommit` / `deleteFragment`
per stale hash onto a single `Promise.all`, a fan-out over the whole stale set. On a
doc with a long history that's a burst of N concurrent storage deletes against a
possibly single-threaded (worker) adapter.

Everywhere else `SubductionSource` already bounds its fan-out (the shutdown
final-sync uses a `Math.min(16, queue.length)` worker pool; other `Promise.all`s
only await already-in-flight work). This makes the compaction deletes consistent
with that: they now run through a small counting `semaphore`. Ordering and failure
handling are unchanged; it only throttles when each delete starts.

The cap is **configurable**, following the existing `flushConcurrency` pattern:
`RepoConfig.subductionCompactionDeleteConcurrency` (or the
`SubductionSourceOptions.compactionDeleteConcurrency` constructor option), with a
fallback default of **16**, the same ceiling the shutdown final-sync pool uses.

Two commits:

- **`feat(helpers): add a bounded-concurrency semaphore`**: a small
  `limit(() => fn())` gate (at most N run at once, the rest queue FIFO), with tests.
- **`perf(subduction): bound the compaction delete fan-out`**: route the deletes
  through it, with the configurable cap.

## Questions / notes for the reviewer

- **Is 16 the right default?** It's now configurable, so a consumer can tune it, but
  the default should be a sane ceiling for the storage adapters in play. I matched
  the shutdown pool; a maintainer may want a different number.
- **The `semaphore` helper is the same one added by #692** (which bounds storage
  fan-out on `main`). This PR copies it because `subductionjs` doesn't have it yet;
  when the two lines converge on `main`, the duplicate should be dropped (the helper
  is byte-identical). If #692 lands on `main` first, that dedup happens when the
  subduction line next merges `main`.
- I'm not deep in the subduction internals, so please sanity-check that bounding the
  deletes can't interact badly with a concurrent compaction pass.

### DIFF
--- a/packages/automerge-repo/src/Repo.ts
+++ b/packages/automerge-repo/src/Repo.ts
@@ -129,6 +129,7 @@ export class Repo extends EventEmitter<RepoEvents> {
     subductionWebsocketEndpoints,
     subductionAdapters,
     subductionTimeouts,
+    subductionCompactionDeleteConcurrency,
     subductionBlobInterceptor,
   }: RepoConfig = {}) {
     super()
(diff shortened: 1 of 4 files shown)


expede (approved):
Makes sense; only holding off on merging because it's in draft

PR #716, opened 2026-07-09
Merge latest main and subductionjs: resolve conflicts and harden shutdown / IndexedDB teardown

darcyparker:
## Summary

Brings the branch up to date with both `main` and the latest `subductionjs`, resolves the conflicts, and hardens `Repo` / `SubductionSource` shutdown and the IndexedDB adapter along the way.

- Merged the current `main` (through `b9c5498a`, the new release engineering): `main` removed `lerna` in favour of a `scripts/version.mjs` version script plus tag-based trusted-publishing workflows, restructured the published `repository` fields, and dropped the `tmp` override. This branch adopts all of that while keeping the fork's version scheme (`2.6.0-subduction.47`) and publish settings (`publishBranch: subductionjs`, `tag: subduction`, `minimumReleaseAge: 0`).
- Merged the latest `subductionjs` (through `84dddee8`): the Chrome Worker WS adapter fix (#717), the `worker-port` module, `@automerge/automerge` on stable `3.3.2` (off the `3.3.0-fragments.2` prerelease), and the cataloged `@automerge/automerge-subduction` `0.16.1` (#724). The branch is built and tested against the real integrated tree, not just a textually-clean merge.

## Reviewing this PR

`Files changed` includes everything `main` brought in, so it is large. The real changes are the sixteen views below. The merge links (1, 2, 4, 8, 9, 10, 13, 14) show only the conflict resolution as a combined diff, so skip the regenerated `pnpm-lock.yaml` in them.

1. Original `main` merge resolutions (fan-out / flush / shutdown): bfe9fcd2
2. Re-merge of newer `main` (best-effort shutdown, IndexedDB read-settling, dep picks): 4c5cbe94
3. `saveBatch()` abort fix: 78068cc3
4. `subductionjs` .44 merge resolutions: 87d1b00c
5. Shutdown / IndexedDB hardening, tests, and dep re-cataloging: [`0c56fe66...b753fd2d`](https://github.com/automerge/automerge-repo/compare/0c56fe66...b753fd2d)
6. Formatting + lint of the incoming `.44` code, no logic change: 0c56fe66 (oxfmt) and 2310cd25 (oxlint)
7. Monorepo catalog sync, `typescript` / `@types/ws` / `vitest` back to `catalog:`: 7a9a2795
8. Latest `subductionjs` re-merge, `@automerge/automerge` off the fragments prerelease to stable `3.3.2`: 73dcca16
[...]

PR #719, opened 2026-07-09
fix(repo): make shutdown() best-effort instead of rejecting on flush failure

darcyparker:
## Summary

`Repo.shutdown()` drained saves via `flush()` in a `try`/`finally`. The `finally` already disconnected the network and closed storage, but with no `catch`, `flush()`'s `AggregateError` propagated after the `finally` ran, so `shutdown()` rejected. That forced callers to handle a rejection just to shut down. This changes `shutdown()` to catch and log the flush failure instead, so it disconnects the network, closes storage, and resolves. Call `flush()` explicitly beforehand if you need to observe save failures.

Teardown is now also individually guarded: `flush()`, `networkSubsystem.disconnect()`, and `storageSubsystem.close()` each run under their own `try`/`catch`. `disconnect()` can throw synchronously (each adapter's `disconnect()` runs in a `forEach`) and `close()` can reject (a user-supplied `adapter.close()`), so guarding each step keeps one failing step from skipping a later one or rejecting `shutdown()`.

`flush()` itself is unchanged and still rejects with an `AggregateError`, since a caller awaiting it should see the failure.

This aligns `main` with the best-effort `shutdown()` behavior already in `subductionjs` (#716).

## Contract change (please review)

This reverses the shutdown-rejects behavior added in #692: `shutdown()` now resolves on a flush failure rather than rejecting. Any caller relying on `shutdown()` rejecting to detect a failed save must call `flush()` first. Flagging explicitly for a maintainer decision.

## Tests

- The shutdown failure-injection test now asserts `shutdown()` resolves, still disconnects, and logs the failure (renamed accordingly).
- A new test forces both `disconnect()` and `close()` to throw and asserts `shutdown()` still closes storage, resolves, and logs each failure.
- Both flush failure-injection tests drain and assert the fire-and-forget throttled save log via `vi.waitFor` (mirroring `StorageSource.test.ts`), so a failed background save is captured and asserted rather than leaking into a later test's stderr as an unattributed "disk full" error.

## Verification

`pnpm --filter @automerge/automerge-repo build`, `tsc --noEmit` (packages + examples + svelte-check), `oxfmt --check`, `oxlint`, and the full `vitest` suite (848 passing, 2 skipped) all green, with no leaked stderr.

### DIFF
--- a/packages/automerge-repo/src/Repo.ts
+++ b/packages/automerge-repo/src/Repo.ts
@@ -756,18 +756,25 @@ export class Repo extends EventEmitter<RepoEvents> {
   }
 
   async shutdown(): Promise<void> {
-    // Best-effort teardown. Drain saves first (flush awaits all of them), but
-    // swallow and log a flush failure rather than rejecting, so shutdown()
-    // always disconnects the network and closes storage instead of leaking
-    // those resources on a partial save failure. Call flush() explicitly before
+    // Best-effort teardown. Drain saves first (flush awaits all of them), then
+    // disconnect the network and close storage. Each step is guarded and its
+    // failure logged rather than thrown, so one failing step neither skips a
+    // later one nor rejects shutdown(). Call flush() explicitly before
     // shutdown() if you need to observe save failures.
     try {
       await this.flush()
     } catch (err) {
       this.#log.error("error flushing documents during shutdown", err)
-    } finally {
+    }
+    try {
       this.networkSubsystem.disconnect()
+    } catch (err) {
+      this.#log.error("error disconnecting network during shutdown", err)
+    }
+    try {
       await this.storageSubsystem?.close()
+    } catch (err) {
+      this.#log.error("error closing storage during shutdown", err)
     }
   }
 
--- a/packages/automerge-repo/src/Repo.ts
+++ b/packages/automerge-repo/src/Repo.ts
@@ -756,11 +756,15 @@ export class Repo extends EventEmitter<RepoEvents> {
   }
 
   async shutdown(): Promise<void> {
-    // Drain saves first (flush awaits all of them), then always tear down in
-    // the finally so a partial flush failure (which rejects flush) still
-    // disconnects the network and closes storage rather than leaking resources.
+    // Best-effort teardown. Drain saves first (flush awaits all of them), but
+    // swallow and log a flush failure rather than rejecting, so shutdown()
+    // always disconnects the network and closes storage instead of leaking
+    // those resources on a partial save failure. Call flush() explicitly before
+    // shutdown() if you need to observe save failures.
     try {
       await this.flush()
+    } catch (err) {
+      this.#log.error("error flushing documents during shutdown", err)
     } finally {
       this.networkSubsystem.disconnect()
       await this.storageSubsystem?.close()
--- a/packages/automerge-repo/test/Repo.test.ts
+++ b/packages/automerge-repo/test/Repo.test.ts
@@ -1068,9 +1068,9 @@ describe("Repo", () => {
 
     it("shutdown() flushes best-effort, logging a flush failure instead of rejecting", async () => {
       const errSpy = vi.spyOn(console, "error").mockImplementation(() => {})
-      const logged = (needle: string) =>
+      const logged = (str: string) =>
         errSpy.mock.calls.some(call =>
-          call.some(arg => String(arg).includes(needle))
+          call.some(arg => String(arg).includes(str))
         )
       try {
         const storage = new DummyStorageAdapter()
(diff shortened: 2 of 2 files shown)


packages/automerge-repo/src/Repo.ts
```diff
@@ -756,11 +756,15 @@ export class Repo extends EventEmitter<RepoEvents> {
   }
 
   async shutdown(): Promise<void> {
-    // Drain saves first (flush awaits all of them), then always tear down in
-    // the finally so a partial flush failure (which rejects flush) still
-    // disconnects the network and closes storage rather than leaking resources.
+    // Best-effort teardown. Drain saves first (flush awaits all of them), but
+    // swallow and log a flush failure rather than rejecting, so shutdown()
+    // always disconnects the network and closes storage instead of leaking
+    // those resources on a partial save failure. Call flush() explicitly before
```
msakrejda: The `disconnect` is in a `finally`—why would it not have been disconnecting and closing the storage subsystem before?

packages/automerge-repo/src/Repo.ts
```diff
@@ -756,11 +756,15 @@ export class Repo extends EventEmitter<RepoEvents> {
...
+    // shutdown() if you need to observe save failures.
     try {
       await this.flush()
+    } catch (err) {
+      this.#log.error("error flushing documents during shutdown", err)
     } finally {
       this.networkSubsystem.disconnect()
```
msakrejda, line 770: While we're here, can this throw? If so, is it a problem that we may disconnect the network subsystem but not close the storage subsystem when shutting down?

packages/automerge-repo/test/Repo.test.ts
```diff
@@ -1050,13 +1050,28 @@ describe("Repo", () => {
...
+        )
       } finally {
         errSpy.mockRestore()
       }
     })
 
-    it("shutdown() disconnects even if flush() fails", async () => {
+    it("shutdown() flushes best-effort, logging a flush failure instead of rejecting", async () => {
       const errSpy = vi.spyOn(console, "error").mockImplementation(() => {})
+      const logged = (needle: string) =>
```
msakrejda: "needle"? Is this a common convention for this type of argument? I haven't seen that before. If not, maybe we should just use "str" or something?

msakrejda (commented):
I had a couple of questions and a minor comment.

packages/automerge-repo/src/Repo.ts
darcyparker: You're right. The `finally` (from #692) already disconnects the network and closes storage on a flush failure, so nothing was leaking on that path. The only behavior change was that shutdown() rethrew flush's AggregateError instead of resolving. I've reworded the comment to drop the "leaking resources" framing and describe the actual change. The follow-up commit also splits teardown into individually guarded steps per your next comment, which does close the one real leak: a throwing disconnect() skipping close().

packages/automerge-repo/src/Repo.ts
darcyparker: Yes, both can throw:

- **`networkSubsystem.disconnect()`** calls each adapter's `disconnect()` synchronously in a `forEach`, so a misbehaving adapter throws and skips the remaining adapters *and* `storageSubsystem.close()`.
- **`storageSubsystem.close()`** awaits a user-supplied `adapter.close?.()`, which can reject.

In the `try`/`finally` form that meant:

- a throwing `disconnect()` leaked storage (it never reached `close()`), and
- a throwing `close()` rejected `shutdown()`,

both defeating the best-effort goal.

**Fix:** split teardown so `flush()`, `disconnect()`, and `close()` each run under their own `try`/`catch` and log on failure. One failing step no longer skips a later one or rejects `shutdown()`.

I also added a test that forces both `disconnect()` and `close()` to throw and asserts `shutdown()` still closes storage, resolves, and logs each failure.

packages/automerge-repo/test/Repo.test.ts
darcyparker: needle is the haystack/needle search convention (as in strpos), but fair, it's not obvious as a parameter name here. Renamed to substring. Happy to use text or fragment instead if you prefer.

darcyparker: Went with `str` as you originally suggested.

darcyparker:
> I had a couple of questions and a minor comment.

Good feedback/questions. I addressed.

packages/automerge-repo/test/Repo.test.ts
msakrejda: Sure, I mean, I've heard the term, I just don't think I've seen it as an argument. But thanks, this way is clearer.

msakrejda (approved):

PR #725, opened 2026-08-02
Improve adapter contract

msakrejda:
Adds typed errors for DocumentDeleted, DocumentedLoadFailed, and
DocumentUnavailable to communicate the relevant document id when
operations fail. This makes it easier to handle errors far from a
failing `find`.

It also distinguishes "I don't have it" from "I could not check if I
have it," and allows the caller to react differently on a definite
absence versus what could be a transient error.

Update the React useDocument hook for this behavior as well.

### DIFF
--- a/packages/automerge-repo-react-hooks/src/useDocument.ts
+++ b/packages/automerge-repo-react-hooks/src/useDocument.ts
@@ -1,4 +1,7 @@
-import { AnyDocumentId } from "@automerge/automerge-repo/slim"
+import {
+  AnyDocumentId,
+  DocumentDeletedError,
+} from "@automerge/automerge-repo/slim"
 import { ChangeFn, ChangeOptions, Doc } from "@automerge/automerge/slim"
 import { useCallback, useEffect, useState } from "react"
 import { useDocHandle } from "./useDocHandle.js"
@@ -70,7 +73,7 @@ export function useDocument<T>(
     }
     const onChange = () => setDoc(handle.doc())
     const onDelete = () => {
-      setDeleteError(new Error(`Document ${id} was deleted`))
+      setDeleteError(new DocumentDeletedError(handle.documentId))
     }
 
     handle.on("change", onChange)
--- a/packages/automerge-repo-react-hooks/test/useDocument.test.tsx
+++ b/packages/automerge-repo-react-hooks/test/useDocument.test.tsx
@@ -1,6 +1,7 @@
 import {
   AutomergeUrl,
   Doc,
+  DocumentDeletedError,
   generateAutomergeUrl,
 } from "@automerge/automerge-repo"
 import { render, screen, waitFor } from "@testing-library/react"
@@ -99,6 +100,10 @@ describe("useDocument", () => {
 
     // Should trigger error boundary
     expect(screen.getByTestId("error")).toHaveTextContent("Error")
+
+    const error = onError.mock.calls[0]?.[0]
+    expect(error).toBeInstanceOf(DocumentDeletedError)
+    expect(error.documentId).toBe(handleA.documentId)
 
     consoleSpy.mockRestore()
   })
--- a/packages/automerge-repo/src/DocumentQuery.ts
+++ b/packages/automerge-repo/src/DocumentQuery.ts
@@ -4,6 +4,7 @@ import { decodeHeads } from "./AutomergeUrl.js"
 import type { DocumentId, UrlHeads } from "./types.js"
 import type { Segment } from "./subdoc-handles/types.js"
 import { type FindProgress, queryStateToFindProgress } from "./_compat.js"
+import { DocumentLoadFailedError, DocumentUnavailableError } from "./errors.js"
 
 /**
  * The state a {@link DocumentSource} reports for a particular document.
@@ -82,7 +83,12 @@ export interface DocumentProgress<T> {
 
 /** Higher numbers represent earlier availability tiers. */
 export type SourcePriority = number
-type SourceInfo = { state: SourceState; priority: SourcePriority }
+type SourceInfo = {
+  state: SourceState
+  priority: SourcePriority
+  /** If unavailable due to fault */
+  error?: Error
+}
 
 const DEFAULT_SOURCE_PRIORITY = 0
 
@@ -178,7 +184,7 @@ export class DocumentQuery<T> implements DocumentProgress<T> {
 
     // Already unavailable — reject immediately
     if (this.#state.state === "unavailable") {
-      throw new Error(`Document ${this.documentId} is unavailable`)
+      throw new DocumentUnavailableError(this.documentId)
     }
 
     return new Promise<DocHandle<T>>((resolve, reject) => {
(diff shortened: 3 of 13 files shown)


msakrejda:
@darcyparker since you've made some related fixes recently, this may be of interest: any thoughts? I ran into this when working with Suspense in React, but I think the impact is broader. I think we should probably also expand the docs in #679 to cover these semantics.

darcyparker:
Thanks for the ping, @msakrejda, and for writing this up. It is of interest: it lines up with the storage and async hardening I have been doing, and I agree the impact is broader than Suspense. The determinate-vs-undetermined split is the right call, and the dedup design (`errorsEquivalent` over reference-stable causes) is subtle and correct: it keeps a derived-`failed` query from churning subscribers and stops the subscribe/evaluate/recompute cycle from looping.

I read through it closely. A few things I would like to strengthen, roughly in priority order. Two speak directly to what you raised: the Suspense behavior (item 4) is worse than it first looks, and I would be glad to help expand the #679 docs for the load path (item 5).

**1. This is a public behavior change worth a migration note.** Cases that used to settle `unavailable` (a failed storage read with no peers; adapters that never came ready) now settle `failed`, so `find()` / `whenReady()` reject with `DocumentLoadFailedError` there instead of the generic unavailable `Error`. Broad `catch` callers are fine, but anyone who branched on the `unavailable` state or message (notably the "treat unavailable as: create a fresh doc" pattern) is affected. The new behavior is safer (a fault is not proof of absence), which is exactly why it is worth calling out: a CHANGELOG note saying `unavailable` now strictly means "determinately not present," faults are `DocumentLoadFailedError` in `failed`, and callers should prefer `instanceof` over message matching. Worth deciding the semver bump. `Repo.export()` is another `find()` consumer affected, and its TSDoc ("undefined if the document is unavailable") is already inaccurate since `find()` rejects rather than resolving undefined.

**2. Extend `DocumentLoadFailedError` from `AggregateError` so loggers can expand the sub-errors (#726).** The logger is consumer-provided via `setLoggerFactory` (console, [pino](https://getpino.io/), [winston](https://github.com/winstonjs/winston), or an [OpenTelemetry](https://opentelemetry.io/) sink), and a standard `AggregateError` is the shape structured sinks understand. pino, for example, expands `AggregateError.errors` out of the box, serializing each underlying failure with its own stack instead of only the flattened message (today the `causes` map logs as empty objects, since Error fields are non-enumerable). #726 is a small PR into this branch that does exactly that: a single file, `causes` kept for programmatic access but non-enumerable, no behavior change, tests pass, with a before/after log example. Take or leave it.

**3. Retention: bounded and mostly pre-existing; eviction is the real follow-up.** A failed storage read now holds the raw adapter error on the query (`causes[source]`); base retained none on a storage-unavailable settle, so #725 introduces this (and #726 is retention-neutral, it adds `errors` referencing the same objects). Two never-evicted caches hold it: `Repo.#queries` (core) and the react-hooks `wrapperCache` (Suspense path). It is bounded (one error per source per query) and rides on the pre-existing never-eviction of those caches, so I would not try to fix it in this PR. The real lever is eviction, and both caches are already on a path:

- Core `Repo.#queries`: the GC-refactor line (#637, my earlier GC exploration, now closed, plus a planned follow-up) makes a query collectable once the consumer drops the handle, so the core-side entry becomes temporary.
- React `wrapperCache`: a separate package and mechanism (a one-shot promise cache), so it deserves its own react-hooks follow-up (item 4's observe-the-query + active/inactive `gcTime` rearchitecture, which fixes the retry gap and the eviction together), not to be folded into #637.

If per-entry size ever matters, storing a `{ name, message }` surrogate instead of the raw adapter error would shrink each entry and sever its `.cause` graph, but that is a separate change. Do not block this PR on any of it.

**4. React Suspense cannot act on the transient/terminal distinction, and the same cache leaks.**

`DocumentLoadFailedError` is advertised as possibly transient, and the engine preserves that (derived `failed` can return to `loading`/`ready`). But in Suspense mode `useDocHandle` caches the first `find()` rejection in the module-level `wrapperCache` and never evicts or re-subscribes, which has two consequences:

- **Permanent error-boundary throw (the retry gap).** `wrapper.read()` re-throws the captured rejection on every render, the eviction effect is skipped for suspense, and even a boundary reset re-reads the same rejected wrapper and re-throws, so the subtree is wedged until a full reload. A transient network blip is therefore indistinguishable from a terminal delete, which defeats the distinction this PR adds.
- **Memory leak.** Every distinct failing id leaves a rejected wrapper resident for the process lifetime, holding the `DocumentLoadFailedError`, so the fault is pinned by both `wrapperCache` and `Repo.#queries`.

Scope: the never-evicts behavior is pre-existing (it predates #725, and is the same pattern #637 tackles for the core caches). #725 does make each retained entry heavier (it now holds a `DocumentLoadFailedError` with per-source `causes`), but the unbounded accumulation is the pre-existing leak. Best deferred, not fixed here.

The clean fix (a follow-up in the react-hooks package, not #637): observe the query (`DocumentQuery` is already `peek()`/`subscribe()`-able) instead of freezing a one-shot promise, plus an error-reset hook and eviction of inactive entries. That is the TanStack Query model, replicable without the dependency: [`QueryErrorResetBoundary`](https://tanstack.com/query/latest/docs/framework/react/reference/QueryErrorResetBoundary) for retry and [active/inactive `gcTime`](https://tanstack.com/query/latest/docs/framework/react/guides/caching) for eviction. It reuses #637's "drop refs to release" model but is a hook-behavior change, so it belongs in its own PR (see item 3). If it does not land now, at least document that the Suspense hooks treat any failed `find()` as terminal.

**5. Docs (happy to take these after this lands).** This dovetails with #679. The load-path typed error is not the storage-error event #679 declined (it rides the awaited return channel, not a global signal), but #679 as written says the logger is the only programmatic surface, which is no longer true for the load path. You suggested on #679 doing this after #725 lands, which is the easier path (the types exist on main, and it avoids a cross-branch tangle), so I am glad to take the two pieces then:

- Source comments: for example the `DocumentSource` interface, which still documents only the determinate `sourceUnavailable(name)` and should cover the `(name, error)` variant (determinate negative vs undetermined give-up).
- A docs page: the determinate -> `DocumentUnavailableError` / undetermined -> `DocumentLoadFailedError` mapping, scoping the "custom LoggerFactory" advice to the save path, and the create-fresh caveat (`unavailable` means "not obtainable right now," not "does not exist," so a holding peer can still appear later).

We can settle placement (the `dev-docs` folder, the storage reference page, or elsewhere) then.

**Nits:**

- The "still tells wanting peers unavailable when the query failed" test ends with two `await new Promise(setImmediate)` that gate on nothing (the message is emitted synchronously); `vi.waitFor` like the sibling tests would be more consistent.
- The StorageSource failed-load test asserts on the flattened message rather than `instanceof DocumentLoadFailedError` and `.causes.storage`.
- No test pins "a pending source keeps the query `loading` even when another source has reported an undetermined error" (reorder-sensitive in `#computeState`).
- `#networkError` is only threaded in the zero-peers branch of `#updateAvailability`, so a partial-adapter failure that still yields live peers records a determinate `unavailable`; probably fine, but a one-line comment on the choice would help.

msakrejda:
I addressed (2) and the nits. Re: (1), I described the changes in the commit message, but we don't really have a proper CHANGELOG, so there's no great place to note these. I think these are subtle enough that this is sufficient. I plan to follow up on (3) and (4) as separate patches. We should probably revisit (5) (and #679 in general) once those land.

PR #726, opened 2026-08-03
DocumentLoadFailedError extends AggregateError

darcyparker:
Small proposal on top of #725, targeting its branch (`improve-adapter-contract`). Take or leave it.

Makes `DocumentLoadFailedError` extend `AggregateError`, so the per-source errors are also exposed as `errors`. A structured logger then serializes each underlying failure with its own stack, instead of only the flattened message.

### Why

The logger is consumer-provided via `setLoggerFactory` (console, [pino](https://getpino.io/), [winston](https://github.com/winstonjs/winston), or an [OpenTelemetry](https://opentelemetry.io/) sink), so the error shape should degrade gracefully across sinks. A standard `AggregateError` is the shape structured sinks understand: pino, for example, expands `AggregateError.errors` out of the box, serializing each sub-error with its own stack.

Logging the caught `DocumentLoadFailedError` today loses the per-source stacks, and pino renders the `causes` map as empty objects (`Error` fields are non-enumerable):

```jsonc
// today (extends Error)
{
  "type": "DocumentLoadFailedError",
  "message": "Document abc could not be loaded (storage: disk fault; automerge-sync: adapters down)",
  "stack": "...",
  "causes": { "storage": {}, "automerge-sync": {} }
}
```

With `extends AggregateError`:

```jsonc
{
  "type": "DocumentLoadFailedError",
  "message": "Document abc could not be loaded (storage: disk fault; automerge-sync: adapters down)",
  "stack": "...",
  "aggregateErrors": [
    { "type": "Error", "message": "disk fault",    "stack": "..." },
    { "type": "Error", "message": "adapters down", "stack": "..." }
  ]
}
```

### Kept minimal, no behavior change

- `causes` is unchanged for programmatic access (`error.causes.storage`), just made non-enumerable so it is not logged as a redundant empty object next to `errors`.
- The message format is unchanged.
- `errorsEquivalent` and the rest of `DocumentQuery` are untouched. All existing tests pass.

Single-file change to `errors.ts`. It reuses what is already the house convention for multi-failure operations: `Repo.flush()` collects multiple per-document save failures into one `AggregateError`, and the subductionjs branch (#716) does the same in `SubductionSource` (per-entry `flush()` and shutdown teardown). This surfaces multiple per-source load failures the same way, so callers and loggers see a consistent structure across all of them.


darcyparker:
@msakrejda FYI

packages/automerge-repo/src/errors.ts
```diff
@@ -35,24 +35,45 @@ export class DocumentUnavailableError extends Error {
  * Distinct from {@link DocumentUnavailableError}, which asserts a determinate
  * negative. Here the document's absence was never established.
  *
- * {@link DocumentLoadFailedError.causes} is keyed by source name (`"storage"`,
- * `"automerge-sync"`), because more than one source can fail for unrelated
- * reasons.
+ * Extends {@link !AggregateError}: the per-source errors are also exposed as
+ * `errors`, so a structured logger (for example pino) serializes each underlying
+ * failure with its own stack, rather than the flattened message alone.
```
msakrejda:
```suggestion
 * `errors`.
```

packages/automerge-repo/src/errors.ts
```diff
@@ -35,24 +35,45 @@ export class DocumentUnavailableError extends Error {
...
+    super(
+      Object.values(causes),
+      DocumentLoadFailedError.#summary(documentId, causes)
+    )
+    this.name = "DocumentLoadFailedError"
+    this.documentId = documentId
+    // `causes` is the same errors as `errors` (from AggregateError), just keyed
+    // by source. Keep it non-enumerable so a structured logger serializes
+    // `errors` once, rather than also emitting a duplicate `causes` (whose
+    // nested Errors, being non-enumerable, would log as empty `{}`).
```
msakrejda:
```suggestion
    // `errors` once.
```

msakrejda (approved):
We should trim some comments, but otherwise this makes sense. Thanks!

darcyparker:
@msakrejda - I accepted those changes.  I see you approved PR.  (I can't merge... so just mentioning it as FYI)

PR #728, opened 2026-08-07
New release engineering

alexjg:
Problem: the current release process is a bit janky. It is not crash safe, if a release action fails then re-running it won't do anything. It also creates a new commit as part of the github action which feels a bit unpleasant. We also use `lerna` which I find kind of annoying to work with.

Solution: remove `lerna` and use our own version management script. Create a new github actions setup which just creates a tag on pushes to `main` which modify the version. Then add another github action which triggers when tags are created that pushes the package to NPM using trusted publishing.

Note that as part of this change I have replaced the `repository` field in every package.json with a structured object which is required for trustued publishing to work.

See the notes in `CONTRIBUTING.md` for more details.


pawelsmolaga:
@alexjg for versioning, you can use the new "pnpm change" feature. It can greatly simplify the workflow.

alexjg:
If I understand correctly the workflow there would be to use `pnpm change` as part of creating a PR in order to create the changesets and then `pnpm version` to perform the version bump. Is that what you are thinking?

pawelsmolaga:
Yes, that’s the exact workflow. Also, from the changesets, you can generate a changelog automatically, which is a quite useful feature.

.github/workflows/docs.yaml
```diff
@@ -0,0 +1,75 @@
...
+      - uses: actions/checkout@v6
+        with:
+          persist-credentials: false
+
+      - uses: actions/setup-node@v6
+        with:
+          node-version: 24
+
+      - name: Enable pnpm
+        run: corepack enable pnpm
```
psmolaga:
```suggestion
      - name: Install pnpm
        uses: pnpm/action-setup@v6
          
      - name: Setup Node
        uses: actions/setup-node@v6
        with:
          cache: "pnpm"
          node-version-file: ".nvmrc"
```

This sets up the node correctly with pnpm cache enabled.

.github/workflows/release.yaml
```diff
@@ -1,120 +1,89 @@
...
 on:
+  # Tags created with GITHUB_TOKEN do not emit a second workflow event, so the
+  # create-release-tag workflow dispatches this workflow explicitly.
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: The existing v-prefixed tag to publish
+        required: true
+        type: string
+  # Keep the tag trigger so tags pushed by a maintainer are handled too.
```
psmolaga, line 12: You can create a GitHub app and use its token to push a tag. Then this workflow will automatically run. This is a known bypass to trigger push-tag workflows.

alexjg: Yeah I considered that. I think I prefer the explicit workflow dispatch though. Setting up a separate github app seems about as much faff as the explicit workflow dispatch. Are there any downsides you know of to the explicit dispatch?

alexjg:
@pawelsmolaga re. the `pnpm version` stuff. I want to think a bit harder about it before switching. It looks to me like the main benefit of changesets is automated changelog generation. The way I have typically approached this in other projects is to have an "Unreleased" section of a changelog which you add to as part of the PR. Then performing a release is a matter of changing the "unreleased" title to the current release version.

I'm wary of using `pnpm version` to handle version management directly. One of the main reasons I went with a self contained script in this PR is that I don't like coupling our (very simple) version management logic to whatever is implemented in `pnpm`. I've had several experiences in the JS ecosystem where a build tool completely breaks things in an upgrade. I want to avoid a scenario where in order to upgrade to some future version of `pnpm` we have to change our release process.

psmolaga:
> @pawelsmolaga re. the `pnpm version` stuff. I want to think a bit harder about it before switching. It looks to me like the main benefit of changesets is automated changelog generation. The way I have typically approached this in other projects is to have an "Unreleased" section of a changelog which you add to as part of the PR. Then performing a release is a matter of changing the "unreleased" title to the current release version.
> 
> I'm wary of using `pnpm version` to handle version management directly. One of the main reasons I went with a self contained script in this PR is that I don't like coupling our (very simple) version management logic to whatever is implemented in `pnpm`. I've had several experiences in the JS ecosystem where a build tool completely breaks things in an upgrade. I want to avoid a scenario where in order to upgrade to some future version of `pnpm` we have to change our release process.

Understood. I just wanted to mention this pnpm feature so you’re aware of it and can make the final decision that fits best.

scripts/version.mjs
```diff
@@ -0,0 +1,236 @@
+import { spawnSync } from "node:child_process"
+import { existsSync, readFileSync, readdirSync, writeFileSync } from "node:fs"
+import path, { dirname } from "node:path"
+import { fileURLToPath } from "node:url"
+
+const REPO_ROOT = dirname(dirname(fileURLToPath(import.meta.url)))
+const ROOT_PACKAGE_PATH = path.join(REPO_ROOT, "package.json")
+const PACKAGES_DIR = path.join(REPO_ROOT, "packages")
```
psmolaga: If you want to avoid hard-coding paths, you can use the pnpm CLI.

To list all workspace package directories:
```
pnpm -r list --depth -1 --parseable
```

To get the workspace root’s node_modules directory:
```
pnpm root -w
```
Note that pnpm root -w outputs the node_modules path, not the workspace root directory itself.

scripts/version.mjs
```diff
@@ -0,0 +1,236 @@
...
+
+  if (result.status !== 0) {
+    throw new Error(`Invalid commitish: ${commitish}`)
+  }
+
+  return result.stdout.trim()
+}
+
+function readPackage(packagePath) {
+  return readJson(readFileSync(packagePath, "utf8"), packagePath)
```
psmolaga, line 203:
```suggestion
  const { default: config } = await import(packagePath, {
    with: { type: 'json' },
});

return config;
```

Node 24 can import JSON files natively

.github/workflows/create-release-tag.yaml
```diff
@@ -0,0 +1,63 @@
...
+concurrency:
+  group: create-release-tag
+  cancel-in-progress: false
+
+jobs:
+  create-tag:
+    name: Create and publish release tag
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
```
FintanH, line 23: Should we start using v7 here?

FintanH (approved):
Overall, looks good to me, and the `CONTRIBUTING` addition is clear 👌 Once the polish is in, I would say merge away 😊

.github/workflows/create-release-tag.yaml
alexjg: Yes but let's do that in a separate PR.

PR #730, opened 2026-08-08
optionally customize WebSocketClientAdapter WebSocket construction

EdVinyard:
I'd like to authenticate clients connecting via WebSocket using [the Kubernetes trick of smuggling a bearer token in via the Sec-WebSocket-Protocol header value](https://github.com/kubernetes/kubernetes/commit/714f97d7baf4975ad3aa47735a868a81a984d1f0) so that it works for clients either in the browser (where that's the only header we can modify) or with the `ws` library (where we have more control).  I couldn't see a good way to do this without copying `WebSocketClientAdapter` into my application and modifying it, although I'd love it if I just missed a better way to accomplish this.

In this PR, `WebSocketClientAdapter` optionally accepts an additional argument, a "web socket factory" function, in the constructor.  If supplied, that function is then called in `.connect()`, rather than calling `new WebSocket(url)` directly.  In the browser, that might look like this:

```ts
    const socketFactory = (url: string) => new WebSocket(url, [
        `base64url.bearer.authorization.io.${bearerTokenWithoutTrailingPadding}`,
    ]);

    const repo = new Automerge.Repo({
        network: [
            new CustomWebSocketClientAdapter(
                url,
                RETRY_INTERVAL_MS,
                socketFactory,
            ),
        ],
        storage: new Automerge.IndexedDBStorageAdapter(),
    });
```

### DIFF
--- a/packages/automerge-repo-network-websocket/src/WebSocketClientAdapter.ts
+++ b/packages/automerge-repo-network-websocket/src/WebSocketClientAdapter.ts
@@ -22,6 +22,8 @@ abstract class WebSocketNetworkAdapter extends NetworkAdapter {
   socket?: WebSocket
 }
 
+type WebSocketFactory = (url: string) => WebSocket
+
 export class WebSocketClientAdapter extends WebSocketNetworkAdapter {
   #ready = false
   #readyResolver?: () => void
@@ -49,15 +51,18 @@ export class WebSocketClientAdapter extends WebSocketNetworkAdapter {
   #forceReadyTimeoutId?: TimeoutId
   #disconnected = false
   #log = debug("automerge-repo:websocket:browser")
+  #socketFactory: WebSocketFactory
 
   remotePeerId?: PeerId // this adapter only connects to one remote client at a time
 
   constructor(
     public readonly url: string,
-    public readonly retryInterval = 5000
+    public readonly retryInterval = 5000,
+    socketFactory?: WebSocketFactory
   ) {
     super()
     this.#log = this.#log.extend(url)
+    this.#socketFactory = socketFactory ?? (url => new WebSocket(url))
   }
 
   connect(peerId: PeerId, peerMetadata?: PeerMetadata) {
@@ -82,7 +87,7 @@ export class WebSocketClientAdapter extends WebSocketNetworkAdapter {
         this.connect(peerId, peerMetadata)
       }, this.retryInterval)
 
-    this.socket = new WebSocket(this.url)
+    this.socket = this.#socketFactory(this.url)
 
     this.socket.binaryType = "arraybuffer"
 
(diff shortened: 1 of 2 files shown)


packages/automerge-repo-network-websocket/src/WebSocketClientAdapter.ts
```diff
@@ -22,6 +22,8 @@ abstract class WebSocketNetworkAdapter extends NetworkAdapter {
   socket?: WebSocket
 }
 
+type WebSocketFactory = (url: string) => WebSocket
```
msakrejda, line 25: The DOM WebSocket is not quite compatible with `ws`'s  WebSocket. We don't type-check our tests (we probably should), so this didn't come up in CI, but we should straighten out the types here. What do we actually want to require of the object we'll get from the factory?

This type should also be exported here and re-exported through `src/index.ts`.

packages/automerge-repo-network-websocket/src/WebSocketClientAdapter.ts
```diff
@@ -49,15 +51,18 @@ export class WebSocketClientAdapter extends WebSocketNetworkAdapter {
...
   #log = debug("automerge-repo:websocket:browser")
+  #socketFactory: WebSocketFactory
 
   remotePeerId?: PeerId // this adapter only connects to one remote client at a time
 
   constructor(
     public readonly url: string,
-    public readonly retryInterval = 5000
+    public readonly retryInterval = 5000,
+    socketFactory?: WebSocketFactory
```
msakrejda, line 61: That's starting to be a lot of positional parameters. Maybe we should turn the second param into `retryIntervalOrOpts` and accept either a `retryInterval` (for compatibility with current code) or an opts object?

packages/automerge-repo-network-websocket/test/Websocket.test.ts
```diff
@@ -365,6 +365,31 @@ describe("Websocket adapters", () => {
 
       server.close()
     })
+
+    it("should use the provided socket factory to connect", async () => {
+      const port = await getPort() //?
```
msakrejda, line 370: > //?

?

packages/automerge-repo-network-websocket/test/Websocket.test.ts
```diff
@@ -365,6 +365,31 @@ describe("Websocket adapters", () => {
...
+      const browserAdapter = await setupClient({
+        port,
+        retryInterval,
+        socketFactory,
+      })
+
+      const _browserRepo = new Repo({
+        network: [browserAdapter],
+        peerId: browserPeerId,
+      })
```
msakrejda, line 387: This is just ignored?

packages/automerge-repo-network-websocket/test/Websocket.test.ts
```diff
@@ -365,6 +365,31 @@ describe("Websocket adapters", () => {
...
+
+      const { serverAdapter } = await setupServer({ port, retryInterval })
```
msakrejda, line 389: You'll need to also destructure the `server` returned and call `server.close()` to avoid leaking resources in the test.

msakrejda (approved):
Thanks for the contribution! This needs some tidying, but overall I think the feature makes sense.

PR #733, opened 2026-08-19
fix: keep shipped helper declarations out of src

darcyparker:
## Summary

`DummyStorageAdapter` and `DummyNetworkAdapter` import their own package through `"../../src/index.js"` instead of the plain relative `"../index.js"`. `tsc` preserves import specifiers when emitting declarations, so the published `dist/helpers/*.d.ts` files point consumers at the shipped `src/` tree.

The package's `exports` map explicitly exposes these helpers (`"./helpers/*.js"`), so any downstream project that imports `@automerge/automerge-repo/helpers/DummyStorageAdapter.js` for its own tests gets the entire automerge-repo `src/` pulled into its type-check program. Under a stricter consumer tsconfig (`noUncheckedIndexedAccess`, `verbatimModuleSyntax`) that fails the consumer's build on code that was never meant to be type-checked downstream. Suppressing the import with a cast or `@ts-expect-error` does not help, since the resolved declaration graph is still checked project-wide.

This changes the two helper sources to package-internal relative imports, so the emitted declarations resolve within `dist/`:

- `dist/helpers/DummyStorageAdapter.d.ts` now imports from `"../index.js"`
- `dist/helpers/DummyNetworkAdapter.d.ts` now imports from `"../index.js"` and `"./pause.js"`

No runtime change: the specifiers resolve to the same modules.

## Tests

Existing suite passes unchanged (the helpers are exercised throughout). Verified by building and inspecting the emitted `dist/helpers/*.d.ts`, and by consuming the helpers from an external project with `noUncheckedIndexedAccess` + `verbatimModuleSyntax`, which type-checks cleanly after the fix and fails before it.


darcyparker:
@msakrejda - FYI. Simple fix. I encountered this when testing some other work.

msakrejda (approved):
Thanks!

PR #734, opened 2026-08-19
Capture commit/fragment ids on bulk blob load and stop compaction deleting blobs awaiting decryption

jtfmumm:
This PR includes two related fixes:

1. The bulk load used `getBlobs()`, which returns bytes but no commit/fragment ids. This means that on initial load, nothing was recorded in `knownHashes` and the save path rewrote the entire received history as if it were new. When a `BlobInterceptor` transforms new bytes on output, this caused it to re-encrypt the history under the current key. The new `idKeyedBlobs()` reads the ids from storage and adds them to `knownHashes`.

* Performance implications: this avoids unnecessarily re-saving the initially loaded history, but on the other hand calls four extra `loadRange` calls on initial load to look up ids. In some local benchmarking, it appeared to halve writes and double reads during initial load. If we created a new subduction API method for this purpose, we could reduce reads here. 

2. Compaction deleted blobs it could not decrypt yet. A blob whose decryption key had not yet arrived was stored but not yet part of the document, which compaction interpreted as being absorbed into a new fragment. The new `untransformedHashes` keeps track of not-yet-decrypted blobs so that compaction can ignore them. This change has no impact if there is no `BlobInterceptor`.

This also adds `BlobRetryAfterKey.test.ts`, which fails on `subductionjs` and passes with fix 2.

### DIFF
--- a/examples/react-counter/package.json
+++ b/examples/react-counter/package.json
@@ -2,7 +2,7 @@
   "name": "@automerge/automerge-repo-demo-counter",
   "repository": "https://github.com/automerge/automerge-repo/tree/main/examples/react-counter",
   "private": true,
-  "version": "2.6.0-subduction.47",
+  "version": "2.6.0-subduction.48",
   "type": "module",
   "scripts": {
     "dev": "vite --open",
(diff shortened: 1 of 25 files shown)


expede (approved):
Makes sense 👍 🚢

PR #735, opened 2026-08-20
test: add a sync-server memory profile harness

darcyparker:
## What

An env-gated memory profile harness with a synthetic sync-server workload: `test/memoryProfile.sync-server.test.ts`. It is not a pass/fail test. It runs only when `MEMORY_PROFILE=<out.json>` is set (CI and normal test runs skip it) and writes per-round memory samples as JSON for charting. It uses only long-stable public API plus the Dummy adapters, so the identical file runs on branches with different memory models to compare them.

## Why

A long-running relay currently retains every document it has ever synced, for the life of the process. This harness makes that visible and measurable from the end-user point of view, and provides the before/after evidence for the upcoming GC changes that fix it.

## Workload

A relay-shaped server Repo (storage-backed, `sharePolicy: async () => false`, never touching a document itself) serves 24 consecutive ephemeral client sessions. Each session connects a fresh client repo that creates 15 documents with 15 changes each, revisits and edits the 10 most recent documents from earlier sessions, waits until everything reaches server storage, then disconnects and is dropped. After each session: flush, let save throttles drain, force GC, sample heap, rss, and live-document counts.

## Results

Run byte-identical on `main` (before) and on the consumer-driven document GC stack (after):

Sync-server memory before and after document GC

| after 24 sessions (360 documents) | main | with the GC stack |
|---|---|---|
| JS heap after GC | 49.4 MB and climbing (about 65 KB per document) | 27.3 MB, flat |
| Documents held in server memory | 360 (every document ever relayed) | 0 (all evicted; revisits re-materialize from storage) |
| RSS | 186 MB and climbing | 152 MB, flat |

The "after" state is #737 (consumer-driven document GC), which builds on #736 (GC groundwork). Chart and raw samples live on the `darcy/memory-benchmark-assets` branch.

## Reproduce

```sh
MEMORY_PROFILE=/tmp/samples.json NODE_OPTIONS=--expose-gc \
  npx vitest run --project @automerge/automerge-repo \
  packages/automerge-repo/test/memoryProfile.sync-server.test.ts
```

Details are in the commit message.


packages/automerge-repo/test/memoryProfile.sync-server.test.ts
```diff
@@ -0,0 +1,158 @@
...
+    if (cond()) return
+    await yieldMacrotask()
+  }
+  throw new Error(`timed out waiting for ${what}`)
+}
+
+/** Let save-throttle timers fire, then force GC with macrotask yields. */
+async function settleAndGC() {
+  await pause(150)
+  for (let i = 0; i < 6; i++) {
```
msakrejda: Six, huh? Should this have a comment or a constant?

msakrejda (approved):
I think this looks good, and I'll merge it as-is, but given the nature of this library, something like this would be really nice to have in a more user-friendly fashion. E.g., we could have a `scripts` target to generate a file named after the branch and tip commit into a .gitignored `profiles` directory, and another to take everything in `profiles` and draw parallel graphs with something like Chart.js (we could even put it on a disconnected branch like you did with `darcy/memory-benchmark-assets` and document it in the README).

packages/automerge-repo/test/memoryProfile.sync-server.test.ts
darcyparker: Fair question, and checking it was worthwhile — the honest answer is that six isn't load-bearing.

I measured it, running the profile with 1, 2, 3, 6 and 10 passes on both a pinning build and a collecting one:

| passes | heap, pinning build | heap, collecting build |
|---|---|---|
| 1 | 49.53 MB | 27.45 MB |
| 2 | 49.51 MB | 27.40 MB |
| 3 | 49.49 MB | 27.40 MB |
| 6 | 49.49 MB | 27.31 MB |
| 10 | 49.49 MB | — |

Under 0.5% across the whole range, and `serverDocsInMemory` — the metric that's exact rather than measured — is identical at every setting.

So rather than invent a threshold rationale, I've named both literals and said what they actually are: `SETTLE_MS = 150` (tied to the Repo's default 100ms `saveDebounceRate`, so a round's writes are issued rather than sitting in a pending throttle that would hold the document), and `GC_PASSES_PER_SAMPLE = 6` documented as deliberate headroom with those numbers. More than one pass does have a reason — a `WeakValueMap` entry is only pruned by the `FinalizationRegistry` callback that follows its value being collected, so observing the released state can take an extra pass — but six specifically is just margin at negligible cost.

Pushed in 07ccdf88.

darcyparker:
> I think this looks good, and I'll merge it as-is, but given the nature of this library, something like this would be really nice to have in a more user-friendly fashion. E.g., we could have a `scripts` target to generate a file named after the branch and tip commit into a .gitignored `profiles` directory, and another to take everything in `profiles` and draw parallel graphs with something like Chart.js (we could even put it on a disconnected branch like you did with `darcy/memory-benchmark-assets` and document it in the README).

Thanks — I'd like to do this as a follow-up PR.

One design note worth agreeing on first. The subtlety in comparing refs is that checking out A then B runs A's harness against A's library and B's harness against B's library, so any change to the harness silently invalidates the comparison. Rather than detect that with a fingerprint, the collector can make it structural: create a `git worktree` per ref and run one canonical harness (copied in from the invoking ref) against each. Comparability by construction, and the failure mode becomes a loud import error instead of quietly different numbers.

Roughly:
```
pnpm profile:compare main my-branch   # worktree per ref, one harness, chart at the end
pnpm profile:collect                  # current tree only
pnpm profile:chart                    # redraw from profiles/
```

with `profiles/` gitignored as you suggested, and a self-contained HTML — data inlined and Chart.js vendored, so it survives the raw-GitHub / orphan-branch route.

 Two small things that fall out: the collector should set `NODE_OPTIONS=--expose-gc` unconditionally, since `main` lacks the vitest `execArgv` wiring later branches have; and the harness stays restricted to API present at every compared ref — already true, now enforced rather than assumed.

msakrejda:
Yeah, that's a great idea. I was just spitballing.

PR #736, opened 2026-08-20
GC groundwork: WeakValueMap, weak view cache, GC test infrastructure, and standalone memory fixes

darcyparker:
## What

Groundwork for consumer-driven document GC, plus standalone memory fixes that stand on their own. Seven commits, each independently green; details are in the commit messages.

- GC test infrastructure: adaptive `flushGC`/`waitForGC` helpers in the shareable `src/helpers/tests/` (alongside the storage/network adapter suites), vitest `--expose-gc` wiring, and a CI canary that fails loudly if that wiring regresses.
- `WeakValueMap`: a primitive-keyed weak-value map (ported from #637, where it was already reviewed).
- Weak per-heads view cache on `Document`: unbounded per-heads growth becomes cached-while-held.
- `progressAtHeads.whenReady` settles when the signal aborts in the microtask gap (was: permanently pending promise plus a leaked listener).
- Peer metadata is released on disconnect (was: one entry per connection, forever).
- Sync re-engagement from persisted sync state on activation: without it, a peer passively subscribed via an announce-nothing relay silently stops receiving updates after the relay loses in-memory peer state.
- Coverage config: shipped `src/helpers` are measured again (the bare `**/helpers` exclude silently exempted them).

## What it achieves

Everything the GC PR builds on, with no behavior change to document lifetime yet: the Repo still pins documents exactly as on `main`. The three fixes (abort settling, peer metadata, re-engagement) fix real leaks/hangs independently of the GC work.

Part of a stack: the memory benchmark #735 measures the before/after effect; #737 (consumer-driven document GC) builds directly on this branch. Also related: the standalone packaging fix #733 should land first or alongside.

One dependent runs the other way: the re-engagement commit rewrites a peer's resolved share policy from `"share"` to `"announce"`, and draft #743 removes that line once #742 lands an outbound gate that acts on `hasRequested` by itself. It cannot come out here, because without that gate this PR's own re-engagement test fails. Nothing here needs to change first.


packages/automerge-repo/src/helpers/tests/flushGC.ts
```diff
@@ -0,0 +1,131 @@
...
+ *
+ * ## CI flakiness defenses
+ *
+ * GC tests are inherently engine-timing-dependent and have a reputation for
+ * being flaky. The defenses here are deliberate:
+ *
+ *   - **Adaptive polling, not fixed sleeps.** {@link waitForGC} exits as
+ *     soon as collection is observed, so a fast engine pays ~5 ms while a
+ *     loaded CI machine still gets the full timeout budget. No wall-clock
+ *     "wait N seconds and hope" anywhere in the success path.
```
msakrejda:
```suggestion
 *   - **Adaptive polling, not fixed sleeps.** {@link waitForGC} exits as
 *     soon as collection is observed: no wall-clock waits.
```

packages/automerge-repo/src/helpers/tests/flushGC.ts
```diff
@@ -0,0 +1,131 @@
...
+ *   - **Timeout is a failure budget, not a wait.** The success path never
+ *     pays the timeout. The 1 s default exists only so a genuine leak
+ *     surfaces as a deterministic failed assertion within bounded time
+ *     instead of hanging the suite.
```
msakrejda:
```suggestion
```
This is basically just repeating the above, no?

packages/automerge-repo/src/helpers/tests/flushGC.ts
```diff
@@ -0,0 +1,131 @@
...
+ *   - **Boolean return + explicit `expect(...).toBe(true)`.** A timeout
+ *     turns into a loud failed assertion, never a silent pass. Compare
+ *     `await flushGC(); expect(probe.deref()).toBeUndefined()`, where a
+ *     missed collection produces a useful error only by luck of which
+ *     assertion fires first.
```
msakrejda:
```suggestion
 *   - **Boolean return + explicit `expect(...).toBe(true)`.** A timeout
 *     failure to GC can be identified separately from the expected state
 *     being asserted.
```

packages/automerge-repo/src/helpers/WeakValueMap.ts
```diff
@@ -0,0 +1,107 @@
+/**
+ * Map keyed by a primitive (`number | string`) where values are held weakly.
+ * Dead entries are removed automatically via `FinalizationRegistry` once the
+ * value is GC'd.
+ *
+ * Use for pure optimization caches where:
+ *   - the key is a primitive (e.g. a stringified path, a stringified head, a
+ *     document id),
```
msakrejda:
```suggestion
 *   - the key is a primitive
```
This mentions what a primitive is right above

packages/automerge-repo/src/helpers/WeakValueMap.ts
```diff
@@ -0,0 +1,107 @@
...
+ *   - V can be cheaply reconstructed on miss,
+ *   - "cache hit" is never observable as program state.
+ *
+ * Why not `WeakMap`? `WeakMap` keys must be `WeakKey` (object or registered
+ * symbol). It cannot be keyed on a `number` or a `string` — try
+ * `new WeakMap<string, T>()` and TypeScript rejects it. `WeakValueMap` fills
+ * exactly that gap: primitive keys, weak values, with the same automatic
+ * eviction guarantee that `WeakMap` provides for object keys.
```
msakrejda:
```suggestion
 * Why not `WeakMap`? `WeakMap` keys must be `WeakKey` (object or registered
 * symbol). It cannot be keyed on a `number` or a `string`. `WeakValueMap` fills
 * that gap with the same automatic eviction guarantee that `WeakMap` provides
 * for object keys.
```

packages/automerge-repo/src/helpers/WeakValueMap.ts
```diff
@@ -0,0 +1,107 @@
...
+ *
+ * If your key *is* an object, prefer the built-in `WeakMap` instead — it ships
+ * with the language and has cleaner lifetime semantics for that shape.
```
msakrejda:
```suggestion
```

packages/automerge-repo/src/helpers/WeakValueMap.ts
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  // Iteration matches `Map`'s shape (`entries`, `keys`, `values`,
+  // `[Symbol.iterator]`). Dead entries are silently skipped — there's
+  // no way to observe one, and the only way for a value to be in the
+  // underlying map is to be alive at the moment we yield it.
+  //
+  // No `size` — it would lie. The count can change between calls
+  // without any mutation, just from GC, so any single read would be
+  // immediately stale.
```
msakrejda:
```suggestion
  // No `size`. The count can change between calls just from GC, so any
  // single read would be immediately stale.
```

packages/automerge-repo/src/helpers/WeakValueMap.ts
```diff
@@ -0,0 +1,107 @@
...
+      if (value !== undefined) yield [key, value]
+    }
+  }
+
+  [Symbol.iterator](): IterableIterator<[K, V]> {
+    return this.entries()
+  }
+
+  // keys/values iterate #map directly: skips entries()'s per-entry tuple.
+
```
msakrejda:
```suggestion
```

packages/automerge-repo/src/synchronizer/DocSynchronizer.ts
```diff
@@ -568,6 +568,15 @@ export class DocSynchronizer extends EventEmitter<DocSynchronizerEvents> {
       documentId: this.#handle.documentId,
     })
 
+    // A persisted sync state with shared heads is evidence this peer has
+    // synced this document before: restore its engagement so a re-created
+    // synchronizer (after cache eviction) resumes pushing updates to a
+    // passively-subscribed peer instead of waiting for it to speak first.
+    if (syncState && state.sharedHeads.length > 0 && !peer.hasRequested) {
```
msakrejda: The test for this passes even if we drop the sharedHeads check here:
```ts
    if (syncState && !peer.hasRequested) {
```
Should we add a test to ensure that we don't enter this branch without shared heads?

packages/automerge-repo/test/Repo.test.ts
```diff
@@ -2362,6 +2378,73 @@ describe("Repo", () => {
     })
   })
 
+  describe("peer re-engagement after cache eviction", () => {
+    it("resumes pushing updates to a passive peer with persisted sync state", async () => {
```
msakrejda, line 2382: I think this doesn't actually cover the `sharedHeads.length > 0` part of the synchronizer change, right? But that's probably not worth testing separately?

packages/automerge-repo/test/WeakValueMap.test.ts
```diff
@@ -0,0 +1,199 @@
...
+  })
+
+  itGC("retains the entry while the value is still referenced", async () => {
+    const m = new WeakValueMap<string, Box>()
+    const kept = new Box(1)
+    m.set("k", kept)
+
+    // Negative assertion: the value is strongly held, so a best-effort GC
+    // should NOT collect it. waitForGC would always time out here, so use
+    // the fixed-rounds variant.
```
msakrejda:
```suggestion
    // Negative assertion: the value is strongly held, so a best-effort GC
    // should NOT collect it.
```

vitest.config.ts
```diff
@@ -12,13 +12,22 @@ export default defineConfig({
...
     coverage: {
       provider: "v8",
       reporter: ["lcov", "text", "html"],
       skipFull: true,
       exclude: [
         "**/fuzz",
-        "**/helpers",
+        // Test doubles and shared test utilities that live under src.
+        // Real src helpers (throttle, semaphore, WeakValueMap, ...) are
+        // shipped code and stay measured.
```
msakrejda:
```suggestion
```

packages/automerge-repo/test/WeakValueMap.test.ts
```diff
@@ -0,0 +1,199 @@
...
+      m.set("k", v1)
+      m.set("k", v2)
+      v1Probe = new WeakRef(v1)
+    })()
+
+    expect(await waitForGC(v1Probe)).toBe(true)
+    expect(m.get("k")).toBe(v2)
+  })
+
+  itGC("evicts many entries when all values are dropped", async () => {
```
msakrejda, line 177: It looks like this passes even if you replace the FinalizationRegistry callback with a no-op, because  in `this.#map.get(key)?.deref()`, a dead WeakRef derefs to undefined whether or not the finalizer ever ran and pruned `this.#map`. I don't think there's a reliable way to assert eviction without exposing something `/** @internal */`.

msakrejda (approved):
I left some feedback suggesting comment pruning and test tweaks, but overall this looks great. Thanks!

packages/automerge-repo/test/Repo.test.ts
darcyparker: You're right — I checked by deleting `state.sharedHeads.length > 0` and all 686 tests still passed.

I do think it's worth covering, though: that clause is what keeps re-engagement conservative. Without it, any peer with a persisted sync state counts as having requested, so an announce-nothing policy would start announcing the document to a peer it was never granted to.

Added the negative case in 3f15cc5e — a peer whose persisted state shares no heads is not re-engaged. It fails if the condition is removed.

packages/automerge-repo/test/WeakValueMap.test.ts
darcyparker: Confirmed — I replaced the `FinalizationRegistry` callback with a no-op and all 15 tests still passed. Good catch: they were asserting a language guarantee rather than our code.

I tried your `@internal` suggestion, since there's precedent for it — `DocHandle._trieNodeCount` exists to assert exactly this kind of prune in `unified.test.ts`. It doesn't hold up here. Exposing the backing map's size and polling it for five seconds reaches zero when this file's earlier tests are skipped, and never fires a single callback when they aren't (`before=100 after=100`). Cleanup callbacks are best-effort by specification, so the assertion would be reporting test ordering rather than our behaviour.

So I dropped the accessor rather than ship unused API, and documented what these tests do and don't prove instead — 214e803d.

Worth adding, since the memory profile from #735 looks like it ought to settle this: it can't. Both of its metrics are blind to the prune by construction. `serverDocsInMemory` counts `Repo.handles`, which iterates the `WeakValueMap` through `entries()` — and that skips dead refs, so it reads zero whether or not any entry was pruned. And the leftover entries are a Map slot plus a document id and a dead `WeakRef` each, so ~360 of them is tens of KB against a 27 MB heap whose run-to-run variation is around 135 KB. The benchmark is good evidence that documents are collected; it has nothing to say about whether the registry tidied up after them.

darcyparker: Followed this up properly in #744, since the answer turned out to be more interesting than "not assertable".

Three things were tangled together in these tests: our registration bookkeeping, our cleanup callback, and the engine's decision to deliver it. Only the third is unreliable — and it's the only one that isn't ours. Substituting `WeakRef` and `FinalizationRegistry` separates them, and needs no seam in the class: `WeakValueMap` reads both off the global object at construction and at store time, never at module load, so stubbing before construction is enough.

That makes your `@internal` suggestion work after all. My objection was really to the *engine's* timing, not to the accessor — with delivery under the test's control, an entry count is a reliable assertion rather than an order-dependent one. It's keyed by a non-exported symbol rather than an underscore, so unlike `_trieNodeCount` it isn't reachable outside the package.

The mutation coverage is the part worth reporting: a no-op cleanup callback and a missing `unregister` on overwrite both used to pass, and both now fail. The second is the one I care about — "a stale finalizer must not evict the value that replaced it" is the subtle bug in this class, and against real GC you can only ever probe it by luck.

The GC-backed tests stay as they are; they cover what the deterministic ones can't, that values genuinely become collectible.

packages/automerge-repo/test/Repo.test.ts
```diff
@@ -2362,6 +2378,140 @@ describe("Repo", () => {
...
+
+      // Alice edits; the server re-creates the synchronizer and re-engages
+      // only peers with shared heads. Eve is not one of them.
+      aliceHandle.change(d => {
+        d.foo = "v2"
+      })
+      await vi.waitFor(() =>
+        expect(server.handles[documentId]?.doc()).toEqual({ foo: "v2" })
+      )
+      await pause(50)
```
msakrejda, line 2510: What do you think about

```diff
diff --git a/packages/automerge-repo/test/Repo.test.ts b/packages/automerge-repo/test/Repo.test.ts
index 20833e46..7badf03a 100644
--- a/packages/automerge-repo/test/Repo.test.ts
+++ b/packages/automerge-repo/test/Repo.test.ts
@@ -21,6 +21,7 @@ import {
   DocHandle,
   DocumentId,
   LegacyDocumentId,
+  Message,
   PeerId,
   SharePolicy,
 } from "../src/index.js"
@@ -2475,6 +2476,7 @@ describe("Repo", () => {
             isEphemeral: false,
           },
         })
+        return toServer
       }
       connectToServer(alice, "alice" as PeerId)
       await alice.networkSubsystem.whenReady()
@@ -2494,9 +2496,13 @@ describe("Repo", () => {
         "eve-storage" as any,
         A.initSyncState()
       )
-      connectToServer(eve, "eve" as PeerId)
+      const evesLink = connectToServer(eve, "eve" as PeerId)
       await eve.networkSubsystem.whenReady()
 
+      // Record what the server sends Eve.
+      const evesMessages: Message[] = []
+      evesLink.on("message", message => evesMessages.push(message))
+
       await server.removeFromCache(documentId)
 
       // Alice edits; the server re-creates the synchronizer and re-engages
@@ -2507,7 +2513,7 @@ describe("Repo", () => {
       await vi.waitFor(() =>
         expect(server.handles[documentId]?.doc()).toEqual({ foo: "v2" })
       )
-      await pause(50)
+      expect(evesMessages.filter(m => m.documentId === documentId)).toEqual([])
       expect(eve.handles[documentId]).toBeUndefined()
     })
   })
```

here to avoid the `pause`?

darcyparker: Yes, great suggestion, thanks. Applied verbatim in af0664ce.

Confirmed it keeps its teeth: with the `sharedHeads.length > 0` guard dropped so Eve does get re-engaged, the new assertion fails, and it fails on the message rather than on `eve.handles`.

PR #737, opened 2026-08-20
Consumer-driven document memory: GC of unobserved documents

darcyparker:
> **#736 has merged, so this is now based directly on `main`: 12 commits, all of them this PR's.** Resurrects the intent of #637 against the subdoc-handles architecture.

## What it addresses

The `Repo` pins every document it has ever touched for the life of the process: strong per-document entries in `Repo.#queries`, `CollectionSynchronizer.#docSynchronizers`, `StorageSource.#saveFns`, and `SyncStateTracker.#syncInfo`. On a long-running sync server that is an unbounded leak.

## What it does

All per-document coordination state becomes weakly held. A document stays loaded exactly while something observes it: a held `DocHandle` (or `DocumentProgress`), an attached listener or subscriber, a pending `whenReady`, or in-flight repo work. Once nothing does, the GC reclaims the whole per-document cluster; a later `find()` re-materializes it from storage or the network.

Each document's objects already form a nearly self-referential cluster, so one weak reference per document suffices: no dual store, no Repo-level FinalizationRegistry, no `dispose()`. The one thing weak primitives cannot express, "keep alive while observed", is a small refcount on `Document` driven from exactly two choke points (registry listener storage and query subscribers), rooting observed documents in `Repo.#retainedDocuments`. Repo-internal listeners attach through non-exported symbols and are exempt by construction. Unlike #637, this preserves main's "a handle with listeners keeps working" contract.

Consumer-visible along the way: `removeAllListeners()`/`off(event)` no longer detach the repo's internal autosave/sync listeners; `off(event, fn)` removes `once()` wrappers by the original function (EventEmitter3 semantics); `removeFromCache` is documented as explicit teardown; sync-state saves are throttled per document (previously one document's save could silently drop another's); the README gains a memory-lifetime contract section. Details are in the 12 commit messages, each independently green and bisectable.

## Measured effect

The BENCH_PLACEHOLDER harness, run byte-identical on `main` and on this stack:

Sync-server memory before and after document GC

| after 24 ephemeral client sessions (360 documents) | main | this stack |
|---|---|---|
| JS heap after GC | 49.4 MB and climbing | 27.3 MB, flat |
| Documents held in server memory | 360 | 0, revisits re-materialize from storage |

## Open design question

Should peer engagement survive eviction for peers without persisted sync state? Two classes fall outside the persisted-sync-state proof that the re-engagement fix uses: a requestor mid-negotiation (nothing shared yet; under an announce-nothing policy its `find()` hangs until reconnect if the relay evicts mid-exchange), and ephemeral storage-less peers (nothing is ever persisted for them). Unlike a restart, eviction does not sever connections, so the peer gets no signal to re-request.

Candidate directions were: (a) a connection-scoped engagement index (peerId to documentIds, cleared on disconnect, consulted on activation); (b) counting engaged remote peers as retainers, so observed documents simply do not evict; (c) accept and document.

**(b) is ruled out.** Reviewing #742 established that a peer's engagement can be set by evidence the repo cannot attribute to a connection: `NetworkSubsystem` validates `senderId` against the arriving adapter for every message type except ephemeral, where it returns before that check, deliberately, because a relayed ephemeral carries its original author's id rather than the hop's. Under (b) that would make a single ephemeral packet a retention primitive: one packet naming any document, from any connected peer or from a third peer relaying honestly, would pin that document in server memory for as long as the connection lasts. That is the growth this PR exists to bound, so engagement must not be what roots a document.

**(a) remains the direction**, and it is now a small addition rather than a new mechanism. With the share-policy gates routed through one predicate (see #742 and #743), an engagement index only has to set `hasRequested` when a peer activates, which is the same input the persisted-sync-state path already uses; no gate changes and no second representation of engagement. Verified by dropping the `share` to `announce` rewrite on the merged tree: setting `hasRequested` alone restores delivery, with this PR's own re-engagement test still passing. `RemoteHeadsSubscriptions.#subscribedDocsByPeer` is already a `Map<PeerId, Set<DocumentId>>` cleared in `removePeer`, so the shape and lifetime have a precedent in the codebase.

Note that #742 does not close class B on its own: it restores *ephemeral* delivery to a storage-less peer once that peer broadcasts, but not document sync, and only for the direction that peer sends on, so a presenter broadcasting to silent viewers still does not recover. Class A is untouched.

## Follow-ups this enables (not included)

Sibling packages that pin documents on their own and need their own small fixes to benefit: react-hooks (#739, module-global wrapper cache), svelte-store (#693), and storage-nodefs (#738, write-through cache below Repo eviction, so a NodeFS sync server will not see bounded memory from this PR alone). Pre-existing core growth for separate discussion: `StorageSubsystem` chunk/heads tables, `NetworkSubsystem` ephemeral session counts, `RemoteHeadsSubscriptions`. Also worth deciding: the `"./helpers/*.js"` exports glob makes the new `WeakValueMap` implicitly public.

## Verification

Every commit typechecks and tests green on its own; full monorepo suite passes. GC suites use adaptive helpers (no wall-clock sleeps), skip loudly without `--expose-gc`, and a CI canary fails if the wiring regresses. Retention is exercised end to end, including eviction races against synchronous `receiveMessage` and re-engagement over an announce-nothing relay.


darcyparker:
Closing out option (b) in the open design question above.

Reviewing #742 turned up the deciding fact: a peer's engagement can be set by evidence we cannot attribute to a connection. `NetworkSubsystem` checks `senderId` against the arriving adapter for every message type except ephemeral, where it returns before that check, deliberately, because a relayed ephemeral carries its original author's id rather than the hop's. Measured on the per-peer-adapter topologies, one such packet on #742 is enough to make a relay push a document to a peer that never asked (`[]` on `main`, `["sync","sync"]` there), or to make a repo request and persist a document it never asked for.

Under (b) that same packet would pin a document in server memory for the life of the connection, which is exactly the growth this PR is for. So engagement is not safe as a retention root, whatever else we do with it. I have updated the description to mark (b) ruled out.

(a) gets cheaper at the same time. With the gates routed through a single share-policy predicate, an engagement index only needs to set `hasRequested` on activation, the same input the persisted-sync-state path uses. Verified on the merged tree by dropping the `share` to `announce` rewrite: setting `hasRequested` alone restores delivery and this PR's own re-engagement test still passes. `RemoteHeadsSubscriptions.#subscribedDocsByPeer` is already the same shape and lifetime.

Worth recording that #742 does not close class B by itself: it restores ephemeral delivery to a storage-less peer once that peer broadcasts, but not document sync, and only for the direction that peer sends on. A presenter broadcasting to silent viewers still does not recover. Class A is untouched.

PR #738, opened 2026-08-20
fix(storage-nodefs): keep chunks cached only while their write is in flight

darcyparker:
## What it addresses

`NodeFSStorageAdapter` keeps a write-through cache that retains every saved chunk's bytes for the adapter's lifetime: `save()` inserts and nothing ever evicts. A long-running sync server on this adapter accumulates its entire storage directory in process memory. This sits below the Repo, so the document GC work (#737) cannot reach it: a NodeFS-backed relay will not see bounded memory from that stack until this lands.

## What it does

The cache exists for read-after-write consistency, not speed: `writeFile` is not atomic, and a concurrent `load()`/`loadRange()` must see a chunk the moment `save()` is called. This PR keeps exactly that property and nothing more: entries now live only while a write for their key is in flight (refcounted across overlapping saves to the same key) and are dropped when the last write settles. Memory is bounded by concurrent writes instead of total data ever saved. Reads of settled data come from disk, where the OS page cache already does this job.

The first commit makes the shared storage-adapter acceptance tests order-insensitive for `loadRange`: chunk order is not part of the contract, and the previous insertion-order expectation only ever held for this adapter through its permanent cache (a fresh adapter over the same directory already returned directory order).

## Tests

New cache-lifetime tests: entries are gone once writes settle, a read racing an in-flight write sees the written bytes, and the latest bytes stay readable across overlapping saves to one key. Acceptance suites pass for nodefs, indexeddb, lmdb, and the in-memory test adapter.

Standalone against `main`; no dependency on the GC stack. Details are in the two commit messages.


packages/automerge-repo-storage-nodefs/src/index.ts
```diff
@@ -36,7 +36,17 @@ export interface NodeFSStorageAdapterOptions {
...
 export class NodeFSStorageAdapter implements StorageAdapterInterface {
   private baseDirectory: string
-  private cache: { [key: string]: Uint8Array } = {}
+  /**
+   * In-flight writes, readable until their file write settles. `writeFile`
+   * is not atomic and a concurrent `load`/`loadRange` must see a chunk the
+   * moment `save` is called, so reads are served from here while a write is
+   * pending. Entries are dropped once the last overlapping write for their
+   * key completes, so memory is bounded by concurrent writes, not by every
+   * chunk ever saved.
```
msakrejda:
```suggestion
   * pending. Entries are dropped once the last overlapping write for their
   * key completes.
```

msakrejda (approved):
This looks reasonable. Since it's independent of the other memory improvement PRs, I plan to merge it soon. I'll leave it open for a bit for others to comment.

PR #742, opened 2026-08-31
fix(sync): make ephemeral message delivery robust to lost peer state and multi-path topologies

alexwarth:
## What

Two fixes for ephemeral messages silently failing to reach peers, which surfaced in practice as *one-directional* loss of presence/cursor-style traffic through a sync server (peer A's messages reach peer B, but B's never reach A), plus regression tests for both.

Substantially reshaped by review from @darcyparker (alexwarth/automerge-repo#1 and alexwarth/automerge-repo#2, merged here); see "Review history" below.

### 1. Ephemeral traffic from a peer now earns ephemeral delivery back to that peer

The relay, broadcast, and outbound-sync paths in `DocSynchronizer` all skip `"share"`-policy peers (access without announce) that have not engaged with the document, and previously only sync/request messages counted as engagement. Ephemeral messages were routed to the `DocSynchronizer` without registering anything about the sender.

That combination wedges: if the server loses its per-peer doc state while the clients believe they're fully synced — e.g. the document is evicted via `removeFromCache` and later re-created by inbound *ephemeral* traffic (`ensureQuery`) — a peer whose traffic is ephemeral-only (presence, cursors, drag updates) can never restore delivery to itself. Its outbound messages relay fine, but nothing is ever relayed back, indefinitely, and the drops are silent. Whether *you* receive ephemeral messages depends entirely on *your* standing at the relay, which is what makes the failure one-directional.

Now, receiving an ephemeral message from a directly-connected, non-denied peer marks that peer `isPresent` on the document. Presence unlocks exactly one privilege — receiving ephemeral relay/broadcasts for that document — via `#mayReceiveEphemeral`. It deliberately does **not** set `hasRequested`, which unlocks document data and is read by share-policy resolution: `senderId` on an ephemeral message is the one sender id the repo cannot attribute to the arriving connection (relayed messages carry their original author's id), so it must not be able to move document data. A peer that wants the document still has to ask over an attributable sync/request message.

The three previously duplicated gate conditions are collapsed onto one predicate, `#mayReceive`, which switches exhaustively over `SharePolicyState`; `#mayReceiveEphemeral` layers the presence rule on top.

### 2. All copies of one broadcast now share a single (senderId, sessionId, count) stamp

Receivers keep the highest `count` seen per `(senderId, sessionId)` and ignore anything not strictly greater, but each per-peer copy of a broadcast was stamped individually at send time (`count: ++this.#count` per `send()` call), so one logical broadcast carried a different count per peer. In any multi-path topology (mesh, multiple adapters, relays) this both defeats deduplication — the app can receive the same broadcast more than once — and can drop an older message entirely when a copy of a newer one overtakes it on a faster path.

`NetworkSubsystem.stampEphemeralMessage()` now allocates the `EphemeralStamp` once per broadcast; the `DocSynchronizer` stamps each broadcast up front and every per-peer copy shares it (plumbed through `AutomergeSyncConfig`, where the allocator is required — the old per-copy fallback *was* the bug, so it is gone rather than kept reachable). `MessageContents<EphemeralMessage>` is now the full message, reflecting that the stamp is present before the network layer sees it. Comments, the `count` docstring, and the websocket package's CDDL block are updated to describe the watermark rule and per-broadcast stamping.

## Tests

`packages/automerge-repo/test/EphemeralMessages.test.ts` (event-driven, no fixed sleeps):

- steady-state relay in both directions through an announce-nothing server (passes before and after)
- **relay restored in both directions from ephemeral traffic alone after the server evicts the doc** — the wedge scenario above; fails without fix 1
- all copies of one broadcast carry the same `sessionId`/`count` — fails without fix 2
- counts keep increasing across documents from one repo (the counter is per repo, the watermark is per session)
- exactly-once delivery per peer in a fully-connected mesh, asserted only after the duplicate copy has physically arrived — fails without fix 2

`packages/automerge-repo/test/DocSynchronizer.test.ts`, for a `"share"`-policy peer that has only sent ephemeral messages: it receives broadcasts (and did not before it spoke), it is sent no sync/request message, it still gets an `open-doc` when it later requests the document and none before, and a heads-empty `sync` (status stays `"unknown"`, `hasRequested` set) is still answered.

Full `packages/automerge-repo` suite: 671 passed / 3 skipped. `tsc --noEmit`, `oxfmt --check`, and `oxlint` clean.

## Notes for reviewers

- The `"share"` arm of `#mayReceive` is `status !== "unknown" || hasRequested`, which is wider than `main`'s inline gates (`status !== "unknown"` only). This is intentional: a peer whose first message is a heads-empty `sync` rather than a `request` keeps status `"unknown"` while `hasRequested` is set, and under the old gate was never answered. Pinned by a test.
- `stampEphemeralMessage` increments the counter even if no peer passes the broadcast gates; receivers only require counts to increase, not to be consecutive, so gaps are harmless.
- `AutomergeSyncConfig`, `DocSynchronizer`, `CollectionSynchronizer`, and `MessageContents` are not exported from the package entrypoints, so requiring the stamp allocator is not a public API change.

## Known limitations (not addressed here)

- **Presence attribution.** `isPresent` is keyed on `senderId`, which for ephemeral messages is deliberately not validated against the arriving connection. So one peer can name another and unlock presence relay *to* a peer that never opened the document. This is bounded — the named peer must already hold `access`, the shipped `WebSocketServerAdapter` binds `senderId` to the socket, and the adapters that don't are same-browser transports — and it is strictly narrower than the earlier `hasRequested` design, which could move document data. Closing it needs the immediate hop on the wire (e.g. a `viaPeerId`) or sender-binding as part of the adapter contract; tracked as a follow-up issue.
- **Silent receivers.** Relay eligibility is per recipient and is re-established by that recipient's own traffic. After the server loses its peer state, a direction reopens only once the *receiving* side also broadcasts; a presenter broadcasting to entirely silent viewers does not recover until a viewer sends something (sync, request, or ephemeral). Same as `main`; noted so the scope is clear.

## Review history

- alexwarth/automerge-repo#1 (@darcyparker): showed empirically that the original fix 1 gate clauses were inert in this PR's tests (the pre-existing `access && hasRequested → announce` promotion was what fired), split presence from document requests (`isPresent`), collapsed the gates onto `#mayReceive`, replaced fixed sleeps with event waits, and added the `DocSynchronizer`-level tests.
- alexwarth/automerge-repo#2 (@darcyparker): removed an unreachable `open-doc` disjunct, guarded `isPresent` against denied peers, made the stamp allocator required, corrected the dedup documentation to the watermark rule (including the websocket CDDL), and pinned the heads-empty sync case.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


alexwarth:
@pvh this is an attempt to fix a bug that's bit us in the task framework and Livelymerge. PTAL!

packages/automerge-repo/test/EphemeralMessages.test.ts
```diff
@@ -0,0 +1,167 @@
...
+  describe("relay through a sync server", () => {
+    // Topology: alice <-> server <-> dan, server configured like
+    // examples/sync-server (sharePolicy: async () => false), so every peer
+    // on the server is in the "share" state and the relay gates in
+    // DocSynchronizer apply.
+    const setup = async () => {
+      const alice = new Repo({ peerId: "alice" as PeerId })
+      const server = new Repo({
+        peerId: "server" as PeerId,
+        sharePolicy: async () => false, // same as examples/sync-server
```
darcyparker: Nit: https://github.com/automerge/automerge-repo/blob/main/packages/automerge-repo/src/helpers/falsePromiseFactory.ts can be reused here.

darcyparker:
@alexwarth I put five suggested commits on top of your branch as alexwarth/automerge-repo#1, rather than leaving them as review comments, since two of them are easier to judge as code than as prose. Would you take a look and consider merging it into `fix/ephemeral-message-delivery`?

Short version: fix 2 looks right to me and I have not touched it beyond naming its type. For fix 1, reverting each edit in isolation showed that the three gate clauses are inert in this PR's own tests (the pre-existing `access && hasRequested` promotion is what fires), and that routing an unattributable `senderId` into `hasRequested` lets one ephemeral packet move document data. The branch separates "is present on this document" from "asked for this document", collapses the three duplicated gates onto one predicate, and makes the tests wait on events instead of ten fixed sleeps. Your four tests all still pass. Details and measurements are in the PR description.

Happy to drop or rework any individual commit, or to split it if you would rather take some of it and not the rest.

packages/automerge-repo/test/EphemeralMessages.test.ts
darcyparker: Resolved in [5302b54](https://github.com/automerge/automerge-repo/pull/742/commits/5302b5432c439cd357880b96ff5a9aff83377a42), so this can be closed.

packages/automerge-repo/test/EphemeralMessages.test.ts
```diff
@@ -0,0 +1,168 @@
...
+
+      return { alice, server, dan }
+    }
+
+    it("relays both directions in steady state", async () => {
+      const { alice, dan } = await setup()
+
+      const aliceHandle = alice.create<TestDoc>({ foo: "bar" })
+      const danHandle = await dan.find<TestDoc>(aliceHandle.url)
+      await pause(50)
```
msakrejda: Can we avoid arbitrary pauses in tests? This is inevitably going to be flaky in CI. Can we await some system state instead, e.g. via `waitFor`?

msakrejda (commented):
Thanks, @alexwarth! I'll kick off CI. Any thoughts on @darcyparker's comments?

packages/automerge-repo/test/EphemeralMessages.test.ts
darcyparker: FYI: https://github.com/alexwarth/automerge-repo/pull/1/changes/e73fa3040ea7965cf8664369f23b9f0331f28352 is a commit with my take on removing the `pause()`s.

alexwarth: @msakrejda, @darcyparker's suggestions look good to me so I merged them into my PR. (Thanks, @darcyparker!)

darcyparker:
Sorry, I caught some things on a second pass that I missed the first time. Opened alexwarth/automerge-repo#2 with the improvements, same shape as before so you can drop any commit individually.

About half of them are mine to fix: the `Partial<>` type widening and the presence flag came from my previous PR. The rest is a dead `open-doc` disjunct whose comment my change falsified, the optional stamp allocator (whose fallback is the duplicate-delivery bug, kept reachable and pinned by a test), and some doc corrections including the websocket CDDL block, which still describes `count` as per-message.

671 passed / 3 skipped, `tsc`/`oxfmt`/`oxlint` clean.

One thing I deliberately left out because it is your call rather than a patch: `isPresent` keys off `senderId`, which is the one sender id we do not validate against the arriving connection, deliberately, since relayed messages carry the original author's id. So a peer can name another peer and unlock presence delivery to it. `WebSocketServerAdapter` binds `senderId` to the socket so the ws hub is fine, but that is not part of `NetworkAdapterInterface`. Closing it needs the immediate hop on the wire. Details are in the "Still open, and not addressed here" section at the end of alexwarth/automerge-repo#2's description, not this PR's.

alexwarth:
Thanks @darcyparker! I merged both of your PRs and updated the description to match the final design (the old text still described the `hasRequested`​ version). Re: the `senderId`​ attribution thing, I've filed it as a follow-up in #745 — if you want to take a crack at it, go for it! Keeping this PR scoped to the two fixes.

packages/automerge-repo/src/synchronizer/DocSynchronizer.ts
```diff
@@ -646,6 +652,50 @@ export class DocSynchronizer extends EventEmitter<DocSynchronizerEvents> {
...
+   * that every send site asks the same question:
+   *
+   * - `loading`: the policy has not resolved yet, so we don't know.
+   * - `denied`: no access, ever.
+   * - `announce`: we proactively share this document with this peer.
+   * - `share`: access without announce. The peer is only a recipient once it
+   *   has engaged with the document itself, by sending a request/sync message
+   *   (`hasRequested`) or by advertising a status. Until then it is merely
+   *   reachable, and sending to it would be an announcement the policy
+   *   declined to make.
```
msakrejda, line 668:
```suggestion
   *   (`hasRequested`) or by advertising a status.
```

msakrejda (approved):
One comment I think we should trim, but otherwise this looks great. I'll merge soon unless anyone has other concerns.

alexwarth:
thanks, @msakrejda and @darcyparker!

PR #759, opened 2026-09-11
build: move to TypeScript 7, except where the toolchain caps it at 6

darcyparker:
The catalog moves to `^7.0.2`, so `catalog:` means TypeScript 7 and **twelve** packages take it. Two cannot, and say so at the point of the exception with `catalog:ts6`:

| | why |
|---|---|
| root | `typedoc` caps at `6.0.x` and resolves `typescript` as a peer from the root importer |
| svelte-store | `svelte2tsx` calls the TS6 compiler API and has no fallback |

## react-hooks and solid-primitives take 7 too, via a documented fallback

Both build declarations through `unplugin-dts`, which does not provide the TS7 JavaScript Compiler API. Its failure names the way out:

```
[unplugin-dts] The installed "typescript" package does not provide the JavaScript
Compiler API (this happens with TypeScript 7+), and the fallback
"@typescript/typescript6" was not found.
```

Adding `@typescript/typescript6` to both packages lets them typecheck with 7 while declarations are emitted by 6. Verified: both build, and their `.d.ts` output is byte identical to the TS6 baseline.

`svelte2tsx` has no equivalent hook, so svelte-store stays on 6.

## svelte-store tracks sveltejs/language-tools

`svelte-package` emits declarations through svelte2tsx's `emitDts`, which since 0.7.59 throws rather than crashing:

> emitDts is not compatible with TypeScript 7.x. Please use a TypeScript version lower than 7.x.

Upstream work is live in [sveltejs/language-tools#3111](https://github.com/sveltejs/language-tools/pull/3111), the TypeScript 7 content mapper, built against 7.1 nightlies. Editor support lands first and svelte-check after, and `svelte-check` already has experimental `--tsgo` flags. No date announced.

## The typedoc ceiling is real today, and now tracks TypeScript 7.1

Worth recording so it does not get re-litigated. [TypeStrong/typedoc#3098](https://github.com/TypeStrong/typedoc/issues/3098): *"the TypeScript 7 API is a complete rewrite … TypeDoc heavily depends on some internal features in TypeScript 6 which are not present in the TypeScript 7 API"*.

**Update.** The "no timeline" line this section used to quote is out of date. Gerrit0's port went from ~500 compiler errors (2026-05-25) to 70 against 7.1 nightlies (2026-07-30) to 70-80% of tests passing (2026-08-04), with a beta targeted around the 7.1 release. That port is local, master is untouched since 2026-07-13, and nothing is published, so the cap holds today. It should lift with 7.1 rather than being open ended.

[#3128](https://github.com/TypeStrong/typedoc/issues/3128), "Add TypeScript 7 to peer dependencies", was **closed** as a duplicate, so widening the range with a `peerDependencyRules` override is not a shortcut. `0.28.20` is `latest` and has the same ceiling; the `1.0.0-dev.*` versions on npm are from 2020, a dead line.

## The published surface does not move

All **113** emitted `.d.ts` files across the twelve packages are byte identical to TS6's. One `.js` file differs, cosmetically:

```diff
-const it = (_it);
+const it = _it;
```

Sourcemap `mappings` are re-encoded, finer rather than coarser (29,670 segments against 29,468 in `automerge-repo` alone). They affect debugger lookup, not runtime.

## Verification

`pnpm build`, `pnpm build:docs` and `pnpm lint` all exit 0, full suite 890 passed / 4 skipped, and every package resolves the TypeScript it should.

## Merge order

Independent. Clean against #756, #757, #758 and #760.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "oxlint-tsgolint": "^0.23.0",
     "portfinder": "catalog:",
     "typedoc": "^0.28.19",
-    "typescript": "catalog:",
+    "typescript": "catalog:ts6",
     "vite": "catalog:",
     "vite-plugin-wasm": "catalog:",
     "vitest": "catalog:"
--- a/packages/automerge-repo-react-hooks/package.json
+++ b/packages/automerge-repo-react-hooks/package.json
@@ -26,6 +26,7 @@
     "@testing-library/dom": "^10.0.0",
     "@testing-library/jest-dom": "catalog:",
     "@testing-library/react": "^16.3.0",
+    "@typescript/typescript6": "^6.0.2",
     "react": "catalog:",
     "react-dom": "catalog:",
     "react-error-boundary": "catalog:",
--- a/packages/automerge-repo-solid-primitives/package.json
+++ b/packages/automerge-repo-solid-primitives/package.json
@@ -18,6 +18,7 @@
     "@solidjs/testing-library": "^0.8.10",
     "@testing-library/jest-dom": "catalog:",
     "@testing-library/user-event": "^14.6.1",
+    "@typescript/typescript6": "^6.0.2",
     "rollup-plugin-visualizer": "catalog:",
     "solid-js": "^1.9.14",
     "typescript": "catalog:",
--- a/packages/automerge-repo-svelte-store/package.json
+++ b/packages/automerge-repo-svelte-store/package.json
@@ -34,7 +34,7 @@
   "devDependencies": {
     "@sveltejs/package": "^2.5.8",
     "svelte": "catalog:",
-    "typescript": "catalog:"
+    "typescript": "catalog:ts6"
   },
   "watch": {
     "build": {
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -64,8 +64,8 @@ catalogs:
       specifier: ^5.56.4
       version: 5.56.4
     typescript:
-      specifier: ^6.0.3
-      version: 6.0.3
+      specifier: ^7.0.2
+      version: 7.0.2
     uuid:
       specifier: ^14.0.1
       version: 14.0.1
@@ -84,6 +84,10 @@ catalogs:
     ws:
       specifier: ^8.21.0
       version: 8.21.0
+  ts6:
+    typescript:
+      specifier: ^6.0.3
+      version: 6.0.3
 
 overrides:
   semver: ^7.5.2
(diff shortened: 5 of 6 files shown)


alexjg:
Does the fact that the root is capped at Typescript 6 matter much? I had been holding off on this update because both typescript eslint (used in another project) and typedoc seem to be stuck on Typescript 6 which makes me wonder whether Typescript 7 is ready for prime time yet.

darcyparker:
> Does the fact that the root is capped at Typescript 6 matter much? I had been holding off on this update because both typescript eslint (used in another project) and typedoc seem to be stuck on Typescript 6 which makes me wonder whether Typescript 7 is ready for prime time yet.

Short answer on the cap: no, it does not matter much. The language is the same, so no source changes, and the PR shows all 113 emitted `.d.ts` files byte identical. For those two importers it is a build-speed question, not a compatibility one, and the other twelve get the speedup.

On "ready for prime time", I think it helps to split it:

- **The compiler is ready.** 7.0 went stable on 2026-07-08 and is npm `latest`.
- **The programmatic API is not.** 7.0 deliberately shipped without one; it is slated for 7.1, currently targeted at 2026-11-24. Both tools you named are instances of that one gap, and so are the two exceptions here: typedoc, svelte2tsx and unplugin-dts all drive the compiler API. It is why Microsoft publishes `@typescript/typescript6` for side-by-side installs, which is exactly the fallback this PR uses for react-hooks and solid-primitives.

So the ceiling is on tools that drive the compiler, not on the code we ship. 6 was the bridge release that got everyone through the breaking changes, and that part has largely happened.

On eslint: there is none in this repo, we are on oxlint and oxfmt, and I would suggest the same for your other project. The part that matters here is that oxlint's type-aware backend, `oxlint-tsgolint`, embeds the typescript-go checker directly. The version encodes it: `7.0.2001` is TypeScript 7.0.2. Type-aware linting is therefore already running on the TS7 checker and never reads the installed `typescript`, which sidesteps the typescript-eslint block entirely. #760 does that bump. Fair caveat: it is not a drop-in for every eslint plugin and oxc still suggests running both on large configs, but it covers 59 of the 61 typescript-eslint type-aware rules.

I have folded two updates into the PR body:

- **typedoc has moved** since I wrote that section. Gerrit0's port went from ~500 compiler errors in May to 70 in July to 70-80% of tests passing on 2026-08-04, with a beta targeted around the 7.1 release. So the ceiling should lift with 7.1 rather than being open ended.
- **svelte-store** is pinned because `svelte-package` calls svelte2tsx's `emitDts`, which now throws explicitly on 7.x. Work is live in sveltejs/language-tools#3111, the TS7 content mapper, built against 7.1 nightlies, editor support first and svelte-check after. svelte-check already has experimental `--tsgo` flags. No date announced.

PR #763, opened 2026-09-14
fix(sync): hold replies to requesting peers until storage has been checked

### LINKED ISSUE #343: Repo and WebSocket provider send two responses for a "request" - the first unavailable, the second a sync
Working on the automerge-repo-swift implementation, and ran into a bit of a snag and unexpected responses.
I'm running the simple code for automerge-repo-sync-server, but verified the same behaviour from `sync.automerge.org`

The versions involved here:
    "@automerge/automerge": "^2.1.13",
    "@automerge/automerge-repo": "^1.1.5",
    "@automerge/automerge-repo-network-websocket": "^1.1.5",
    "@automerge/automerge-repo-storage-nodefs": "^1.1.5",

The code for the sync-server is being run in a docker container locally, per PR https://github.com/automerge/automerge-repo-sync-server/pull/7

I've created an integration test that does the following:
- creates an ephemeral local repo, adds a document locally, and syncs it to the remote (JS) automerge-repo over web sockets
- It retains the documentID, discards the local repo
- It then creates a new ephemeral repo and requests that documentID from the remote (JS) automerge-repo

In the traces, I'm joining the repo, accepting the peer message, sending a request for the documentID
I get _two_ WebSocket message responses, the first an UNAVAILBLE message, and the second a SYNC message with the contents from the server:

Test log snippet from my tracing/diagnostics
```
WebSocket received: UNAVAILABLE[documentId: 2QJgUvf8oVRFZCiEo8CPsGbH3YHu, sender: storage-server-sync-automerge-org, target: 61851380-52F1-4675-ABA8-7ED98E74B9BE]
We've requested 2QJgUvf8oVRFZCiEo8CPsGbH3YHu from 1 peers:
 - Peer: storage-server-sync-automerge-org
Removing the sending peer, there are 0 remaining:
No further peers with requests outstanding, so marking document 2QJgUvf8oVRFZCiEo8CPsGbH3YHu as unavailable
updating state of 2QJgUvf8oVRFZCiEo8CPsGbH3YHu to unavailable

WebSocket received: SYNC[documentId: 2QJgUvf8oVRFZCiEo8CPsGbH3YHu, sender: storage-server-sync-automerge-org, target: 61851380-52F1-4675-ABA8-7ED98E74B9BE, data: 25 bytes]
PEER: 61851380-52F1-4675-ABA8-7ED98E74B9BE - handling a sync msg from storage-server-sync-automerge-org to 61851380-52F1-4675-ABA8-7ED98E74B9BE
```

At the moment, I'm taking the first `UNAVAILABLE` response as a declarative result and marking the document as such, which is then catching an assertion for being in an unexpected state when I later get a `SYNC` message attempting to update it.

I believe the sending of the first `UNAVAILABLE` message is a bug in the current implementation.

Tanyayayya:
When a peer requests a document the repo does not have in memory yet, `#evaluate` answers it straight away from the empty doc while the storage source is still loading. The storage-deferral guard only covers peers in the `unknown` state; a requesting peer is already `wants`, so it is skipped.

Some clients treat that first empty-heads reply as "unavailable". We hit this with a sync server that creates documents through a separate REST service and loads them from S3 on first request: the Python automerge client got an empty reply, marked the document unavailable, and stalled until its 120 s timeout.

This is the same failure class as #343, reported in 2024 from the Swift client: a premature "unavailable" for a document that then syncs fine. The envelope has changed since (empty-heads sync now, explicit `doc-unavailable` then) but the cause is the same, answering a requester before storage has been checked. The test here is the one asked for in that thread. Fixes #343.

This applies the existing guard to `wants` peers too, so the reply waits until storage has settled. Once storage resolves, the deferred peer is still dirty and gets its message on the next evaluate pass.

The deferral is bounded: `shouldDeferAvailability` only waits on a strictly higher-priority source, so a requester waits for the storage lookup and nothing else. Other peers that are connected but silent do not hold it up. I tried keying on the query being in `loading` instead; it passes the suite, but a single silent peer would then stall a requester with no bound.

When storage settles empty, the wire order in one evaluate pass is an empty-heads sync followed by `doc-unavailable`. So a requester can still see an empty sync for a document that does not exist, just not before storage has been checked, and never without the `doc-unavailable` behind it.

Note on blast radius: a JS peer is unaffected by the empty reply either way. It only promotes a sender to `has` on non-empty heads, so the empty sync leaves the sender `unknown`, the sync source stays pending, and the query stays in `loading` until real data arrives. The change matters for clients that read an initial empty-heads sync as "unavailable", but the guard's own comment already states the principle: don't publish something other peers may read as evidence we don't have the document. An empty-heads sync to a requesting peer is that same evidence in a different envelope.

Full test suite passes (891 tests). Includes a regression test in `DocSynchronizer.test.ts`: a requesting peer gets no reply while the storage source is pending, and does get one once storage settles. It fails on `main` and passes with this change.


alexjg:
This change LGTM, if you can fix the linter error then I'm happy to merge.

Tanyayayya:
Thanks Alex. Linter fixed, it was one line in the test that needed wrapping. Folded into the test commit and force-pushed, so the PR is still two commits.

Tanyayayya:
Thanks for merging. Is a release with this on the horizon? We'd like to move off our local patch as soon as there's a published version to pin.

PR #765, opened 2026-09-15
Add repro: a loose tip absorbed by a fragment lingers in getAllHeads

mykola-vrmchk:
Follow-up to #741. Bumps `@automerge/automerge-subduction` to 0.17.2 (sedimentree_core 0.14.2), which fixed inkandswitch/subduction#286, and `@automerge/automerge` to 3.5.0, whose fragment metadata no longer lists a fragment's own head among its checkpoints — the producer-side trigger for inkandswitch/subduction#301. What remains is a second, independent way for `Subduction.getAllHeads()` to disagree with the materialized document; this PR adds a repro for it.

## The bug

An incremental sync sends each record once, as it appears: `SubductionSource` forwards level-0 loose commits while they are loose, then the fragment once automerge packs them (`src/subduction/source.ts`, the `knownHashes` filter). An accept-only receiver never deletes, so after the fragment-forming change N it holds loose commits 1..N-1 plus a fragment whose head N was never a loose commit. Sedimentree's loose-commit DAG sees no successor for N-1 and advertises it as a head beside N. Automerge, reading the whole change graph, knows N-1 is interior.

Every record in the test comes from `Automerge.getFragmentMetadata` + `bundleFragmentMetadata`, unmodified — the only thing the test chooses is *when* it looks: one snapshot before the fragment-forming change, one after.

```
records fed to Subduction                          getAllHeads    A.getHeads
single snapshot, after the fragment formed         {N}            {N}      ok
before ∪ after (what an incremental receiver       {N, N-1}       {N}      <-- bug
                accumulates)
before ∪ after, plus change N stored loose too     {N}            {N}      ok
```

The last row is the discriminating control: identical storage plus one loose record for N, and N-1 disappears — the missing edge is in the loose-commit DAG, not in the data.

## Relation to inkandswitch/subduction#301

The same tree under `@automerge/automerge` 3.3.2 gives `{N-1}` for the middle row: the fragment self-checkpoints, sedimentree's head filter erases it, and only the stale loose tip survives — exactly the production symptom (`072349b1` advertised, `00cd93ac` held). 3.5.0 removes that trigger for newly emitted fragments; fragments already in storage still carry the self-checkpoint, so the sedimentree-side exemption discussed in inkandswitch/subduction#301 still matters for existing data. The previous version of this test asserted the self-checkpoint directly and no longer reproduces on 3.5.0 (2 of 3 assertions fail), which prompted the rework.

## Why it matters

A peer that advertises sedimentree heads (the DXOS edge service) permanently disagrees with a peer that advertises materialized automerge heads, while both hold byte-identical documents — collection sync reports one differing document forever. 3.5.0 alone does not resolve it: any document whose loose commits reached the receiver before the fragment that absorbed them advertises the stale tip.

## Result

```
SedimentreeHeads.test.ts   Tests  3 passed (3)
packages/automerge-repo (full)       Test Files  47 passed | 1 skipped    Tests  778 passed | 8 skipped
```

Behaviour is stated positively — the BUG test pins the wrong result and fails on any change; no `it.fails`. It also asserts the 3.5.0 precondition (no self-checkpoint) inline, so a run against an older automerge fails for the right reason.

## Note on #741

#741 documents the pre-0.17.2 boundary bug and is asserted against 0.16.1, so it is still valid on the current `subductionjs` pin. On the bumped pins in this PR its expectations need regrading — happy to fold both files into one, or rebase #741, whichever you prefer.

## Run

```
pnpm exec vitest run packages/automerge-repo/test/subduction/SedimentreeHeads.test.ts
```

Changes are appended until one hashes with a `00` prefix (a level ≥ 1 commit); stopping at the first such hash makes the shape deterministic — exactly one fragment, no stored fixtures.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mykola-vrmchk:
Reworked the test so every record it feeds to Subduction comes from
`Automerge.getFragmentMetadata` + `bundleFragmentMetadata` — the same calls
`SubductionSource` makes — rather than from hand-built checkpoint lists. Nothing in the
assert path is chosen by the test.

Two things came out of it:

- `getFragmentMetadata` returns `checkpoints: [head]` for every fragment it emits (1/1,
  14/14, 20/20 across doc sizes), so self-checkpointing is the normal output of the
  production path, not an unusual record.
- Fed that output verbatim, the tree advertises **no heads at all** — with no loose
  commits to fall back on, every fragment erases itself through its own checkpoint.
  Dropping only the checkpoints restores exactly `A.getHeads(doc)`.

3 passed (3); dropped the `it.fails` since the behavior is now stated positively.

